PR #738, opened 2017-01-19
Port README from bitbucket

TheBB:


ninrod:
I had quite some difficulty setting up a clean emacs install with evil mode, comming from vim. 

As evil-mode is a prereq for any vim expatriate, I think we could provide a modern, full and clean init.el for new vim converts in the readme:

```elisp
(require 'package)
(setq package-archives '(("melpa" . "~/.emacs.d/mirror-elpa/melpa/")))
(package-initialize)
(unless (package-installed-p 'use-package)
  (package-refresh-contents)
  (package-install 'use-package))

(use-package evil
  :ensure t
  :config
  (evil-mode))
```

Thoughts?

TheBB:
We can overhaul it later, right now I just want it visible. There are some issues with your version though.

ninrod:
Oh yes, I see, I'm using a local package archive. I'll open a separate issue later with the fixes. thanks.

PR #746, opened 2017-01-25
evil-core.el: Add <leader> functionality

justbur:
Hi @TheBB, thanks for taking this over. I had this idea to make "leader" keys more familiar to vim users. @noctuid let me know what you think too. I think the idea is straightforward and gets to similar syntax to vim. Not a lot of code either :)

Add `evil-send-leader` and `evil-send-localleader` which will push the symbol
<leader> or <localleader> into the `unread-command-events` list. This allows one
to bind a key to `evil-send-leader` to make it a leader key and activate bindings
like the following.

```elisp
(defun hi ()
  (interactive)
  (message "hi"))
(define-key evil-normal-state-map (kbd "<leader> h") 'hi)
```

For convenience, evil-set-leader can be used as follows

```elisp
(evil-set-leader 'normal (kbd "C-c"))
```

An optional argument makes it a localleader.

### DIFF
--- a/evil-core.el
+++ b/evil-core.el
@@ -941,6 +941,36 @@ A return value of t means all states."
      (t
       state))))
 
+(defun evil-send-leader ()
+  "Put symbol leader in `unread-command-events' to trigger any
+<leader> bindings."
+  (interactive)
+  (setq prefix-arg current-prefix-arg)
+  (push '(t . leader) unread-command-events))
+
+(defun evil-send-localleader ()
+  "Put symbol localleader in `unread-command-events' to trigger any
+<localleader> bindings."
+  (interactive)
+  (setq prefix-arg current-prefix-arg)
+  (push '(t. localleader) unread-command-events))
+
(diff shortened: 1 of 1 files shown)


therockmandolinist:
Out of curiosity, would this be used instead of packages like `bind-map` and `general`, or just as a simpler fallback default, if users wanted a leader without an external package?

TheBB:
This looks quite elegant. bind-map at least is quite a bit more powerful than this but it's not like we need to take a position on that.

noctuid:
I don't think this really overlaps much with the functionality of `bind-map` or `general`. It's more similar to what `<leader>` does in vim. Seems like a good idea to me for familiarity/compatibility.

justbur:
That's right. I didn't mean to replace existing packages necessarily, just to fill in a missing feature in the evil package itself.

TheBB:
This didn't deserve to sit so long. Sorry.

PR #747, opened 2017-01-26
fix evil-up-block

willghatch:
This fixes issues with text objects. For example, if we run the code:

    (evil-define-text-object inner-racket-comment (count &optional beg end type)
      (evil-select-paren "#|" "|#" beg end type count nil))
    (evil-define-text-object outer-racket-comment (count &optional beg end type)
      (evil-select-paren "#|" "|#" beg end type count t))

    (define-key evil-inner-text-objects-map "#" 'inner-racket-comment)
    (define-key evil-outer-text-objects-map "#" 'outer-racket-comment)

then we go to some text like this (where ? will represent the cursor position):

    #|this is? a test #|with rackety|# multiline
      and nestable comments|#

then hit vi# we get the following (with < and > denoting the visual range endings)

    #|this is a test #|<with rackety>|# multiline
      and nestable comments|#

This patch makes it work as expected, so the range is:

    #|<this is a test #|with rackety|# multiline
      and nestable comments>|#

The patch passes all the tests run by make terminal, and from a bit of general use with the patch on everything seems fine.

Thanks

### DIFF
--- a/evil-common.el
+++ b/evil-common.el
@@ -1375,48 +1375,41 @@ match-data reflects the last successful match (that caused COUNT
 to reach zero). The behaviour of this functions is similar to
 `up-list'."
   (let* ((count (or count 1))
-         (dir (if (> count 0) +1 -1)))
+         (forwardp (> count 0))
+         (dir (if forwardp +1 -1)))
     (catch 'done
       (while (not (zerop count))
         (let* ((pnt (point))
                (cl (save-excursion
-                     (and (re-search-forward end nil t dir)
+                     (and (re-search-forward (if forwardp end beg) nil t dir)
                           (or (/= pnt (point))
                               (progn
                                 ;; zero size match, repeat search from
                                 ;; the next position
                                 (forward-char dir)
-                                (re-search-forward end nil t dir)))
+                                (re-search-forward (if forwardp end beg) nil t dir)))
                           (point))))
                (match (match-data t))
                (op (save-excursion
-                     (and (re-search-forward beg cl t dir)
+                     (and (re-search-forward (if forwardp beg end) cl t dir)
                           (or (/= pnt (point))
                               (progn
                                 ;; zero size match, repeat search from
                                 ;; the next position
                                 (forward-char dir)
-                                (re-search-forward beg cl t dir)))
+                                (re-search-forward (if forwardp beg end) cl t dir)))
                           (point)))))
           (cond
-           ((and (not op) (not cl))
-            (goto-char (if (> dir 0) (point-max) (point-min)))
+           ((not cl)
+            (goto-char (if forwardp (point-max) (point-min)))
             (set-match-data nil)
             (throw 'done count))
-           ((> dir 0)
-            (if cl
-                (progn
-                  (setq count (1- count))
-                  (if (zerop count) (set-match-data match))
-                  (goto-char cl))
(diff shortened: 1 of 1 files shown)


braham-snyder:
@willghatch this fixes the issue in question for me as well, but also gives me: `evil-select-block: No surrounding delimiters found` when point is instead on the other side of the nested pair -- e.g., here:

```
#|this is a test #|with rackety|# multiline
  and ?nestable comments|#
```

ninrod:
@willghatch, you may want to take a look at [targets.el](https://github.com/noctuid/targets.el)

willghatch:
Bleh, sorry for making a PR with bad code.  I've redone it so it handles the forward and backward cases symmetrically rather than doing something different for each like the original code did.  Now it gets both cases right.

@ninrod thanks for the tip, I'll look at it.  It's been a while since I've updated my emacs config much, so I may be due for some overhaul.

evil-common.el
```diff
@@ -1375,48 +1375,43 @@ match-data reflects the last successful match (that caused COUNT
 to reach zero). The behaviour of this functions is similar to
 `up-list'."
   (let* ((count (or count 1))
-         (dir (if (> count 0) +1 -1)))
+         (forwardp (> count 0))
+         (dir (if forwardp +1 -1))
+         (beg+ (if forwardp beg end))
```
wasamasa: The `+` is rather misleading here, I'd have expected it to be an incremented beginning.  Perhaps it's best to replace the two references to each with the relevant if expression?

willghatch: I wasn't wild about the name either, but I hadn't thought of anything else I thought was particularly better.  I've now replaced the references with the expression itself.

willghatch:
Apparently when I force pushed over the previous commit the comments went away too.  But the new difference is that I took out the beg+/end+ names and just put the literal expressions where they were used.  I think I liked it better with names, though I agree those weren't great.  I just hadn't thought of any better ones.

wasamasa:
Hm, for some reason tests fail here.  In graphical mode it's about ten, in textual mode I get a segfault.

willghatch:
Thanks!

ninrod:
people. this actually breaks [targets.el](https://github.com/noctuid/targets.el) and also this famous macro to define new text objects. 

see #884 for discussion. Also see noctuid/targets.el#19

```elisp
;; modified to be able to specify name and use `function'
(defmacro define-and-bind-text-object (name key start-regex end-regex)
  (let ((inner-name (make-symbol (concat "evil-inner-" name)))
        (outer-name (make-symbol (concat "evil-a-" name))))
    `(progn
       (evil-define-text-object ,inner-name (count &optional beg end type)
         (evil-select-paren ,start-regex ,end-regex beg end type count nil))
       (evil-define-text-object ,outer-name (count &optional beg end type)
         (evil-select-paren ,start-regex ,end-regex beg end type count t))
       (define-key evil-inner-text-objects-map ,key #',inner-name)
       (define-key evil-outer-text-objects-map ,key #',outer-name))))

(define-and-bind-text-object "pipe" "|" "|" "|")
```

ninrod:
__note__: emacs-evil/evil@a5465f3 can be reverted cleanly from evil git head. This solves the problem.

willghatch:
Note that #889 fixes this and adds some tests for the broken behavior.

Issue #684, opened 2016-07-07
evil-jump-backward switches buffers


TheBB:
Originally reported by: **Constantine (Bitbucket: [Hi-Angel](https://bitbucket.org/Hi-Angel), GitHub: [Hi-Angel](https://github.com/Hi-Angel))**

----------------------------------------

Since some moment evil-jump-backward *(C-o in normal mode)* instead of going to a previous location in the current buffer, switches to a previous buffer, rendering the command almost unusable.

Emacs version: GNU Emacs 25.1.50.2 (x86_64-pc-linux-gnu, GTK+ Version 3.10.8) of 2016-04-25

Evil version: 1.2.12

Operating System: Ubuntu-16.04

Whether you use Emacs in X or terminal mode: X

A minimal example to reproduce the error:

* Open Emacs (i.e. it would show \*scratch\* buffer).
* Open a file.
 * Press C-o whilst in «normal mode» — it ought to do nothing, as no movement was in the buffer yet.
 * Suddenly find yourself in the \*scratch\* buffer
* ☹

----------------------------------------
- Bitbucket: https://bitbucket.org/lyro/evil/issue/684

salamynder:
+1 for defcustom switch between these two behaviours :)

Hi-Angel:
> BTW, if C-o opens a new buffer, you can always come back with C-i.

@VanLaser yeah, except that I gonna stuck in a random place instead of getting where I was before in the buffer.

VanLaser:
@Hi-Angel - random place? Maybe using 'save-place-mode' would help?

Hi-Angel:
@VanLaser being random is in the sense that I have been editing a function in buffer α, then I jumped to a definition buffer β, then I jumped to a new definition, which in turn led me to buffer α. Now I want to get back to the function I been editing, but <kbd>C-o</kbd> leads me to buffer β, so I stuck in random place in buffer α, and need to manually search where I was editing the function.

A workaround in this particular case is to press <kbd>u</kbd> then <kbd>C-r</kbd>, but there're many situations, like for code exploration, when I can't easily return back.

VanLaser:
No, it's not random, you should simply continue to press `C-o` until you land back to the exact desired position.

Hi-Angel:
[...]

PR #749, opened 2017-01-27
implement a flag for prohibiting the jump commands to switch buffers

c4eater:
This is a fix for #684.
It implements a new flag, evil-jumps-cross-buffers, which is allowing the user to switch to the old behavior.
Note, the current implementation is a bit sub-optimal, because when this option is set to false, every jump command basically makes Evil to filter out irrelevant jump marks from the jump list, and this is repeated over and over. Unfortunately, a more optimized implementation would significantly diverge from the original one (for it would require to completely refactor evil--jumps-window-jumps).


Hi-Angel:
Original reporter here, tested, works for me.

Hi-Angel:
Actually I found a small issue: open a file, search for the word under cursor with `*` *(so that it jumps to the next occurence)*, then press <kbd>C-o</kbd> to return where you were. It does only work for the second press, like if the first <kbd>C-o</kbd> gets eaten.

Hi-Angel:
I found the problem. In `evil--jump-backward` the code


	+        (if (= idx -1)
	+            (progn
	+              (setq idx 0)
	+              (setf (evil-jumps-struct-idx struct) 0)
	+              (evil--jumps-push))
	+          (evil--jumps-jump idx 1))))))

should actually be 

	+        (when (= idx -1)
	+            (progn
	+              (setq idx 0)
	+              (setf (evil-jumps-struct-idx struct) 0)
	+              (evil--jumps-push)))
	+        (evil--jumps-jump idx 1)))))

I don't see problems with `evil--jump-forward`, but I guess it's the same.

As I'm not sure about the last one, I think I'll make the changes locally, and will try to use it for a few days. If everything is okay, I'll send my t-b.

Hi-Angel:
Alternatively somebody more acknowledgeable about Evil's internals might just say what are the changes in the `jump-forward` and the `jump-backward` for.

wasamasa:
Um, have you updated yet?  This is what I have on master:

```elisp
(defun evil--jump-backward (count)
  (let ((count (or count 1)))
    (evil-motion-loop (nil count)
      (let* ((struct (evil--jumps-get-current))
             (idx (evil-jumps-struct-idx struct)))
        (evil--jumps-message "jumping back %s" idx)
        (when (= idx -1)
          (setq idx (+ idx 1))
          (setf (evil-jumps-struct-idx struct) 0)
          (evil--jumps-push))
        (evil--jumps-jump-to-index (+ idx 1))))))
```

Hi-Angel:
@wasamasa I was referring to modifications the patch making :)

wasamasa:
Well, no idea, you'd have to ask its author about that.  If you make a PR that does better than this one, I'll close this thread in favor of it.

c4eater:
Thanks for your feedback @Hi-Angel,

Your assumption is correct. When I was using the "if" construct, evil--jumps-jump hasn't been called on first visit. One minor update to your proposed code is that I'd like to remove the progn construct as "when" makes "progn" unnecessary.

Just updated the pull request.

Hi-Angel:
Great, thank you! The series

Tested-by: Constantine Kharlamov <Hi-Angel@yandex.ru>

Please can anybody push the code?

wasamasa:
@c4eater Thanks for the update.  From quick testing nothing seems to be broken with this change, both with and without the customizable enabled.

Issue #748, opened 2017-01-27
Search causes warning about isearch-search-fun-function scope


ivanbrennan:
Upon starting Emacs in evil-mode, the first time I attempt a search (`/`), accessing the search history (`M-p`) causes a warning to display in the echo area:

> Making isearch-search-fun-function local to \*Minibuf-1\* while let-bound!

Subsequent searches are unaffected. If the first search doesn't try to access search history, it too is unaffected.

I was able to repro with a minimal init.el
```elisp
(package-initialize)
(require 'evil)
(add-hook 'after-init-hook #'evil-mode)
```
and the following steps:
 - start Emacs
 - type `/`
 - type `M-p`

If I enable `evil-mode` manually instead of via `after-init-hook`, the warning never appears. Also, if I perform a normal `isearch` before performing the evil search, the warning never appears (neither for the `isearch` nor the evil search, regardless whether I access search history).

From what I can tell, the warning originates from [make-local-variable](https://github.com/emacs-mirror/emacs/blob/65eee8392ff95f58f7b0bd036e1fe065523658c6/src/data.c#L1918) and is being triggered in [evil-search-incrementally](https://github.com/emacs-evil/evil/blob/d36020ee5f4e82cee08dbaa5a2f82e32e45195f1/evil-search.el#L77) when it [let binds](https://github.com/emacs-evil/evil/blob/d36020ee5f4e82cee08dbaa5a2f82e32e45195f1/evil-search.el#L80) `isearch-search-fun-function` to `evil-isearch-function`. I was able to prevent the warning by removing that line, but I suspect that's not an ideal fix.

Screenshot:
![screen shot 2017-01-27 at 12 31 32 am](https://cloud.githubusercontent.com/assets/1672874/22362005/44e48fac-e42d-11e6-80d0-1a18dc595e4d.png)

wasamasa:
Thanks for the bug report!  I was initially skeptical of this as I couldn't find the warning in the `*Messages*` buffer, but managed reproducing it with your minimal init file.

[...]

PR #750, opened 2017-01-28
Prevent warning when accessing search history

ivanbrennan:
Making a variable buffer-local within the BODY of a `let` generates a warning if that variable is currently let-bound (i.e. it appeared in the `let` VARLIST). Such warnings are only issued once per symbol, in order to avoid harassing the user too much.

Within evil-search-incrementally, `'isearch-search-fun-function` is let-bound, and within the body of the `let`, isearch makes that same variable buffer-local (see minibuffer-history-isearch-setup). As a result, upon starting Emacs in evil-mode, the first time you attempt a search (`/`), accessing the search history (`M-p`) causes the following warning to display in the echo area:

> Making isearch-search-fun-function local to *Minibuf-1* while let-bound!

Subsequent searches are unaffected, since these warnings are only displayed once per symbol. If the first search doesn't try to access search history, it too is unaffected.

This was reproducible with a minimal init.el

```elisp
(package-initialize)
(require 'evil)
(add-hook 'after-init-hook #'evil-mode)
```

and the following steps:
 - start Emacs
 - type `/`
 - type `M-p`

See issue #748 
#### Proposed fix

This change prevents such warnings by making `'isearch-search-fun-function` buffer-local before entering the `let` in evil-search-incrementally.

### DIFF
--- a/evil-search.el
+++ b/evil-search.el
@@ -76,6 +76,7 @@ search module is used."
 
 (defun evil-search-incrementally (forward regexp-p)
   "Search incrementally for user-entered text."
+  (make-local-variable 'isearch-search-fun-function)
   (let ((evil-search-prompt (evil-search-prompt forward))
         (isearch-search-fun-function 'evil-isearch-function)
         (point (point))


wasamasa:
Closed as there will be no more warnings in Emacs 26.

PR #755, opened 2017-02-05
Fix for append from visual state

dvzubarev:
Steps to reproduce the problem: 
1. paste to the scratch buffer 'Sample text' 
2. press `0veA11` 
3. text in the buffer: Sample 11text

Expected result: Sample11 text


TheBB:
Thanks!

Issue #731, opened 2016-12-05
Test suite fails with Emacs 25


TheBB:
Originally reported by: **Sean Whitton (Bitbucket: [spwhitton](https://bitbucket.org/spwhitton), GitHub: [spwhitton](https://github.com/spwhitton))**

----------------------------------------

* Emacs version: 25.1.1
* Evil version: commit `f2648b8`
* Operating system: Debian stretch

Several tests fails under Emacs 25:

```

10 unexpected results:
   FAILED  evil-test-insert-repeat-info
   FAILED  evil-test-normal-repeat-info-char-command
   FAILED  evil-test-normal-repeat-info-simple-command
   FAILED  evil-test-repeat
   FAILED  evil-test-repeat-append
   FAILED  evil-test-repeat-append-line
   FAILED  evil-test-repeat-insert
   FAILED  evil-test-repeat-insert-line
   FAILED  evil-test-repeat-open-above
   FAILED  evil-test-repeat-open-below
```

Full log attached.  Thanks.

----------------------------------------
- Bitbucket: https://bitbucket.org/lyro/evil/issue/731

TheBB:
At least one of these failures (probably all) are dute to https://lists.gnu.org/archive/html/bug-gnu-emacs/2016-05/msg00635.html

We need to find a workaround, I guess.

TheBB:
https://github.com/emacs-evil/evil/blob/master/evil-repeat.el#L330

Violating use of `this-command-keys` is here.

noctuid:
From reading the linked discussion, it sounds like including the prefix argument in `(this-command-keys)` is not considered the correct behavior (which is sensible), and there won't be any changes to add this behavior back. The `prefix-arg` (when non-nil) should be recorded in the `pre` section of `evil-repeat-keystrokes` to fix this.

TheBB:
Yeah, I'm just not sure to replicate that in a way that works with all the different ways users might have of specifying prefix arguments.

noctuid:
I'm not sure I follow. It doesn't matter how it is set;`prefix-arg` holds the raw value.

noctuid:
[...]

PR #757, opened 2017-02-08
Possible fix for prefix arg change in this-command-keys

justbur:
Emacs 25.1 no longer records prefix-args in this-command-keys, breaking
evil-repeat. This change tries to fix that by adding back the prefix arg when
appropriate using evil-this-command-keys.

All tests pass on my machine on Emacs 25.1 and 24.5 with this patch.

See #731

### DIFF
--- a/evil-repeat.el
+++ b/evil-repeat.el
@@ -320,6 +320,16 @@ invoked the current command"
   (clear-this-command-keys t)
   (setq evil-repeat-keys ""))
 
+(defun evil-this-command-keys (&optional post-cmd)
+  "Version of `this-command-keys' with finer control over prefix args."
+  (let ((arg (if post-cmd current-prefix-arg prefix-arg)))
+    (vconcat
+     (when (and (numberp arg)
+                ;; Only add prefix if no repeat info recorded yet
+                (null evil-repeat-info))
+       (string-to-vector (number-to-string arg)))
+     (this-single-command-keys))))
+
 (defun evil-repeat-keystrokes (flag)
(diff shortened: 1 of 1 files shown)


djstauffer:
Cool!  I'll definitely try out this patch on 25.1 and see how it works.

TheBB:
Thanks for this. The more I think about it, I would like to rewrite the repeat system to work with just commands and arguments, and no key sequences at all (at least, as few as possible). I don't know if that's possible, but it's anyway a ways off, and this is probably good enough to keep the ball rolling in the mean time. As you say, the tests all pass. Well done. :-)

justbur:
@TheBB  I understand the instinct, but I'm not sure recording commands is much easier since a command may fake that it's really another command and so forth.

wasamasa:
Looks good to me, the only nitpick I have is about a purely cosmetic thing (`prfx-arg`, really?).

justbur:
@wasamasa what would you like me to call it? I wanted to distinguish it from the global version.

wasamasa:
@justbur I'd be tempted to just shadow the global version or alternatively, call it `prefix` or even `arg`.  Sometimes when I had to call it like the original thing it's imitating, I just appended a `*` to the name, but anyways, it's a purely cosmetic thing, I'd accept the PR as is.  Just mentioning that it sticks out a bit, that's all.

justbur:
@wasamasa `arg` it is. I just force pushed by the way

TheBB:
Thanks! Merged manually.

PR #758, opened 2017-02-08
Add travis CI

TheBB:
Note that Emacs 25 segfaults when running `make test` (but it works fine when running tests in `make emacs`). I've marked them as allowed failures for this reason. At least this should help out those of us who don't want to keep 24 installed just for testing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,24 @@
+language: generic
+sudo: false
+before_install:
+  - curl -fsSkL https://gist.github.com/rejeep/ebcd57c3af83b049833b/raw > x.sh && source ./x.sh
+  - evm install $EVM_EMACS --use --skip
+  - cask --debug
+
+env:
+  matrix:
+    - EVM_EMACS=emacs-24.3-travis
+    - EVM_EMACS=emacs-24.4-travis
+    - EVM_EMACS=emacs-24.5-travis
+    - EVM_EMACS=emacs-25.1-travis
+    - EVM_EMACS=emacs-git-snapshot-travis
+
+# FIXME: Emacs 25 segfaults when running make test.
+matrix:
+  allow_failures:
+    - env: EVM_EMACS=emacs-25.1-travis
+    - env: EVM_EMACS=emacs-git-snapshot-travis
+
+script:
+  - emacs --version
+  - make test
--- a/Cask
+++ b/Cask
@@ -1,0 +1,6 @@
+(package "evil" "1.2.12" "Extensible Vi layer for Emacs.")
+
+(files "*.el")
+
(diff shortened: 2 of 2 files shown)


wasamasa:
Assuming #757 gets merged before this, the allowed failures could be removed from the config, right?

TheBB:
Oh I didn't even check if it was the tests failing that caused the segfault. I just assumed it was Emacs itself being dumb (which it still is, I guess, segfaults shouldn't happen anyway).

So, yes.

justbur:
Just confirmed that the Travis failures were due to the tests failing and that after #757 the allowed failures can be removed

TheBB:
Merged manually, removed the allowed failures.

TheBB:
Sorry, I had to add them back.

https://travis-ci.org/emacs-evil/evil/builds/199909622  
https://travis-ci.org/emacs-evil/evil/builds/199910124  
https://travis-ci.org/emacs-evil/evil/builds/199910831

justbur:
That's weird. They didn't segfault on my repo https://travis-ci.org/justbur/evil/jobs/199844954

rski:
I reproduced this locally with latest evil & Emacs-26@55e313f7bea318f03bac0a2c9db362ed361961e6
It was a debug Emacs build so assertions are on:

$ for i in `seq 1 10`; do; make test; done
~/Code/emacs/src/emacs -nw -Q -L . -L lib -l evil-tests.el \
--eval "(evil-tests-initialize '() '())"
Fatal error 6: Aborted
lisp.h:1355: Emacs fatal error: assertion failed: 0 <= nchars
Aborted (core dumped)

I wonder what test causes it

wasamasa:
Did you try adding code that prints out the test name before each run test? And yes, I agree this should be moved elsewhere.

edkolev:
Regarding the segfaults, I recently started running almost all of evil's tests with `evil-goggles-mode` enabled (https://github.com/edkolev/evil-goggles), and, to address the segfaults, I configured Travis to run the tests in emacs batch mode. I haven't seen a single segfault since, however, the downside is that some tests can't be run in batch mode, like the evil-test-command-window* and evil-test-jump tests.

Here's the related code which comments out the inappropriate tests and runs emacs in batch mode https://github.com/edkolev/evil-goggles/blob/master/Makefile#L19-L27

wasamasa:
Did you try running them in non-windowed mode? My last state on this is that some glyph matrix stuff in the terminal specific code causes the segfault and it only happens after running quite a lot of tests.

edkolev:
Initially I tried running them exactly like evil did, but it didn't work out due to the segfault. If I'm not mistaken evil runs them in non-windowed mode `-nw`

PR #754, opened 2017-02-01
Fix repo url in commentary section

therockmandolinist:
#753

Issue #753, opened 2017-01-31
fix melpa evil description


ninrod:
the description of evil in https://melpa.org/#/evil says this:

```txt
...
Evil lives in a Git repository. To obtain Evil, do

     git clone git://gitorious.org/evil/evil.git
...
```

Update to reflect Evil's new home.

therockmandolinist:
This actually comes from the evil.el file, in the `;;; Commentary:` section. A PR may be more apt.

wasamasa:
This should hopefully be updated in a few hours, but I've had issues before over cached descriptions.  If that happens, we'll have to prod MELPA.

TheBB:
Looks good now.

wasamasa:
MELPA Stable does still show the old one.  Perhaps that will go away once every release has been tagged and a new one is out?

TheBB:
Yeah it should. If everyone are ok with https://github.com/emacs-evil/evil/pull/757 we should merge it and tag a 1.3.

PR #759, opened 2017-02-08
further update information about evil's new home.

ninrod:
This is a follow up of #753 and #754. Corrected other locations that needed to be updated to reflect evil's new home, bug tracking location and mantainer information.


TheBB:
Thank you!

wasamasa:
There's a few more in the repo, like in `evil.texi` (gitorious) and `scripts/evilupdate` (bitbucket).  The latter is something @TheBB might want to look into *after* releasing 1.3.

ninrod:
Sure thing, I'll later hunt them down!

PR #761, opened 2017-02-09
further update repository information in docs and scripts directories

ninrod:
following up on @wasamasa's [comment](https://github.com/emacs-evil/evil/pull/759#issuecomment-278587741), this further complements #753, #754 and #759 updating additional instances of official repository locations found in docs and scripts directories.


wasamasa:
Thanks!  That should be hopefully all instances.

PR #762, opened 2017-02-09
Add missing mention to goto-cgh in the appropriate README.md section

ninrod:
The readme mentions `undo-tree.el` as a dependency in the installation section, but misses stating the final dependency which is `goto-chg.el`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,8 +23,14 @@ Move Evil to `~/.emacs.d/evil`. Then add the following lines to `~/.emacs`:
 (evil-mode 1)
 ```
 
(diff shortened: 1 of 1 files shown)


edkolev:
Aren't the dependencies auto installed when evil is installed from melpa? If they are, maybe it's not necessary to list them in the readme.

TheBB:
They are, but even though MELPA may have become the de facto standard for packaging, this repo is still upstream from it, and not to be identified with it.

TheBB:
Thanks!

Issue #765, opened 2017-02-11
abbrev.el integration


edkolev:
I noticed that abbrevs aren't expanded when leaving insert state. For example, with `undefined` abbreviated as `undef` (for haskell):
- typing this expands the abbrev: `undef<SPC>` => `undefined`
- typing this doesn't expand `undef<ESC>` => `undef`

Vim's behaviour is to expand abbrevs when leaving insert mode.

I believe this should fix it in evil-integration.el:
```
(with-eval-after-load 'abbrev
  (add-hook 'evil-insert-state-exit-hook 'expand-abbrev))
```

I'll create a PR if this is the right behaviour and the right way to implement it.

therockmandolinist:
I think it would be nice to implement this as a user option (sure, it could default to `t`), since on first thought this sounds a bit odd to me, but if it's vim's behavior I can see why it should be included.

PR #766, opened 2017-02-12
abbrev.el integration - add customizable variable `evil-want-abbrev-expand-on-insert-exit`.

edkolev:
When non-nil (which is default) the `expand-abbrev` will be called in
a `evil-insert-state-exit-hook` hook. This is in-line with Vim's behaviour.

Refs #765

### DIFF
--- a/evil-integration.el
+++ b/evil-integration.el
@@ -552,4 +552,9 @@ Based on `evil-enclose-ace-jump-for-motion'."
 
 (provide 'evil-integration)
 
+;;; abbrev.el
+(when evil-want-abbrev-expand-on-insert-exit
+  (eval-after-load 'abbrev
+    '(add-hook 'evil-insert-state-exit-hook 'expand-abbrev)))
+
 ;;; evil-integration.el ends here
--- a/evil-tests.el
+++ b/evil-tests.el
@@ -8184,6 +8184,25 @@ maybe we need one line more with some text\n")
         ("3\C-i") ;; even after jumping forward 3 times it can't get past the 3rd z
         "z z [z] z z z z z"))))
 
+(ert-deftest evil-test-abbrev-expand ()
+  :tags '(evil abbrev)
+  (ert-info ("Test abbrev expansion on insert state exit")
+    (define-abbrev-table 'global-abbrev-table
+      '(("undef" "undefined"))) ;; add global abbrev
+    (evil-test-buffer
+     "foo unde[f] bar"
+     ("a" [escape])
+     "foo undefine[d] bar") ;; 'undef' should be expanded
+    (evil-test-buffer
+     "fo[o] undef bar"
+     ("a" [escape])
+     "fo[o] undef bar") ;; 'foo' shouldn't be expanded, it's not an abbrev
+    (kill-all-abbrevs) ;; remove abbrevs
+    (evil-test-buffer
+     "foo unde[f] bar"
+     ("a" [escape])
+     "foo unde[f] bar"))) ;; 'undef' is not an abbrev, shouldn't be expanded
+
 (provide 'evil-tests)
 
 ;;; evil-tests.el ends here
--- a/evil-vars.el
+++ b/evil-vars.el
@@ -1180,6 +1180,12 @@ command is non-zero."
   :type 'boolean
   :group 'evil)
 
+(defcustom evil-want-abbrev-expand-on-insert-exit t
+  "If non-nil abbrevs will be expanded when leaving Insert state
+like in Vim. This variable is read only on load."
+  :type 'boolean
+  :group 'evil)
+
 ;;; Variables
 
 (defmacro evil-define-local-var (symbol &optional initvalue docstring)


edkolev:
I'm not sure `evil-want-abbrev-expand-on-insert-exit` is an appropriate name - some of the  other customizable vars have `*want*` in their name, some don't.

Also, changing the variable after evil-intregration.el has been loaded will not take effect (similar to `evil-search-module). Should that be mentioned in the documentation?

TheBB:
I think the name is fine. It's probably good to mention in the docstring that the variable is read on load.

Most notably, please use `eval-after-load` instead of `with-eval-after-load` on account of Emacs 24.3 and older.

edkolev:
Done. Let me know in case something else needs to be modified.

TheBB:
I like it, merged!

Had to `(setq abbrevs-changed nil)` to prevent Emacs from wanting to save abbrevs after running tests.

rgrinberg:
This is a good setting but isn't entirely usable because you cannot set it after loading evil. It would be nice if this setting was changeable at any time. https://github.com/ProofGeneral/PG/issues/174#issuecomment-296499547

wasamasa:
If you want to apply that change at a later time for whatever reason (why wouldn't you want customizations to apply as early as possible?), just take a look at its body.  All that customizable does is adding a hook...

rgrinberg:
Per the linked comment, spacemacs has a special place for setting these variables that unfortunately ends up being too late for evil to pick it up. This isn't technical justification, but it does mean that toggling this feature is inconvenient for a big chunk of users.

wasamasa:
All the more reason for Spacemacs to fix this because Evil is just one of many packages requiring that kind of setup.  A plain old init file allows for this just fine, even one with `use-package` works for this purpose.

edkolev:
I could extend the `defcustom` to allow toggling via the customize functions:
```emacs-lisp
(custom-set-variables
 '(evil-want-abbrev-expand-on-insert-exit t))

(custom-set-variables
 '(evil-want-abbrev-expand-on-insert-exit nil))
```

The diff needed for this to be possible is:
```diff
diff --git a/evil-integration.el b/evil-integration.el
index 79558ea..a6b5f96 100644
--- a/evil-integration.el
+++ b/evil-integration.el
@@ -552,9 +552,4 @@ Based on `evil-enclose-ace-jump-for-motion'."

 (provide 'evil-integration)

-;;; abbrev.el
-(when evil-want-abbrev-expand-on-insert-exit
-  (eval-after-load 'abbrev
-    '(add-hook 'evil-insert-state-exit-hook 'expand-abbrev)))
-
 ;;; evil-integration.el ends here
diff --git a/evil-vars.el b/evil-vars.el
index a5c8ddd..0cd3117 100644
--- a/evil-vars.el
+++ b/evil-vars.el
@@ -1191,7 +1191,12 @@ command is non-zero."
   "If non-nil abbrevs will be expanded when leaving Insert state
 like in Vim. This variable is read only on load."
   :type 'boolean
-  :group 'evil)
+  :group 'evil
+  :set #'(lambda (sym value)
+           (set-default sym value)
+           (if value
+               (add-hook 'evil-insert-state-exit-hook 'expand-abbrev)
+             (remove-hook 'evil-insert-state-exit-hook 'expand-abbrev))))

 ;;; Variables
```

Note: I'm not sure if I need to wrap the code in `:set` with `(eval-after-load 'abbrev ...)`.

@rgrinberg will this solve the problem?

edkolev:
@rgrinberg I'm not sure I was clear above with the diff pasted, so let me re-phrase: is this what your're asking for:
- `(custom-set-variables  '(evil-want-abbrev-expand-on-insert-exit t))` to enable
- `(custom-set-variables '(evil-want-abbrev-expand-on-insert-exit nil))` to disable

rgrinberg:
I think so. Will it work with setq as well?

edkolev:
setq will work only if evil hasn't been loaded yet. After it's been loaded, the  custom-set-variables function will be able to toggle.

A lot of evil customization options work like this.

cpitclaudel:
I think I'm missing something here… this looks like a complicated solution to a simple problem :/ @edkolev , would the naive solution of just checking for that variable in the hook not work? That is, can't `evil-insert-state-exit-hook` just start by checking the value of that variable?

> A lot of evil customization options work like this.

Really? But you can't set these buffer-locally, nor change them easily without restarting Emacs, nor temporarily toggle them, etc…

edkolev:
> That is, can't evil-insert-state-exit-hook just start by checking the value of that variable?

No, `evil-insert-state-exit-hook` is a variable. `evil-want-abbrev-expand-on-insert-exit` is also a variable. We need some "code" to modify the former when the latter is changed. The customize functions allow us to do so (this is my proposal above).

I can submit a PR with my proposal, but can someone confirm this will solve the problem?

cpitclaudel:
> No, evil-insert-state-exit-hook is a variable. 

Sorry I wasn't clear. I'm suggesting to define an `evil-maybe-expand-abbrev-after-insert-state-exit` function, and unconditionally add that to `evil-insert-state-exit-hook`.  That function would call `expand-abbrev` if `evil-want-abbrev-expand-on-insert-exit` is non-nil, and do nothing otherwise.

wasamasa:
Considering how rarely you change settings and how much more you make use of a package's commands, I don't see any reason for complicating this any further.  If you don't want to use the customizable, just put the code it runs into your init file.  In this case it's a matter of `(add-hook 'evil-insert-state-exit-hook 'expand-abbrev)`, so I don't get why you keep discussing this.

cpitclaudel:
>  I don't get why you keep discussing this.

Uh? I didn't comment on this until today.

Issue #718, opened 2016-10-03
Create [s and ]s motions for ispell


TheBB:
Originally reported by: **Anonymous**

----------------------------------------

The thing I miss about vim is how easy it is to use spellcheck. With evil, to navigate to a misspelled word I have to press `b` until I reach the word. I really miss `[s` to travel to the last mispelled word, fixing it, and then returning to the last position of my cursor using `gi`.

----------------------------------------
- Bitbucket: https://bitbucket.org/lyro/evil/issue/718

wasamasa:
`]s` can be implemented in terms of `flyspell-goto-next-error`, however this is not perfect for a number of reasons:

- Requires `M-x flyspell-buffer` to work correctly (which is OK as Vim also requires previous setup and errors out if you try using the aforementioned commands before)
- `z=` is using an ispell command, so that's sort of inconsistent
- No built-in equivalent for going the other direction
- Wrap-around behavior is different, in Vim it wraps around if you're going beyond the last/first word and the appropriate option is set, with Flyspell you get an irritating "No more miss-spelled word!" message (which is just wrong), only after that it wraps around

Writing my own commands based on some basic flyspell functions shouldn't pose much effort, so I'm giving that a try.

PR #768, opened 2017-02-13
Implement ]s and [s

wasamasa:
See #718.  This is a start and doesn't behave quite like Vim yet.  In Vim I don't get wraparound for `]s`, but for `[s`, there is no error when using it on a buffer without spelling mistakes and wraparound is customizable (`:help 'wrapscan'`).

edit: Looks like there's neither `ispell` nor `aspell` nor `hunspell` installed in the testing environment...


wasamasa:
I consider this ready for use now.  While it still doesn't handle Vim's full range of spell checking (it knows bad and rare words as well...), it otherwise works as described in the linked ticket.

@TheBB Could you please test this to make sure I haven't overlooked any strange corner cases?

wasamasa:
I consider this ready for use now.  While it still doesn't handle Vim's full range of spell checking (it knows bad and rare words as well...), it otherwise works as described in the linked ticket.

@TheBB Could you please test this to make sure I haven't overlooked any strange corner cases?

TheBB:
This looks okay to me. Your dedication to replicating the vim behaviour is commendable, but I'm not convinced it's worth it to fly around looking for overlays instead of just using the higher level flyspell API and screw the differences.

Merged, at any rate. :-)

wasamasa:
Flyspell doesn't provide a complete API, so I replicated it and added the missing parts.

PR #770, opened 2017-02-14
Require cl-lib instead of cl

wasamasa:
As parts of Evil have already been using `cl-loop` from `cl-lib` while others still relied on unprefixed `cl.el` functionality (more specifically, `incf`), I've taken the liberty to remove all `cl.el` usage.  A side effect of this is that Evil depends on `cl-lib` to accomodate for older Emacsen (24.1 and 24.2?) that don't have it included, for them the dependency can be installed from GNU ELPA.  This also means that code specific to Emacs 23 should be deleted as Emacs 23 isn't compatible with cl-lib and we don't test it anyways.

The biggest change is the dropped `lexical-let` from `evil-select-paren`.  I've thought for quite a while about why one would want to bind a variable to its value and came up with two cases:

1. The body of the let mutates one of the bindings which happen to refer to dynamically bound variables outside (most common example: `(let ((this-command this-command)) <code potentially changing this-command>)`).
2. The body of the let contains lambdas that close over free variables involving one of its bindings (think event handlers and alike).  This always works with lexical binding enabled as closures are created for that case which capture the free variables with their current values, creating an environment permanently attached to the lambda.  With dynamic binding, it depends.  If you immediately call the lambda, the free variables can be looked up in the calling frames and resolved meaningfully.  Calling it at a later point in time (like with `run-at-time` or from `url-retrieve`) will fail, unless you go for:
- lexical binding, be it real (by enabling it for the buffer with a file variable prop line) or emulated (with `lexical-let` from `cl.el`)
- a backquoted lambda, with the free variables unquoted into it, giving you literals instead of a mutable environment

Clearly this happens to be case 2 as there are two sharp-quoted lambdas referring to the let-bound variables.  As the lambda is called immediately, the call stack contains them, so neither fake lexical binding nor backquoted lambdas are needed.  This is why I've removed the `lexical-let` altogether.  Even better, if we ever switch to true lexical binding, the lambdas will capture the arguments, so the code will continue working.

### DIFF
--- a/evil-common.el
+++ b/evil-common.el
@@ -3208,52 +3208,50 @@ must be regular expressions and `evil-up-block' is used.
 If the selection is exclusive, whitespace at the end or at the
 beginning of the selection until the end-of-line or beginning-of-line
 is ignored."
-  (lexical-let
-      ((open open) (close close))
-    ;; we need special linewise exclusive selection
-    (unless inclusive (setq inclusive 'exclusive-line))
-    (cond
-     ((and (characterp open) (characterp close))
-      (let ((thing #'(lambda (&optional cnt)
-                       (evil-up-paren open close cnt)))
-            (bnd (or (bounds-of-thing-at-point 'evil-string)
-                     (bounds-of-thing-at-point 'evil-comment)
-                     ;; If point is at the opening quote of a string,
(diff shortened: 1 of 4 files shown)


TheBB:
Thanks for this!

PR #771, opened 2017-02-15
Remove support for Emacs < 24

TheBB:
The logical continuation of #770

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -32,10 +32,6 @@
 (require 'evil-command-window)
 (require 'evil-jumps)
 
-;;; Compatibility for Emacs 23
-(unless (fboundp 'window-body-width)
-  (defalias 'window-body-width 'window-width))
-
 ;;; Motions
 
 ;; Movement commands, or motions, are defined with the macro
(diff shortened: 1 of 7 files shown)


wasamasa:
Excellent, thanks!

bmag:
Seems to have broken 3rd party packages that rely on, e.g. evil-nerd-commenter. (see https://gitter.im/syl20bnr/spacemacs?at=58a980ddaa800ee52cb60776)
I didn't check what else breaks

wasamasa:
@bmag OK, that is indeed silly.  Why does that package even use Evil's wrapper of `called-interactively-p`?  It's not like it's trying to paper over Emacs 23 idiosyncracies...

wasamasa:
According to [a simple search](https://github.com/search?q=evil-called-interactively-p&type=Code), this broke a few more:

- https://github.com/tarao/evil-plugins/blob/master/evil-operator-comment.el
- https://github.com/Dewdrops/evil-exchange/blob/master/evil-exchange.el
- https://github.com/roman/evil-paredit/blob/master/evil-paredit.el

TheBB:
We can add back the functions with a deprecation warning perhaps?

wasamasa:
Yeah, that would work, though I've considered handing in PRs to these packages instead.

TheBB:
Okay, well I wrote https://github.com/emacs-evil/evil/pull/776 in the meantime.

Issue #734, opened 2016-12-21
Scrolling with C-e and C-y doesn't scroll cursor


TheBB:
Originally reported by: **Anonymous**

----------------------------------------

Scrolling with C-e and C-y in Vim also scrolls the cursor while in Evil it stays in the same place visually (but not contextually).  So the behaviour is different from Vim.

* Emacs version: 24.4.1
* Evil version: 1.0-dev
* OS: Linux «host» 4.2.0-0.bpo.1-amd64 #1 SMP Debian 4.2.6-3~bpo8+2 (2015-12-14) x86_64 GNU/Linux

----------------------------------------
- Bitbucket: https://bitbucket.org/lyro/evil/issue/734

jojojames:
You can set scroll-preserve-screen-position to nil to get evil's functionality.

The problem is I'd like to get something like this for 'smoother' scrolling in Emacs.

```elisp
;; https://stackoverflow.com/questions/3631220/fix-to-get-smooth-scrolling-in-emacs
(setq scroll-margin 5
      scroll-step 1
      scroll-conservatively 10000
      scroll-preserve-screen-position 1) ;; <----
```

```elisp
(evil-define-command evil-scroll-line-up (count)
  "Scrolls the window COUNT lines upwards.
If COUNT is not specified the function uses
`evil-scroll-line-count', which is the last used count."
  :repeat nil
  :keep-visual t
  (interactive "<c>")
  (progn
    (setq count (or count evil-scroll-line-count))
    (setq evil-scroll-line-count count)
    (scroll-down count)))

(evil-define-command evil-scroll-line-down (count)
  "Scrolls the window COUNT lines downwards.
If COUNT is not specified the function uses
`evil-scroll-line-count', which is the last used count."
  :repeat nil
  :keep-visual t
  (interactive "<c>")
  (progn
    (setq count (or count evil-scroll-line-count))
    (setq evil-scroll-line-count count)
    (scroll-up count)))
```

I wonder if it would be appropriate to just set it to nil before calling scroll-up/scroll-down and resetting it to its original value after the fact.

Something like this,

```elisp
  (evil-define-command evil-scroll-line-up (count)
    "Scrolls the window COUNT lines upwards.
If COUNT is not specified the function uses
`evil-scroll-line-count', which is the last used count."
    :repeat nil
    :keep-visual t
    (interactive "<c>")
    (progn
      (setq count (or count evil-scroll-line-count))
      (setq evil-scroll-line-count count)
[...]

PR #774, opened 2017-02-18
Match C-e/C-y scrolling behavior with Vim's

jojojames:
`scroll-preserve-screen-position' is a user settable variable that
controls scroll-up/scroll-down's behavior.

Set it to nil before calling scroll-* so that the cursor stays on the
same line it was already on after scrolling.

https://github.com/emacs-evil/evil/issues/734


wasamasa:
Thanks!

PR #785, opened 2017-02-26
Look up bound key status correctly

wasamasa:
A user reported on IRC that they failed at customizing Evil to *not* join lines when deleting with the backspace key.  This appears to be the fault of code in `evil-maps.el` which goes through a list of mappings, then checks whether they're bound and what they're bound to before removing or adding a binding.  While this works perfectly for nearly all cases, it fails on remapping definitions such as `([remap delete-backward-char] . evil-delete-backward-char-and-join)` as Emacs signals an in-band error by returning a number as definition.

I've taken the liberty of introducing a helper function that handles this case.  Perhaps we should look into using it in other places that go for regular `lookup-key` after merging this PR.

### DIFF
--- a/evil-common.el
+++ b/evil-common.el
@@ -743,6 +743,13 @@ filename."
   (when prompt
     (setcdr map (cons prompt (cdr map)))))
 
+(defun evil-lookup-key (map key)
+  "Returns non-nil value if KEY is bound in MAP."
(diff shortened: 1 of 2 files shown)


ivanbrennan:
This fixes #799

TheBB:
Merged, thanks!

Issue #760, opened 2017-02-08
join and delete ex commands don't take count argument


quaker4lyf:
Example lines in my file:
```
pattern
  a
  b

pattern
  c
  d

pattern
  e
  f
```
`:%g/pattern/j 3` in vim it produces what I want:
```
pattern a b

pattern c d

pattern e f
```
whereas in spacemacs (using evil) it only joins 2 lines (which is the same as ignoring the count):
```
pattern a
  b

pattern c
  d

pattern e
  f
```
Similar thing with the `:delete` command.
`:%g/pattern/d 3` in vim leaves only the empty lines:
```


```
whereas in spacemacs it deletes only the lines matching the pattern
```
  a
  b

  c
  d

  e
  f
```
I figure the bug has to do with evil hence filing it here.

Emacs version: GNU Emacs 25.1.91.1
Spacemacs version: spacemacs 0.200.5@25.1.91.1
(Please let me know if you need the Evil version)

edkolev:
Since #818 is now merged, this issue can be closed.

wasamasa:
Good call!  BTW, putting a `Fixes #818` or `Closes #818` line into the commit message causes the linked issue to be closed as soon as the MR gets merged.

edkolev:
I see, thanks!

PR #786, opened 2017-03-03
Add support for passing COUNT and BANG to :join

edkolev:
This tries to fix one of the reported issues in #760 (the other one is :delete)

Also, `:join` now supports bang `!` - it was simple to add.

Edit: the second commit tries to emulate vim's behaviour.
~~However, I failed emulating vim's behaviour when both COUNT and visual range are given (). Hence I left this corner-case unchanged (seems like a corner case to *me*, but maybe some people are regularly using it) - the COUNT will be ignored if `region-active-p` is true. I personally don't find vim's behaviour intuitive, seems like they had to chose some default and went with it.~~

Let me know if the PR can be improved in any way.

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -1515,6 +1515,7 @@ of the block."
   (let ((count (count-lines beg end)))
     (when (> count 1)
       (setq count (1- count)))
+    (goto-char beg)
     (dotimes (var count)
       (join-line 1))))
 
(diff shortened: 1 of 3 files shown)


edkolev:
I've added a second commit which emulates vim's behaviour when both COUNT and RANGE are given.

Let me know if they should be squashed or could be improved in any other way.

edkolev:
I forgot to mention: I added `(goto-char beg)` in `evil-join` & `evil-join-whitespace` because they were _assuming_ the point is on BEG; they just counted how many lines are b/w BEG and END and joined that many times.

TheBB:
Thanks! This looks good. Merged.

edkolev:
Great :)

I could try to add support for COUNT in `:delete`, but that would be a bit more complex because COUNT can either be the first or second argument (register `[x]` is optional):
```
# from vim's ':help :delete'
:[range]d[elete] [x] {count}
```

I'll probably first add support for COUNT, then for `[x]`

Issue #699, opened 2016-08-10
could you add an option evil-want-P-move-cursor-back


TheBB:
Originally reported by: **Ji Xiufeng (Bitbucket: [jixiuf](https://bitbucket.org/jixiuf), GitHub: [jixiuf](https://github.com/jixiuf))**

----------------------------------------

It seems that evil-paste-before does not respect evil-move-cursor-back

any reason?

if you do not want make  evil-paste-before  respect evil-move-cursor-back
could you add an option evil-want-P-move-cursor-back or add another command 
like "gP" in vim 

is seems the difference between P and gP is: gP never move back cursor

and I found evil-mode does not have "gp" and "gP" 

----------------------------------------
- Bitbucket: https://bitbucket.org/lyro/evil/issue/699

jixiuf:
after #787 
this can be closed now .

wasamasa:
Alright!

PR #787, opened 2017-03-06
make  evil-paste-before respect evil-move-cursor-back

jixiuf:
#699


wasamasa:
Thanks, looks good to me.  If you want to implement `gp` and `gP`, feel free to open a separate issue for them.

Issue #794, opened 2017-03-16
copy-from-{above,below} doesn't respect replace-state


ivanbrennan:
In vim, copying from above (C-y) or below (C-e) while in replace mode does what you'd expect: replace the current character with the character directly above/below it. In evil-replace-state, however, calling evil-copy-from-{above,below} *inserts* the character, pushing subsequent text over to make room. This defeats the purpose of replace mode, and is actually a workflow I use frequently.

To repro, start emacs with this minimal init file:
```emacs-lisp
(require 'package)
(setq package-archives '(("gnu"   . "https://elpa.gnu.org/packages/")
                         ("melpa" . "https://melpa.org/packages/")))
(package-initialize)
(require 'evil)
(evil-mode)
```

Insert the following text in the scratch buffer:
```emacs-lisp
'abc
'xyz
```

Then, with point before the *x* in *xyz*, enter evil-replace-state and call evil-copy-from-above. The result:
```emacs-lisp
'abc
'axyz
```

ivanbrennan:
I've written a quick and dirty workaround, but I'm not familiar with the internals of evil-replace-state so I don't know that it covers all cases. It has worked for me so far, though.

```emacs-lisp
(defun evil-consume-next-char-if-replacing (&rest _args)
  (when (evil-replace-state-p)
    (delete-char 1)))

(advice-add 'evil-copy-from-above :after #'evil-consume-next-char-if-replacing)
(advice-add 'evil-copy-from-below :after #'evil-consume-next-char-if-replacing)
```

wasamasa:
Ugh, you've reminded me that `self-insert-command` [is special](http://emacshorrors.com/posts/self-insert-command.html).  One of the special things it does is implementing `overwrite-mode` which is what Evil reuses for replace state.  So, unless Evil uses `self-insert-command` to insert the character from above, replace state won't work perfectly.

Not sure whether it's better to reinvent it (and keep using `insert`) or to coerce these commands into using `self-insert-command` to handle this perfectly.

ninrod:
Good one. I did not even know that this existed.

PR #795, opened 2017-03-17
Correct C-e and C-y in replace state

wasamasa:
See #794.

Notes:

- This doesn't handle a sequence like `R C-3 i <backspace> <backspace> <backspace>` correctly.  Note how repetition of insertion works as expected (I've fixed a case of swapped arguments for this), but undoing doesn't.  While this peeves me greatly, Vim doesn't even allow specifying a prefix arg in replace state, so that's fine I guess.
- I've noticed that `C-e` in replace state runs the appropriate Emacs command, not the copying command.  To make the example from the linked issue work out of the box, `evil-replace-state-map` would need to get augmented, then the respective lines in the tests could be omitted.  I don't think that's part of the PR though...

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -2306,6 +2306,13 @@ next VCOUNT - 1 lines below the current one."
                    (setq row nil)))
         rows))))
 
+(defun evil--self-insert-string (string)
+  "Insert STRING as if typed interactively."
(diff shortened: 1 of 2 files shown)


TheBB:
Thanks for this! Love the tests.

PR #798, opened 2017-03-21
Update evil-use-register: show "- in the minibuffer

duianto:
problem:  pressing double quote, doesn't show anything in the minibuffer
solution: copy the single quote and backtick behaviour

I don't know what the angle brackets around the uppercase `C` in `(interactive "<C>")` do? So I'm not sure if there is any lost of functionality when the interactive command is changed. But copying and pasting to and from a register still works as expected.

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -1899,7 +1899,7 @@ The return value is the yanked text."
   "Use REGISTER for the next command."
   :keep-visual t
   :repeat ignore
-  (interactive "<C>")
+  (interactive (list (read-char)))
   (setq evil-this-register register))
 
 (defvar evil-macro-buffer nil


TheBB:
`<C>` is one of the custom evil interactive forms.

https://github.com/emacs-evil/evil/blob/master/evil-types.el#L277-L282

wasamasa:
I guess the reason this is here is to support registers using non-ASCII letters as interpreted by an Emacs input method.  However I didn't manage to prove this hypothesis, probably because normal state (which Emacs is in when " is interpreted) doesn't take input methods into account, so I'm a bit puzzled what its purpose is.

duianto:
Sorry for being unclear. I made this PR because there's no feedback when the `"` (double quote) key binding `S-'` or the following register key is pressed. That makes it unclear if one actually pressed the keys. 

The current behaviour also differ from Vim, where both `"` and the following register key is shown.

duianto:
While i was trying to get the `f`, `F`, `t` and `T` keys to also show feedback in the minibuffer, like they do in Vim. 

I discovered that it seems to be enough to just call: `(interactive "c")` instead of 
`(interactive (list (read-char)))` to get the pressed key and the following character to show up in the minibuffer.

I tested the `evil-use-register` function with "c" and it seems to work the same way as `(list (read-char))` (that's currently used in this PR). 

Should I update this PR to use "c" instead?

wasamasa:
I've researched this a bit further and came to the conclusion that this PR is solving a minor annoyance while disallowing Evil to read in a key with an Emacs input method (currently broken) or under inclusion of `evil-read-key-map` (which allows using `C-q` for quoted input, without that it's impossible to search forward to special characters or abort with `ESC`).  Therefore I cannot accept it.

Evil does disable input methods in most of its states with the exception of Emacs and insert state.    You can imagine this as if it were using `toggle-input-method` when appropriate so that you don't have to use it as often.  Now, while it's useful that it ensures you will not accidentally type japanese in normal state, there are normal state commands where it makes sense to read in a key as if your input method were active, such as `f` where one might want to jump to a japanese glyph.  Currently this doesn't appear to work at all, I'd welcome a PR fixing this.

duianto:
Is there a risk that the same input issue currently is occurring in these functions: `evil-goto-mark`, 
`evil-goto-mark-line` and `evil-set-marker` because all of them call: `(interactive (list (read-char)))`.

Should they be changed to use `(interactive "<C>")` like `evil-use-register` currently does?

wasamasa:
No.  Marks are defined to use an alphanumerical single-char identifier, so that wouldn't make any sense.

PR #806, opened 2017-04-10
:global should use last pattern if none given

edkolev:
With this change, you can for example:
1. search for foo `/foo`
2. delete all lines with foo with `:g//d`

Without this commit, `:g//d` would delete *all* lines in the buffer since the empty `""` pattern matches on every line.

Please, let me know if the PR can be improved.

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -3502,6 +3502,9 @@ This is the same as :%s//~/&"
     (user-error "No pattern given"))
   (unless command
     (user-error "No command given"))
+  ;; TODO: `evil-ex-make-substitute-pattern' should be executed so
+  ;; :substitute can re-use :global's pattern depending on its `r'
+  ;; flag. This isn't supported currently but should be simple to add
   (evil-with-single-undo
     (let ((case-fold-search
            (eq (evil-ex-regex-case pattern 'smart) 'insensitive))
--- a/evil-search.el
+++ b/evil-search.el
@@ -1204,7 +1204,19 @@ This handler highlights the pattern of the current substitution."
 
 (defun evil-ex-parse-global (string)
   "Parse STRING as a global argument."
-  (evil-delimited-arguments string 2))
+  (let* ((args (evil-delimited-arguments string 2))
+         (pattern (nth 0 args))
+         (command (nth 1 args)))
+    ;; use last pattern if none given
+    (when (zerop (length pattern))
+      (setq pattern
+            (cond
+             ((and (eq evil-search-module 'evil-search) evil-ex-search-pattern)
+              (evil-ex-pattern-regex evil-ex-search-pattern))
+             ((and (eq evil-search-module 'isearch) (not (zerop (length isearch-string))))
+              isearch-string)
+             (t (user-error "No previous pattern")))))
+    (list pattern command)))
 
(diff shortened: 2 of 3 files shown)


wasamasa:
- Use `when` instead of `if` without an else clause
- I'd signal an `user-error` if there is no last pattern.  Vim confusingly prints "Pattern not found:" instead.

edkolev:
Thanks. I've updated the commit. Again, let know if the code can be improved.

wasamasa:
I've just tested this and noticed it doesn't pick up the last search because nothing is stored in `evil-ex-search-pattern`.  I guess this is due to me still not using the `evil-search` search module.  I guess it would be fair to make this work the same way for the default, simply because `:s//` works the same way.  Maybe you can look at that one to see how it's implemented?

edkolev:
Thanks for looking into this and catching the issue.

I guess you use `isearch` search module? I've updated the test to cover `isearch` as well, and updated the code to cover this case. I'm not sure how to refactor out the duplicated code into a separate function because `evil-ex-get-substitute-info` takes its input flag `r` into account.

Looking at the `r` flag in vim's docs, I see this:
```
[r]     ... When the search pattern is empty, use the
        previously used search pattern instead of the search pattern from the
        last substitute or :global. ...
```

Seems to me that running `:global` should modify the following var, otherwise vim's behavior will not be correctly emulated:
```
(defvar evil-ex-substitute-pattern nil
  "The last substitute pattern.")
```
I'll try to better understand vim's behaviour in this case.

edkolev:
Seems to me that the evil-ex-global function should just do this:
```
(setq pattern (evil-ex-make-substitute-pattern pattern ""))
```

So the next time `:s///` is called it can use :global's pattern (depending on the `r` flag).
However, I couldn't come up with a test case for this scenario at all. Looking at vim's docs, I see that the `r` flag is On if you use `:s//red/` or `:~` or `:&r` but it's not On with `:&`. From `:help :s_flags`:
```
        For :s with an argument this already happens:
                :s/blue/red/
                /green
                :s//red/   or  :~   or  :&r
        The last commands will replace "green" with "red".
                :s/blue/red/
                /green
                :&
        The last command will replace "blue" with "red".
```

I don't feel comfortable trying to emulate vim's `r` flag since I don't understand it that well. Maybe we should not focus on this flag now, not until someone reports a real use case that I can work on?

wasamasa:
Sure, just put a note into the code that the flag is deliberately unsupported.

edkolev:
The PR is now modified with the following comment in `evil-ex-global`:
```
  ;; here `evil-ex-make-substitute-pattern' should be executed so that
  ;; :substitute can re-use :global's pattern depending on its `r'
  ;; flag. This isn't supported currently but should be simple to add
```

Let me know in case I should modify this or add some tag like TODO/Note/XXX.

wasamasa:
Pick any prefix you like. Without one, it's hard to tell at first glance whether the comment explains the following code or is a remark what ought to be fixed about it.

edkolev:
I added `TODO` since it's already used in other places in the file.

wasamasa:
Great, thanks!

edkolev:
Thanks!

PR #808, opened 2017-04-10
Use `save-excursion` to preserve point

### LINKED ISSUE #807: `c i "` does not work outside of function definitions in go mode
I can reproduce the error with this go file 

```go
package x

const AUTH_HEADER_PREFIX = "X-Auth-"
```

Pressing `c i "` inside the string does not work. 

If i remove those to lines from the go-mode implementation the error disappears:
```elisp
(set (make-local-variable 'beginning-of-defun-function) #'go-beginning-of-defun)                                                                                                                                                                            
(set (make-local-variable 'end-of-defun-function) #'go-end-of-defun) 
```

I tested `c i "` inside a go function definition and i works without problems.

A comparison with the rust `beginning-of-defun` implementation shows that rust-mode moves to the beginnging of the line outside of a function definition while go-mode moves to the end of the line. In rust `c i "` works fine outside of function definitions.

wasamasa:
Closes #807

### DIFF
--- a/evil-common.el
+++ b/evil-common.el
@@ -1529,8 +1529,9 @@ backwards."
         (setq reset-parser t))
       ;; global parser state is out of state, use local one
       (let* ((pnt (point))
-             (state (progn (beginning-of-defun)
-                           (parse-partial-sexp (point) pnt nil nil (syntax-ppss))))
+             (state (save-excursion
+                      (beginning-of-defun)
+                      (parse-partial-sexp (point) pnt nil nil (syntax-ppss))))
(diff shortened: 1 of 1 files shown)


mriehl:
Any progress on this? It fixes an existing issue that is very annoying when using evil in conjunction with the golang mode.

TheBB:
@mriehl I'll merge some PRs this week. I've been moving and only recently that's been resolved.

@wasamasa Maybe it'd be reasonable just to abolish this "two person" rule? I don't know.

PR #818, opened 2017-04-28
Add support for passing REGISTER and COUNT to :delete

edkolev:
With this change, `:delete` can be called like so:
- `:d 2` to del 2 lines
- `:d x` to del the line and put the text into register `x`
- `:d x 2` to del 2 lines into register `x`

Passing COUNT while in visual mode works a bit strange in my opinion, but the behaviour should be the same as in vim.

Also, I'm not sure function `evil-ex-get-delete-info` should be in evil-types.el, but I couldn't find a better place.

As always, please let me know if the PR can be improved.

Refs #760

### DIFF
--- a/evil-types.el
+++ b/evil-types.el
@@ -382,8 +382,7 @@ Returns a list (REGISTER COUNT)."
          (arg-count (length split-args))
          (arg0 (car split-args))
          (arg1 (cadr split-args))
-         (number-regex "^[1-9][0-9]*$")
-         (register-regex "^[a-zA-Z]$")
+         (number-regex "^-?[1-9][0-9]*$")
          (register nil)
          (count nil))
     (cond
@@ -400,13 +399,21 @@ Returns a list (REGISTER COUNT)."
      ((> arg-count 2)
       (user-error "Invalid use")))
 
-    (when (and register (not (string-match-p register-regex register)))
-      (user-error "Invalid register"))
-    (when (and count (not (string-match-p number-regex count)))
-      (user-error "Invalid count"))
+    ;; if register is given, check it's valid
+    (when register
+      (unless (= (length register) 1)
+        (user-error "Invalid register"))
+      (setq register (string-to-char register)))
 
-    (list (if register (string-to-char register) nil)
-          (if count (string-to-number count) nil))))
+    ;; if count is given, check it's valid
+    (when count
+      (unless (string-match-p number-regex count)
+        (user-error "Invalid count"))
+      (setq count (string-to-number count))
+      (unless (> count 0)
+        (user-error "Invalid count")))
+
+    (list register count)))
 
 (provide 'evil-types)
 
--- a/evil-types.el
+++ b/evil-types.el
@@ -370,6 +370,44 @@ If visual state is inactive then those values are nil."
   (when (evil-ex-p)
     (evil-ex-get-substitute-info evil-ex-argument t)))
 
+(evil-define-interactive-code "<d/>"
+  "Ex delete argument."
+  (when (evil-ex-p)
+    (evil-ex-get-delete-info evil-ex-argument)))
+
+(defun evil-ex-get-delete-info (string)
+  "Parse STRING as a :delete argument.
+Returns a list (REGISTER COUNT)."
+  (let* ((split-args (split-string (or string "")))
+         (arg-count (length split-args))
+         (arg0 (car split-args))
+         (arg1 (cadr split-args))
+         (number-regex "^[1-9][0-9]*$")
+         (register-regex "^[a-zA-Z]$")
+         (register nil)
+         (count nil))
+    (cond
+     ;; :delete REGISTER or :delete COUNT
+     ((= arg-count 1)
+      (if (string-match-p number-regex arg0)
+          (setq count arg0)
+        (setq register arg0)))
+     ;; :delete REGISTER COUNT
+     ((eq arg-count 2)
+      (setq register arg0
+            count arg1))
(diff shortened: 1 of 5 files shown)


evil-types.el
```diff
@@ -370,6 +370,45 @@ If visual state is inactive then those values are nil."
...
 
+(evil-define-interactive-code "<d/>"
+  "Ex delete argument."
+  (when (evil-ex-p)
+    (evil-ex-get-delete-info evil-ex-argument)))
+
+(defun evil-ex-get-delete-info (string)
+  "Parse STRING as a :delete argument.
+Return a list (REGISTER COUNT)"
+  (let* (
```
wasamasa: Lone parentheses make me sad :<

evil-types.el
```diff
@@ -370,6 +370,45 @@ If visual state is inactive then those values are nil."
...
+         (split-args (split-string (or string "")))
+         (arg-count (length split-args))
+         (arg0 (nth 0 split-args))
```
wasamasa: The general convention is to use `car` for the first element, `cadr` for the second and `nth` for anything with a greater index than 1.

evil-types.el
```diff
@@ -370,6 +370,45 @@ If visual state is inactive then those values are nil."
...
+         (arg1 (nth 1 split-args))
+         (number-regex "^[1-9][0-9]*$")
+         (register-regex "^[a-zA-Z]$")
+         (register nil)
+         (count nil))
+    (cond
+     ;; :delete REGISTER or :delete COUNT
+     ((eq arg-count 1)
```
wasamasa: Use `=` for number comparisons.  The fact that `eq` works on integers (and characters, as they happen to be integers) is an implementation detail.

evil-types.el
```diff
@@ -370,6 +370,45 @@ If visual state is inactive then those values are nil."
...
     (evil-ex-get-substitute-info evil-ex-argument t)))
 
+(evil-define-interactive-code "<d/>"
+  "Ex delete argument."
+  (when (evil-ex-p)
+    (evil-ex-get-delete-info evil-ex-argument)))
+
+(defun evil-ex-get-delete-info (string)
+  "Parse STRING as a :delete argument.
+Return a list (REGISTER COUNT)"
```
wasamasa: Missing a period at the end of the sentence.  Also, shouldn't it be "Returns"?

edkolev:
Thanks for the feedback. I've applied the suggested changes.

Getting the first and second arg is now implemented like so:
```
(let ....
         (arg0 (car split-args))
         (arg1 (car (cdr split-args)))
)
```

Is this what you meant or should the first arg be extracted with `car`, and the second with `nth 1`?

Let me know if something can be further improved.
Maybe `(defun evil-ex-get-delete-info (string) ...)` should be renamed to `(defun evil-ex-parse-delete-argument (string) ...)`

wasamasa:
No, the second should be extracted with `cadr`.  Everything after the second should be extracted with `nth`.  The name of the helper function is fine as is.  I'll test the functionality next to see whether anything is weird here.

edkolev:
Thanks, I've updated to code to use `cadr`.

wasamasa:
Looks like Vim handles a bit more errors, `:d 0` fails with "Positive count required" (same for 0.1 and other weird kinds of numbers), `:d -1` however deletes the current line (and clearly violates that previous error message), so no need to emulate that.  Your solution gives me an "Invalid register" error, so it's probably a matter of modifying `number-regex` and the error handling in `evil-ex-get-delete-info`.

Register handling appears to cover most cases, but fails on a command like `:d _ 1` which Vim happily accepts.  I'd take a closer look at what the allowed registers are in other code dealing with them.

edkolev:
I found this list of valid registers in vim's `:help registers`

```
There are ten types of registers:                       registers E354
1. The unnamed register ""
2. 10 numbered registers "0 to "9
3. The small delete register "-
4. 26 named registers "a to "z or "A to "Z
5. three read-only registers ":, "., "%
6. alternate buffer register "#
7. the expression register "=
8. The selection and drop registers "*, "+ and "~
9. The black hole register "_
10. Last search pattern register "/
```

I'll search a bit more in evil's code for code checking if a register is valid and modifiable by the user, but at a first glance it seems to me that post places do a `read-char` to get a register char, and the char eventually ends up in `evil-set-register` which does special checks for certain registers (like `_`) but at the end has a catch-all
```
(cond
.... ;; handle special registers like _
(t
    (set-register register text))
```

Maybe I should add a new function `(evil-user-modifiable-registerp register)` which will return true if the register is valid and the user can modify it?

wasamasa:
No, I'd just use `evil-set-register` then and essentially ignore all invalid use.  No need to make it complicated or have ad-hoc checking when there is one source where it's done for us.  We might even improve the function later and have it error out, conditionally if needed.

edkolev:
I just pushed a commit which:
- removes `register-regex`
- prints user error message "Invalid count" when the count is negative, e.g. `:d -1`

wasamasa:
Looks good!  Providing a float for the count gives you an "Invalid register" error, but that's roughly the same as Vim's "Trailing characters".  If you're feeling like it, you can try enhancing `evil-set-register` next so that it errors out on the three read-only registers.

edkolev:
Sure! I just pushed a commit -- `evil-set-register` now checks the given register is a valid character (with `characterp`) and isn't read-only.

As a result, `:delete fake` gives error `Invalid register`

wasamasa:
Thanks!  I'll just go ahead and merge this one then.

PR #822, opened 2017-05-01
Add issue template, move bug report section

wasamasa:
This should hopefully improve the quality of bug reports.  I've taken the liberty of moving the existing bug report section to a `CONTRIBUTING.md` file.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,29 @@
+<!-- Replace every comment and delete this line -->
+# Issue type
+
+<!-- Pick one of the following, delete the rest -->
+- Bug report
+- Enhancement request
+- Question
+- Other
+
+# Environment
+
+Emacs version: <!-- M-x emacs-version -->
+Operating System: <!-- Name and version, e.g. Windows 10, macOS Sierra, Debian Jessie -->
+Evil version: <!-- M-x evil-version -->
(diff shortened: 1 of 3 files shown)


edkolev:
FYI `make emacs` works for me only after removing the `&` at the end of the line (in the Makefile). This probably has to do with the fact that I always run ` make emacs` inside of tmux, on OSX.

wasamasa:
Interesting, I suspect this is one of the countless features exclusive to GNU make.  Can you verify this by installing coreutils from brew or alike and re-running with `gmake emacs`?

edkolev:
It's the same with `gmake`, emacs prints the same message `emacs: standard input is not a tty` (output of make if pasted below).

On my OSX system, when I run `emacs` in terminal (with or without tmux) it always starts emacs in the terminal, it doesn't start it as a desktop app.

Does running `emacs` start a desktop app on other platforms?

```
~/dev/evil $ gmake -v
GNU Make 4.2.1
...
~/dev/evil $ gmake emacs
emacs -Q -L . -L lib -l goto-chg.el -l evil-tests.el \
--eval "(evil-mode 1)" \
--eval "(evil-tests-initialize '() '() t)" &
~/dev/evil $ emacs: standard input is not a tty
^C
```

wasamasa:
Yes it does, so this is very weird.  Anyways, I'm open to removing the backgrounding attempt, just not in this PR.

Issue #824, opened 2017-05-01
Can't replace newlines with :s


oskarkv:
OS: Linux Mint
Emacs version: 24.5.1
Evil version: f0684c1f6644a8c4e0e5fce50a1a04796555eaf1
Running GUI in X.

# Steps to reproduce.

1. `make emacs`
2. (All tests pass.)
3. In a buffer with
```
line1
line2
line3
```
4. Try to replace the newlines with the character a, by pressing `:%s/ C-q C-j /a/ RET`.
5. Observe the preview before pressing `RET`.

# Observed behavior

The preview is wrong. It looks like only the newlines after `line1` and `line3` (but not `line2`) are going to be replaced. When pressing `RET`, nothing changes, i.e. the buffer remains unchanged. Emacs says "Replaced 0 occurrences".

justbur:
I think this one is a problem with the bounds on the line by line search. I'm submitting a PR that fixes it.

oskarkv:
Apparently it works with the `g` flag. But I think it should work without. It does in Vim. And the preview is still wrong without `g`.

justbur:
@oskarkv Try #825

Yes, when the g flag is used the replacements are performed according to a different function. It's the line by line changes that had a bug in them.

oskarkv:
It works, but the preview is still wrong. It only shows every second replacement. Typing that extra g immediately shows all replacements in the preview.

justbur:
Different part of the code. I think I got it now. Try the PR again when you get a chance.

oskarkv:
Yeah, it works. Good job!

justbur:
Here's another problem that I just found and haven't fixed

Start (note extra empty line on last line)
```
[a]bc
def
ghi

```
Command `:%s/\n/Z/`
```
abcZdefZghiZZ
```
Command `:%s/\n/Z/g`
```
abcZdefZghiZ
```

The cause is [here](https://github.com/emacs-evil/evil/blob/master/evil-commands.el#L3347-L3350)

cc @wasamasa

wasamasa:
Looks like the edge case wasn't handled well enough to me.

PR #825, opened 2017-05-02
Fix bounds on line by line search in evil-ex-substitute

justbur:
The per line bound should be the beginning of the next line in order to capture
new line chars (Note this won't allow the search to spill over into the new
line). All tests pass with this change.

See #824

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -3404,7 +3404,10 @@ resp.  after executing the command."
                                (goto-char m)
                                (move-marker m nil))
                              (when (re-search-forward evil-ex-substitute-regex
-                                                      (line-end-position) t nil)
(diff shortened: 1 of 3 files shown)


wasamasa:
Thanks, looks good!  Could you add a regression test for the originally reported condition that triggered the fixed bug?

justbur:
@wasamasa I found another inconsistency in the substitute code (see the inconsistency in the tests). I'll open a new issue.

wasamasa:
Is this MR good to go, now that the other one has been merged?

justbur:
The substitute code needs some more work to deal with the inconsistency I talked about above, but that can be done in another pr

justbur:
Closed in favor of #832

PR #827, opened 2017-05-03
Fix evil-search-next when search string is newline

### LINKED ISSUE #823: evil-search-next gets stuck when searching for newlines
OS: Linux Mint
Emacs version: 24.5.1
Running GUI in X.

# Steps to reproduce.

1. `make emacs`
2. (All tests pass.)
3. In a buffer with
```
line1
line2
line3
```
4. Search for a newline with `/ C-q C-j RET`. Press `n` several times. Press `N` several times.

# Observed behavior

The cursor gets stuck on the first match when using `n`, i.e. pressing `n` again does not move on to the next match. But `N` works as expected.

Setting `evil-search-module` to `evil-search` via the customization interface fixes the problem.

justbur:
When the search string is a newline and evil-move-beyond-eol is nil, point gets
moved back an extra character after moving to the beginning of the matched
character (the newline). evil-search-next then finds that same newline again and
we get stuck.

Add regression test.

Fixes #823

### DIFF
--- a/evil-search.el
+++ b/evil-search.el
@@ -229,6 +229,13 @@ one more than the current position."
       (set-text-properties 0 (length string) nil string)
       ;; position to search from
       (goto-char start)
+      ;; when `evil-move-beyond-eol' is nil and we're looking at a \n after a
+      ;; search, evil will move the cursor back so that we start the new search
+      ;; at a position prior to the \n that we just found.
+      (when (and (not evil-move-beyond-eol)
+                 (not (eobp))
+                 (= (char-after) ?\n))
+        (forward-char))
       (setq isearch-string string)
       (isearch-update-ring string regexp-p)
       (condition-case nil
--- a/evil-tests.el
+++ b/evil-tests.el
@@ -7225,7 +7225,18 @@ if no previous selection")
         (error search-failed "n")
         "foo foo foo\nbar [b]ar\nbaz baz baz\n"
         (error search-failed "N")
-        "foo foo foo\nbar [b]ar\nbaz baz baz\n"))))
+        "foo foo foo\nbar [b]ar\nbaz baz baz\n"))
+    (ert-info ("Test search for newline")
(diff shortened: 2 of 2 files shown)


wasamasa:
Hm, not sure whether this is the best way of fixing it.  Isn't `evil-move-beyond-eol` only relevant for interactive use anyways?

justbur:
> Hm, not sure whether this is the best way of fixing it. 

It's ugly, but I didn't find another way that was clearly better. If you find something nicer, I'm open to that. 

> Isn't evil-move-beyond-eol only relevant for interactive use anyways?

I'm not sure exactly what you mean. The adjustment is done in a post-command-hook, so if I do `n` to search for the next newline, point is first moved to the beginning of the match, but because the beginning of the match is past the "end of the line" it is pulled back one more time when evil-move-beyond-eol is nil (note that the bug doesn't occur when evil-move-beyond-eol is t). If you look in the `evil-search` code there is already a one unit adjustment forward to avoid matching the same string again in every case. This takes care of the case in which point was moved back two places following a match of a newline.

wasamasa:
So, if I understand correctly, the following happens (with | denoting the cursor *between* characters):

```
foo bar |baz
qux quux
```

Search for `\n`:

```
foo bar baz|
qux quux
```

Post-command adjustment:

```
foo bar ba|z
qux quux
```

Another search attempt, but with `evil-move-beyond-eol` enabled:

```
foo bar baz|
qux quux
```

Post-command adjustment

```
foo bar baz|
qux quux
```

Hitting `n`:
```
foo bar baz
qux quux|
```

justbur:
Yes that looks right.

It might be better to make evil-search-next handle this case but I'm not sure.

wasamasa:
OK, I'm just trying to come up with a legitimate search coming out funny due to the workaround to see whether it's a good enough solution.

justbur:
Closing in favor of #828

PR #828, opened 2017-05-04
Fix evil-search-next when search string is newline

### LINKED ISSUE #823: evil-search-next gets stuck when searching for newlines

justbur:
When the search string is a newline and evil-move-beyond-eol is nil, point gets
moved back an extra character after moving to the beginning of the matched
character (the newline). evil-search-next then finds that same newline again and
we get stuck.

Add regression test.

Fixes #823

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -2565,14 +2565,18 @@ for `isearch-forward',\nwhich lists available keys:\n\n%s"
         (search-string (if evil-regexp-search
                            (car-safe regexp-search-ring)
                          (car-safe search-ring))))
-    (evil-search search-string isearch-forward evil-regexp-search)
-    (when (and (> (point) orig)
-               (save-excursion
-                 (evil-adjust-cursor)
-                 (= (point) orig)))
-      ;; Point won't move after first attempt and `evil-adjust-cursor' takes
-      ;; effect, so try again.
-      (evil-search search-string isearch-forward evil-regexp-search))
+    (goto-char
+     ;; Wrap in `save-excursion' so that multiple searches have no visual effect.
+     (save-excursion
+       (evil-search search-string isearch-forward evil-regexp-search)
+       (when (and (> (point) orig)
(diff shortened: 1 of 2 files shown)


justbur:
@wasamasa This is a narrower and I think better solution.

wasamasa:
I agree, the only thing that took me a while to get is that the command doesn't simply retry, but retry one time less than specified.  The only problem I can spot with this is a bit of flickering as hitting `n` triggers *two* searches.

justbur:
Found one problem that I just fixed. 

I saw some flickering too but I think that's from the `evil-adjust-cursor` logic, meaning it wasn't introduced here.

wasamasa:
Huh, shouldn't flickering by `evil-adjust-cursor` be prevented by the `save-excursion`?

justbur:
Not inside the new evil-search-next. I meant the post-command call to evil-adjust-cursor, which actually moves it back.

wasamasa:
Hm, adding several `sit-for` calls reveals that if I search for a newline at the end of a line, point moves to the end, one back, to the end, to the next line's end and only then is adjusted by the post-command loop.  So I reckon it's not only the post-command call, but the search calls, too.

justbur:
@wasamasa OK, I pushed another commit to eliminate the visual effect due to an extra search.

wasamasa:
Hm, still not quite gone, but better than before.

PR #829, opened 2017-05-09
List view buffers kbd Q kill buffer and window

duianto:
```
problem:  evil-list-view buffers like these: evil-show-registers,
          evil-show-marks, evil-show-jumps and evil-ex-show-digraphs, leaves
          behind their window, when they are closed by pressing: q
solution: add Q as an additional key binding that calls: kill-buffer-and-window
```

The issue occurs when only one window is open and one calls `:reg`. Because the evil registers buffer opens in a new split window, and when `q` is pressed, then only the buffer gets removed, but the new split window stays open.

`q` works as expected when two or more windows were open before calling `:reg`. Because the previous window configuration gets restored when the evil register buffer is closed.

This PR makes it easier to choose if the window also should be closed.

### DIFF
--- a/evil-common.el
+++ b/evil-common.el
@@ -3838,6 +3838,7 @@ should be left-aligned for left justification."
       (funcall evil-list-view-select-action (nth 1 entry)))))
 
 (define-key evil-list-view-mode-map (kbd "q") #'kill-this-buffer)
+(define-key evil-list-view-mode-map (kbd "Q") #'kill-buffer-and-window)
 (define-key evil-list-view-mode-map [follow-link] nil) ;; allows mouse-1 to be activated
 (define-key evil-list-view-mode-map [mouse-1] #'evil-list-view-goto-entry)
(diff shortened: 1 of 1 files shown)


edkolev:
I personally would prefer to have `q` or `ESC` close this window, without introducing `Q`. I don't think `Q` is used for closing in either vim or evil.

duianto:
There's one case where `Q` is part of a key binding for closing/quitting.

In Vim the description for `ZQ`, states that:
>Quit without checking for changes (same as ":q!")

In Evil `ZQ` (evil-normal-state-map) is bound to the `evil-quit` command. It's doc-string says:
>Closes the current window, current frame, Emacs.
If the current frame belongs to some client the client connection
is closed.

wasamasa:
I don't see any reason why this non-Vim behavior (remember, `Q` is for entering ex mode) couldn't be a line in the user config.  There is no precedent for doing this, neither in my Emacs installation nor the ELPA packages I've installed.  If anything, this is a candidate for Spacemacs or alike, not Evil.

PR #831, opened 2017-05-10
Don't run in emacs in background with "make emacs"

edkolev:
As discussed in #822, with this, `make emacs` will not run emacs in the background


wasamasa:
Thanks!

Issue #811, opened 2017-04-14
%s/foo//gn doesn't work


daphnisz:
In vim, you can type `:%s/foo//gn` to get the number of occurrences of `foo` without changing anything, but this doesn't work in evil.

edkolev:
I tried adding support for `n` in `evil-ex-substitute`, but it turned out a big task:

It isn't a simple task when both `g` and `n` are given, because with `g` all the replacing is done with `perform-replace`. Hence, `evil-ex-substitute` will have to be refactored to not use `perform-replace` and that is a change I'm not comfortable doing. The diff which add support for `n` (but *doesn't* work with both `gn`) is pasted below in case some wants to pick that up.

I would like to suggest this alternative approach: instead of supporting `n`, we can show the number of matches like so:

![screen shot 2017-04-19 at 09 41 32](https://cloud.githubusercontent.com/assets/1532071/25166528/7f8b477a-24e4-11e7-8417-938bae572c50.png)

The code is very simple, it simply prints the count of overlays (no extra counting is done).
We may add a configuration variable which, if nil, will not bother the user with the number of matches.
I can wrap this in a PR if the maintainers think this this is a good change.

The half-working diff adding support for `n`:

```diff
git diff evil-commands.el
diff --git a/evil-commands.el b/evil-commands.el
index 44ec3aa..281f904 100644
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -3329,7 +3329,8 @@ resp.  after executing the command."
   (setq replacement (or replacement ""))
   (setq evil-ex-last-was-search nil)
   (let* ((flags (append flags nil))
-         (confirm (memq ?c flags))
+         (count-only (memq ?n flags))
+         (confirm (and (not count-only) (memq ?c flags))) ;; ignore "c" flag if "n" if given
          (case-fold-search (evil-ex-pattern-ignore-case pattern))
          (case-replace case-fold-search)
          (evil-ex-substitute-regex (evil-ex-pattern-regex pattern)))
@@ -3420,14 +3421,16 @@ resp.  after executing the command."
                                      (line-end-position) t nil)
               (setq evil-ex-substitute-nreplaced
                     (1+ evil-ex-substitute-nreplaced))
-              (evil-replace-match evil-ex-substitute-replacement
-                                  (not case-replace))
+              (unless count-only
+                (evil-replace-match evil-ex-substitute-replacement
+                                    (not case-replace)))
               (setq evil-ex-substitute-last-point (point)))))

[...]

PR #832, opened 2017-05-11
Rewrite evil-ex-substitute

### LINKED ISSUE #824: Can't replace newlines with :s

justbur:
New version of #825 which now fixes #824 and #811. 

I rewrote evil-ex-substitute to remove the call to `perform-replace` and to streamline the logic. 

I added the n flag as suggested by @edkolev in #811 and also took the liberty of adding support for the vim confirmation prompt which looks like
```
Replace a with b (y/n/a/q/l/^E/^Y)?
```
and includes the C-e and C-y keybindings like vim.


wasamasa:
Wow, thanks! Behavior seems to be like previously, so far the only oddity I've run into is that I've managed to run into an odd state where typing `:s/\n/g/` displays a `[stringp]` message next to it in the echo area.  Trying to debug where exactly in `evil-ex-update-highlights` it came from undid the display of that message.

I've also noticed that in the same file the `(evil-ex-define-argument-type substitution ...)` call typo'd the `:expanded` property. Hopefully that's not the real cause of anything fixed in this PR?

Regarding code, the only thing sticking out so far is that markers aren't made to point nowhere after the command is done.  The elisp reference specially mentions that they can only be garbage-collected if they don't reference a buffer.  This can be done with `(move-marker my-marker nil)`.

justbur:
> so far the only oddity I've run into is that I've managed to run into an odd state where typing :s/\n/g/ displays a [stringp] message next to it in the echo area. Trying to debug where exactly in evil-ex-update-highlights it came from undid the display of that message.

I'm not seeing that

> I've also noticed that in the same file the (evil-ex-define-argument-type substitution ...) call typo'd the :expanded property. Hopefully that's not the real cause of anything fixed in this PR?

I fixed that, but I'm not sure how that would affect things here.

> Regarding code, the only thing sticking out so far is that markers aren't made to point nowhere after the command is done. The elisp reference specially mentions that they can only be garbage-collected if they don't reference a buffer. This can be done with (move-marker my-marker nil).

Fixed, thanks

wasamasa:
No issue, I'll put some debug prints into the code to catch it next time. Thanks again!

wasamasa:
@justbur I've caught it this time, the culprit is the `(string= pattern "^")` in your new code.  `string=` is intended to be used on strings only and will error out on anything else (not 100% true, but close enough), `pattern` is a list of three elements and does therefore not work with this check.  I'd suggest you to improve the check, otherwise it's just dead code for an untested corner case.

justbur:
@wasamasa you're absolutely right. See #836

PR #835, opened 2017-05-20
Fix minor issue with evil-paste-from-register

justbur:
With point on a word, :%s/<C-r><C-w> gives an end of buffer warning from the
forward-char at the end of evil-paste-from-register.


wasamasa:
I'm sorry, but how is substitution related to `evil-paste-from-register`?

justbur:
That's how I found it. The issue occurs when you paste at the end of the buffer which is often the case in the minibuffer.

wasamasa:
I see, thanks!

PR #836, opened 2017-05-21
Fixup f005120 Fix edge case with evil-search highlighting

justbur:
pattern is an object. We need to get the regex first.

See #832


wasamasa:
Thanks!  You think this can be tested meaningfully or it's not worth it adding a regression test?

justbur:
It would be nice to test it I guess, but I'm not sure how to test interactive highlighting like this. Since it's a visual thing, I see it as less important.

wasamasa:
Alright, I've found a different visual glitch in the meantime, but will open a separate issue for it.

wasamasa:
Hm, can't reproduce it again.

Issue #846, opened 2017-06-05
Suggestion: create a reusable evil-test-helpers package to make testing evil extensions simpler


edkolev:
# Issue type

- Enhancement request

I think it would make sense to split `evil-tests.el` into `evil-tests-helpers.el` and `evil-tests.el`and  create a new Melpa package from `evil-tests-helpers.el`. It will contain functions like `evil-test-buffer`.

As a result, evil extensions will be able to reuse the helper functions. Now, I copy-paste parts of `evil-tests.el` to make writing tests easier and consistent with evil itself. Example: https://github.com/edkolev/evil-lion/blob/master/test/evil-lion-test.el#L257

I can prepare a PR (here and in melpa's repo) in case you think this is an improvement.

wasamasa:
Sounds good to me.  I'd be in for this change, as long as `evil-tests-helpers.el` remains in this repository so that Evil itself doesn't need to do anything else than `(require 'evil-tests-helpers)` in its test file.

expez:
I wanted this earlier, I ended up copying what I needed into [evil-smartparens](https://github.com/expez/evil-smartparens/blob/master/tests/evil-tests/evil-tests.el#L12) too.

ninrod:
@edkolev excelent ideia.

edkolev:
https://github.com/emacs-evil/evil/pull/850 is now merged, closing.

edkolev:
As a consequence of #850, the `evil` melpa package now includes evil-test-helpers.el. This is due to the default melpa regex for package files which is:
```
  ("*.el" "*.el.in" "dir"
   "*.info" "*.texi" "*.texinfo"
   "doc/dir" "doc/*.info" "doc/*.texi" "doc/*.texinfo"
   (:exclude ".dir-locals.el" "test.el" "tests.el" "*-test.el" "*-tests.el"))
```

Infesting the official `evil` package with evil-test-helpers.el doesn't seem right to me.

Maybe the file should be renamed to `evil-util-tests.el` so it's not included in the package (it will be matched by `*-tests.el`)? Alternatively, the melpa recipe can be modified.

wasamasa:
I'd modify the recipe to explicitly excluse the helper file.  Less surprises and all that.

edkolev:
OK, I suggest this Melpa recipe:
```emacs-lisp
(evil :repo "emacs-evil/evil"
      :fetcher github
      :files ("evil*.el"
              "doc/*.texi"
[...]

PR #850, opened 2017-06-12
Split evil-tests.el into two files

edkolev:
The new file evil-test-helpers.el can be used by evil plugin authors
to write tests in the same manner as evil-mode's tests.

Refs #846


wasamasa:
This breaks `make`.  Is this intentional?  Full error:

```
make: *** No rule to make target 'evil-test-helpers.el,evil-types.el', needed by 'evil-test-helpers.el,evil-types.elc'.  Stop.
```

edkolev:
It's not intentional - all make commands (`make test`, `make emacs`, etc) must not be affected. 

However, I can't reproduce the error. Maybe the `.depend` file needs to be recreated.

Could you please try to remove this file with `rm .depend`. `make clean` removes this file as well but I guess you won't be able to run it due to the error.

TheBB:
As far as I'm aware the correct variant is

    FILES = $(filter-out evil-tests.el evil-test-helpers.el evil-pkg.el,$(wildcard evil*.el))

There doesn't need to be multiple `filter-out`s like before, and you need to use spaces instead of commas.

Hooray for stringly types.

edkolev:
Thanks, I just amended the commit with the suggested change in the Makefile.

wasamasa:
Thanks, this makes it build again.  The other thing sticking out here is that there is no `(require 'evil)` or `(require 'elt)` in the test file.  Technically, it's not needed, but it still seems wrong to me that it isn't there.

edkolev:
OK, makes sense, the commit is amended.

I was thinking that it might be a good idea to make it obvious that `evil-test-buffer` is the only thing exported by `evil-test-helpers.el`. As a result, this file can be easily changed in the future while only trying to preserve this single macro's behaviour.

Should I rename the functions/macros in `evil-test-helpers.el` to make it obvious they are private? Maybe with a prefix `evil--...`.

wasamasa:
Nah, exposed API is exposed API.  That aside, there's some candidates that are most likely used, like `evil-tests-initialize` (Makefile), `evil-with-temp-file` (evil-tests.el) and others.

Could it be that you cut off at the wrong point?  Everything from `evil-test-file-contents` shouldn't be in evil-test-helpers.el.

edkolev:
I'll try to finish this PR within the following week, sorry for the delay, I've been busy.

edkolev:
I finally had some time for this PR. I probably cut off at the wrong place last time. This time however, I started from scratch, moved `evil-test-buffer` to its own file, then moved its underlying functions to the new file.

I successfully managed to run some basic tests on an evil plugin using the `evil-test-buffer` macro from the newly created file.

wasamasa:
Great, thanks!

PR #853, opened 2017-06-15
Add evil-kill-on-visual-paste option

justbur:
Allows one to prevent evil-visual-paste from automatically adding the replaced
text to the kill ring, which changes what will be pasted next.

It may just be me, but I hate the default vim behavior of essentially yanking the replaced text. I don't see why you would want to replace some text with a paste and then paste the replaced text somewhere else.

This is a simple change to make in the evil source that would be difficult to accomplish with an external plugin.

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -1980,7 +1980,8 @@ The return value is the yanked text."
         (if paste-eob
             (evil-paste-after count register)
           (evil-paste-before count register)))
-      (kill-new new-kill)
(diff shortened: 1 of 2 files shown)


wasamasa:
Thanks, merged!

ninrod:
I use https://github.com/Dewdrops/evil-ReplaceWithRegister for that as visual pasting is not repeatable with `.` operator. 

but it is great to have that option anyway.

PR #856, opened 2017-06-19
Keep Emacs from clearing text properties on yank

### LINKED ISSUE #855: Yanking a line removes wrong text's properties
# Issue type

- Bug reports

# Environment
Emacs version: 25.2.1
Operating System: Arch
Evil version: 1.2.12
Evil installation type: MELPA
Graphical/Terminal: X

# Reproduction steps
in scratch buffer:
_(cursor)line1_

type yyp in normal mode -> 

_(cursor-old)line1
(cursor-new)line1(expect-position)_

# Expected behavior
Text properties of text between (cursor-new) and (expect-position) will be removed

# Actual behavior
Text properties of text between (cursor-old) and (cursor-new) is removed

# Further notes
remove-yank-excluded-properties may not be called properly.

When we yank a line, we use evil-paste-after which involves a calling to insert-for-yank-1. The insert-for-yank-1 removes the yanked text's text properties ([remove-yank-excluded-properties](https://github.com/emacs-mirror/emacs/blob/695d50fa0a11c6bd98f9f4a9b5a2905d33282316/lisp/subr.el#L3035)). However, in this context, the parameters for remove-yank-excluded-properties is wrong.

wasamasa:
Closes #855

### DIFF
--- a/evil-common.el
+++ b/evil-common.el
@@ -2430,7 +2430,9 @@ The tracked insertion is set to `evil-last-insertion'."
   "Saves the lines in the region BEG and END into the kill-ring."
   (let* ((text (filter-buffer-substring beg end))
          (yank-handler (list (or yank-handler
-                                 #'evil-yank-line-handler))))
+                                 #'evil-yank-line-handler)
+                             nil
+                             t)))
     ;; Ensure the text ends with a newline. This is required
     ;; if the deleted lines were the last lines in the buffer.
     (when (or (zerop (length text))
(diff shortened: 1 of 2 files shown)


wasamasa:
I'm not sure whether this catches everything, from quick testing it does.  Curiously enough the other paste operation doesn't seem to be affected.  Given this change, it makes little sense for `evil-yank-line-handler` to remove text properties itself as `insert-for-yank` already does, but maybe I'm overlooking something.

TheBB:
Looks good, thanks!

Issue #726, opened 2016-12-01
vt' vf' vT' vF' vt" vf" vT" vF" do not work for deadkeys in us-international keyboard layout


TheBB:
Originally reported by: **Filipe Silva (Bitbucket: [ninrod](https://bitbucket.org/ninrod), GitHub: [ninrod](https://github.com/ninrod))**

----------------------------------------

test case (cursor [|])

```
#!txt


[t]his line ends with a quote' -> command: vt'
this line ends with a quot[e]' -> result
```

```
#!txt

[t]his line ends with a quote" -> command: vt"
this line ends with a quot[e]" -> result
```
The issue here is visual-mode operator pending with any character that has to be formed with deadkeys in an international keyboard. eg: the command `vté` won't work either.

the same happens for all other combinations. if instead of quotes we place any other character, then it works as expected.

the expected behaviour is: evil should visually select the text from point up to the char before the quote.

Interesting enough, this works fine in terminal emacs. It only breaks on gui emacs.

emacs-25.1 gui
evil-1.2.12
osx sierra

----------------------------------------
- Bitbucket: https://bitbucket.org/lyro/evil/issue/726

ninrod:
I'm the original reporter. Just adding that this works on terminal emacs, but not on gui emacs.

ninrod:
I've also found that this works in GNU/Linux gui emacs, but not on MacOS Gui Emacs. Seems very specific: You will encounter the bug if:

- use gui emacs on OSx with a deakey keaboard layout like US-International, where you can write things like `äíóúãõàüö`.

ninrod:
Maybe the root cause for these lies in #605.

Issue #605, opened 2016-01-02
evil-replace should take into account current input method


TheBB:
Originally reported by: **Arthur Fayzrakhmanov (Bitbucket: [Geraldus](https://bitbucket.org/Geraldus), GitHub: [Geraldus](https://github.com/Geraldus))**

----------------------------------------

Hi friends!

Emacs 24.5.1, Evil from MELPA.

This is what we have:  

* go to `NORMAL` mode  
* change input method, e.g. set `russian-computer`  
* hit `R` key to replace char under cursor (runs `evil-replace`)  
* type any char  

**Result**: char under cursor will be replaced with latin char ignoring current input method.  
**Expected**: char should be replaced respecting to current input method.

----------------------------------------
- Bitbucket: https://bitbucket.org/lyro/evil/issue/605

khaoos-abominable:
I am a new evil (spacemacs) user. I use russian-computer input method for writing comments and notes. And this issue causes a little pain in the neck while I am editing them. 

I am going to fix it in my envirionment with minimal effort and invasion. Could anyone suggest a function that translates characters according to the current input method (or passed as a parameter)? For example, which translates `"asdf"` to `"фыва"` for `russian-*` input methods. Is this possible? Thanks in advance.

ninrod:
I'm pretty sure that this is closely related to #726, maybe almost a duplicate.

khaoos-abominable:
By the way, could anyone explain me a part of documentation of a function `read-char` which says:
> Read a character from the command input (keyboard or **macro**).

What does it mean to read a character from a macro? How to make the function read from a macro?

duianto:
@khaoos-abominable This page seems to show an example of the `read-char` function reading from a macro: https://www.gnu.org/software/emacs/manual/html_node/elisp/Reading-One-Event.html

In the description for the second example in the function:
> Function: read-char &optional prompt inherit-input-method seconds

> The second example shows a keyboard macro definition that calls `read-char` from the minibuffer using `eval-expression`. `read-char` reads the keyboard macro’s very next character, which is 1. Then `eval-expression` displays its return value in the echo area.

``` elisp
;; We assume here you use M-: to evaluate this.
(symbol-function 'foo)
     ⇒ "^[:(read-char)^M1"
(execute-kbd-macro 'foo)
     -| 49
     ⇒ nil
```

khaoos-abominable:
[...]

PR #857, opened 2017-06-20
use read-key instead of read-char

willghatch:
In a terminal, read char will read only part of a multibyte unicode character, whereas read-key will get the whole thing.  Graphical emacs doesn't have this problem.  This is related to #726 and potentially #605.  So for instance when an evil operator prompts me for a character and I type `»` from my non-US/qwerty layout, it gets gibberish instead (`Â` is sent, then `\273` is sent to the buffer when I press another key).  This fixes it.

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -732,7 +732,7 @@ Columns are counted from zero."
   :keep-visual t
   :repeat nil
   :type exclusive
-  (interactive (list (read-char)))
+  (interactive (list (read-key)))
   (let ((marker (evil-get-marker char)))
     (cond
      ((markerp marker)
@@ -756,7 +756,7 @@ Columns are counted from zero."
   :keep-visual t
   :repeat nil
   :type line
-  (interactive (list (read-char)))
+  (interactive (list (read-key)))
   (evil-goto-mark char noerror)
   (evil-first-non-blank))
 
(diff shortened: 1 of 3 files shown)


ninrod:
Gentleman, Let's be very careful now. This read-key thing was the main culprit of a [very painful](https://github.com/emacs-evil/evil-surround/commit/9724dcce1b1a390320da839d5c0a7c02e20bdf04) regression on [evil-surround](github.com/emacs-evil/evil-surround) which I had to revert. It broke the `.` operation on evil-surround, which is the main feature of evil-surround. 

I'd be extra careful in merging this in.

wasamasa:
Hm, my objection to this would be based on a somewhat related PR for preserving the prompt of a preceding key which broke a keymap: https://github.com/emacs-evil/evil/pull/798

As far I can see this PR does it the other way around and the difference is basically that commands allowing for Emacs input methods use `read-key` whereas the rest uses `read-char`.

willghatch:
It seems this patch is fraught with peril that I was unaware of.  I'm not aware of anything it breaks (it passed the test suite and my light usage since making it is fine), but a similar patch does break `.` in evil-surround.  So... yeah.  Maybe don't merge it.  Anyone have any ideas on how to fix the issue?  `read-char` is clearly wrong, and I don't know how `read-key` is mucking up other things.

willghatch:
@ninrod thanks for the warning.

justbur:
I could be wrong, but try using `read-char` with the `INHERIT-INPUT-METHOD` flag, like this `(read-char "Char: " t)`.

willghatch:
@justbur thanks!  That seems to work for both evil and evil-surround.

PR #860, opened 2017-06-22
Fix regression in evil-ex-substitute with nxml-mode

### LINKED ISSUE #859: Search and replace in nXML mode fails
# Bug

- Bug report

# Environment

Emacs version: GNU Emacs 25.2.1 (x86_64-unknown-linux-gnu) of 2017-06-21
Emacs version: GNU Emacs 25.2.1 (x86_64-apple-darwin15.6.0) of 2017-04-24
Operating System:  Ubuntu 16.04.2 LTS 
Operating System: OS X El Capitan
Evil version: Evil version 1.2.12
Evil installation type: MELPA
Graphical/Terminal: Terminal
Terminal multiplexer: tmux (no difference w/o)

# Reproduction steps
- Start Emacs
- Create XML buffer:

`
<a>
  <init></init>
  <init></init>
  <init></init>
  <init></init>
  <init></init>
  <init></init>
</a>
`

- Search and replace:
:%s/init/foo/gc
y
y

# Expected behavior
`
<a>
  <foo></foo>
  <init></init>
  <init></init>
  <init></init>
  <init></init>
  <init></init>
</a>
`

# Actual behavior
`
<a>
  <foo></init>
  <init></init>
  <init></init>
  <init></init>
  <init></init>
  <init></init>
<foo
`

# Further notes

justbur:
nxml-mode somehow changes the match data during the course of using
evil-ex-substitute with the gc flags. Make evil-ex-substitute very careful about
using the right match-data for moving point and performing replacements. This
should not be necessary, but I guess we can't rely on modes to be careful about
not overwriting the match data randomly.

Fixes #859

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -3391,58 +3391,59 @@ resp.  after executing the command."
           (goto-char beg)
           (catch 'exit-search
             (while (re-search-forward evil-ex-substitute-regex end-marker t)
-              (goto-char (match-beginning 0))
-              (setq match-contains-newline
-                    (string-match-p "\n" (buffer-substring-no-properties
-                                          (match-beginning 0)
-                                          (match-end 0))))
-              (if confirm
-                  (let ((prompt
-                         (format "Replace %s with %s (y/n/a/q/l/^E/^Y)? "
-                                 (match-string 0)
-                                 (evil-match-substitute-replacement
-                                  evil-ex-substitute-replacement
-                                  (not case-replace))))
-                        response)
-                    (move-overlay evil-ex-substitute-overlay
-                                  (match-beginning 0)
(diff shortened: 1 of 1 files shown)


justbur:
Wait on merging this. A test is failing, but I don't have time now to look into why

justbur:
I think this is good now. Please check/test

TheBB:
What was the issue? (with the tests I mean) I looked at it myself and was mystified. Not easy to see from the diff what you changed.

justbur:
Oh, I force pushed. I needed to create markers for match-beg and match-end,
so the positions were correct after replacing text. I didn't do that
correctly initially.

TheBB:
Thanks!

PR #861, opened 2017-06-24
Use evil-esc in minibuffers

jojojames:
This avoids having to input ESC ESC ESC in Terminals when trying to
exit out of Minibuffer/Ivy/Helm/etc.

Can then write something like this for Ivy to escape out.

```
(eval-after-load 'evil
    (lambda ()
      (define-key ivy-minibuffer-map [escape] 'minibuffer-keyboard-quit)))
```


wasamasa:
Ah, this is a nice addition.  Thanks!

PR #865, opened 2017-06-26
advise `read-key` as well as `read-key-sequence`

willghatch:
As discussed in #857 and #858, the use of `read-key` in the evil-surround package broke `evil-repeat`, or `.`.  This is because `read-key` changes the result of `this-command-keys` just like `read-key-sequence`.  `read-key-sequence` is advised to fix this issue.  This patch merely gives the same advice to `read-key`, fixing the issues that have been caused by its use.

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -3767,7 +3767,7 @@ resp.  after executing the command."
     ;; tag, which bypasses the source of `this-command-keys', we'd be able
     ;; to capture the key(s) in the post-command of the operator as usual.
     ;; Fortunately however, `last-input-event' can see the key (by default, `return')
-    (unless (append (this-command-keys) nil)
+    (when (= (length (this-command-keys)) 0)
       (evil-repeat-record (vector last-input-event))))
    (t (evil-repeat-motion flag))))
 
--- a/evil-common.el
+++ b/evil-common.el
@@ -580,13 +580,9 @@ Translates it according to the input method."
                 (setq char (aref cmd 0)
                       cmd (key-binding cmd)))
               (cond
-               ((eq cmd 'self-insert-command)
-                char)
-               (cmd
-                (setq evil-last-read-digraph-char nil)
-                (call-interactively cmd))
-               (t
-                (user-error "No replacement character typed"))))
+               ((eq cmd 'self-insert-command) char)
+               (cmd (call-interactively cmd))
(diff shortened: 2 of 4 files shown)


ninrod:
hum. 1 test case failed. can't find it.

wasamasa:
In graphical Emacs, this fails two repeat-related tests, `evil-test-repeat-digraph` and `evil-test-repeat-kill-buffer`.

willghatch:
Ah, when I ran the tests I think I didn't see the failure because it was picking up .elc files.  I've now fixed the tests and I get no errors in terminal or graphical mode.  But having fixed the test, I'm now not sure that the advice for `read-key-sequence` (and with this patch also `read-key`) is actually correct.  As is, it calls any function that is the result of `(evil-repeat-type this-command t)`, and when that is `evil-repeat-keys` that makes a lot of sense.  But when it is `evil-repeat-changes` (which is what `evil-insert-digraph` uses), it makes a lot less sense to me.  I think maybe the advice should only call it when it is `evil-repeat-keys`.

willghatch:
Huh, I get no errors on my machine, but travis is getting one and I can't tell what it is that failed.  If I understand the travis page right, it's passing for emacs 24.3 but failing for all the others it's trying.  I'm on emacs 26 built from emacs git master of a couple months ago...  Maybe tomorrow I'll get a version of emacs it's failing on and test it there.

wasamasa:
This is unfortunate indeed.  IIRC @TheBB is working on a version of the tests where as much as possible runs in batch mode.

TheBB:
Yeah I am, they're so damn fickle though.

TheBB:
If anyone is still interested in this, I rebased it on master (with a more sensible travis setup) and there's only one failure for 24.5.

willghatch:
On Fri, Nov 22, 2019 at 06:28:02AM -0800, Eivind Fonn wrote:
>If anyone is still interested in this, I rebased it on master (with a more sensible travis setup) and there's only one failure for 24.5.

Cool.  I still care (to get evil-surround to work properly with
multi-byte utf-8 characters, which I frequently want), but
unfortunately emacs improvements are pretty low on my priority list
right now.  So I don't really have time to work on this in the near
future.

wbolster:
also still interested since evil-swap-keys.el and evil's repeat facility seem to interfere and it seems like this is related

Issue #866, opened 2017-06-27
Evil motion for `avy-goto-char-timer` missing


rieje:
# Issue type

- Enhancement request

# Environment

Emacs version:  25.2.1
Operating System: Arch Linux
Evil version: 1.2.12
Evil installation type: MELPA
Graphical/Terminal: X

# Further notes

Evil motion for `avy-goto-char-timer` is missing. Evil motion exists for all other functions (evil-avy-*); unfortunately, I only use this particular function for avy.

wasamasa:
Does `(evil-define-avy-motion avy-goto-char-timer)` do the right thing?  If yes, it should be a one line change.

rieje:
Sorry, I'm an elisp noob. If you meant evaluating it as a sexp then debugger says wrong number of arguments.

TheBB:
It should be

```elisp
(evil-define-avy-motion avy-goto-char-timer inclusive)
```

rieje:
Actually, it seems it's not needed? I use evil-surround package and I bound `avy-goto-char-timer` to `s`, I think it might have conflicted with evil-surround somehow.

Does `avy-goto-char-timer` work with evil motions without that change? Maybe `evil-avy-goto-char-timer` is not necessary.

edkolev:
I use `avy-goto-char-timer` without anything special, I just bind it to the motion state map so it works for operations like `c gh` (where `gh` is bound to the avy function, see below).

This is how I bind it:
```
(use-package avy
  :ensure t
  :bind (:map evil-motion-state-map
              ("gh" . avy-goto-char-timer)))
```

edkolev:
To be clear - I believe no change in evil is necessary for `avy-goto-char-timer` to work nicely with evil.

wasamasa:
OK, then I'll close this issue.  Feel free to reopen if you do find something lacking with that resolution.

PR #867, opened 2017-07-06
Add evil-avy-goto-char-timer

Somelauw:
So it can be used when in pending operation state.


wasamasa:
Huh, I guess https://github.com/emacs-evil/evil/issues/866 was needed after all.  Thanks!

PR #868, opened 2017-07-09
Wire up semantic and tags for "goto-definition" when available

Hi-Angel:
Usually I am using a one year outdated evil-mode because [of this bug](https://github.com/emacs-evil/evil/issues/684), and upon occasional recent update I found the vanilla evil is missing this nice little feature of using semantic and tags.

So I thought others might be happy to get this as well. I've used this code for may be a year.

I forgot to mention in the commit: I am trying imenu first because I vaguelly remember that on a rare occasion, when both are available, it's better than semantic. At the very least because semantic have to parse includes/imports as well, when imenu might already have the info about location of the definition.

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -2652,6 +2652,14 @@ The search is unbounded, i.e., the pattern is not wrapped in
   (dotimes (var (or count 1))
     (evil-search-word t t symbol)))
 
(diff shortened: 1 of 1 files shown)


Hi-Angel:
Wait, I noticed little things that may be improved, I'm cancelling it, will resend.

PR #869, opened 2017-07-09
Wire up semantic and tags for "goto-definition" when available

Hi-Angel:
Usually I am using a one year outdated evil-mode because [of a regression](https://github.com/emacs-evil/evil/issues/684), and with occasional recent update I found the vanilla evil is missing this nice little feature of using semantic and tags.

So I thought others might be happy to get this as well. I've used this code for may be a year *(cleaned up a bit though before the commit).

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -2683,10 +2683,13 @@ The search is unbounded, i.e., the pattern is not wrapped in
          ;; highlight the occurrence
          ((numberp ipos)
           (evil-search search t t ipos))
-         ;; imenu failed, so just go to first occurrence in buffer
-         (t
-          (evil-search search t t (point-min)))))
-       ;; no imenu, so just go to first occurrence in buffer
+         ;; imenu failed, try semantic
+         ((and (fboundp 'semantic-ia-fast-jump)
+               (ignore-errors (semantic-ia-fast-jump ipos)))
+          ()) ;; noop, already jumped
+         ((fboundp 'xref-find-definitions) ;; semantic failed, try the generic func
+          (xref-find-definitions string))))
+       ;; otherwise just go to first occurrence in buffer
        (t
         (evil-search search t t (point-min)))))))
 


evil-commands.el
```diff
@@ -2652,6 +2652,16 @@ The search is unbounded, i.e., the pattern is not wrapped in
   (dotimes (var (or count 1))
     (evil-search-word t t symbol)))
 
+(defun evil-semantic-goto-definition (offset)
+  "Wrapper around semantic-ia-fast-jump returning nil on error or
+when the function not bound. On success it α) goes to the
```
wasamasa: ಠ_ಠ

evil-commands.el
```diff
@@ -2652,6 +2652,16 @@ The search is unbounded, i.e., the pattern is not wrapped in
...
+definition, and β) returns non-nil"
+  (if (fboundp 'semantic-ia-fast-jump)
+      (condition-case nil
```
wasamasa: You can just use `ignore-errors`.

Hi-Angel: Cool, thanks, it reduced ½ the code! Re-pushed.

Hi-Angel:
@wasamasa alright, do not pull please, let me check something.

Hi-Angel:
ok, nvm.

wasamasa:
Thanks!  I do wonder though at which point people will ask for a hook so that they can customize how things are found...

Dickby:
Until now :)
I use gtags for navigating and until now i used `evil-goto-definition` for stupidly jumping to the first occurence of a word in the buffer. After this change instead of jumping it askes for a tags table.

wasamasa:
[Obligatory XKCD reference](https://xkcd.com/1172/).

Point taken though, I'll try preparing a PR that allows you what I've proposed.

Hi-Angel:
@wasamasa may be you meant `xref-find-definition`? Because `semantic-ia-fast-jump` upon wrapped into `(ignore-errors α)` does just that.

Hi-Angel:
@wasamasa ah, I see what you mean — it's that `semantic-ia-fast-jump` takes a point instead of a string. Well, in this case it's complicated — you can't rewrite it to take a string, at all. Imagine a situation when there're two functions with same name local variables. If you pass it only a string — how could it know to which of 2 definitions jump?

Hi-Angel:
So to sum up — I think a position number is a better argument for the functions in your list, because it contains more information than a string only.

wasamasa:
Well, I disagree.  Anyway, if it's handled with a hook, anyone can dumb it down or make it smarter.  If there is no equivalent function for semantic, I'll simply not include a hook for that until someone rectifies this.

Hi-Angel:
Why?

Hi-Angel:
I mean, what's the reasoning — position number is technically better.

wasamasa:
I disagree because I'm not convinced by your arguments at all.  Besides, I do not wish to continue this thread any further than needed, especially not if most posts in it are going to be low-quality noise.

Hi-Angel:
So you disagree that we should design better solution? Great. Do whatever you wish, I can't stop you.

P.S. I want also to note if there will be any other sematic *(pun intended)* backends, they'll require position too.

Hi-Angel:
@wasamasa I mean, I could understand if there were contras to my solution. But there's none! It's the same as yours, just a different type of argument! It's a non-technical decision.

PR #873, opened 2017-07-12
Change evil-goto-definition to use a hook

wasamasa:
See also https://github.com/emacs-evil/evil/pull/869#issuecomment-314392574

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -2709,46 +2709,50 @@ The search is unbounded, i.e., the pattern is not wrapped in
   (dotimes (var (or count 1))
     (evil-search-word t t symbol)))
 
+(defun evil-goto-definition-imenu (string _position)
+  "Find definition for STRING with imenu."
+  (require 'imenu nil t)
+  (let (ientry ipos)
+    (when (fboundp 'imenu--make-index-alist)
+      (ignore-errors (setq ientry (imenu--make-index-alist)))
+      (setq ientry (assoc string ientry))
+      (setq ipos (cdr ientry))
+      (when (and (markerp ipos)
+                 (eq (marker-buffer ipos) (current-buffer)))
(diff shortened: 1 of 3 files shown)


TheBB:
We did something similar to this in Spacemacs, and it turned out to be much less useful than we thought because so many find-definition implementations are asynchronous, and thus can't indicate success or failure in the return value. I'm curious if anyone has any solution to this that doesn't involve a lot more complications.

wasamasa:
Can you give an example for asynchronous ones?  I find using `run-hook-with-args-until-success` a rather elegant solution and until Emacs gets futures/promises, I can't imagine anything elegant for asynchronous functions.

TheBB:
Currently in Spacemacs there seems to be `ycmd-goto`, `tern-find-definition`, `omnisharp-go-to-definition`, `eclim-java-find-declaration` and `anaconda-mode-find-definitions` from various packages. I haven't looked at them all in detail. What we did was to force-abort the on asynchronous ones on the assumption that language servers are typically powerful enough not to fail to find things.

joehillen:
I agree with @Hi-Angel in the previous thread that a position would be a better argument.

Sometimes a string can refer to different definitions based on context. It's not safe to assume that all instances of a string have the same definition. It might not even be correct that the occurrence at the top of the buffer is the definition.

Maybe a compromise would be to include both STRING and POSITION as arguments.

edkolev:
I don't see how the position will be used - both imenu and xref take strings as input. I'm not sure about semantic, I haven't used it.

wasamasa:
@joehillen This is a compromise I can live with.  Using point exclusively is inacceptable as it makes nearly every user of this hook look up the word at point, using it as additional argument makes it a minor annoyance that can be ignored in the argument list.

PR #876, opened 2017-07-19
Cleanup evil-test-helper.el

edkolev:
- add Package-Requires header
- fix byte compile warning about evil-test-selection
- move call to `provide` to the bottom of the file

Refs https://github.com/melpa/melpa/pull/4837
Refs https://github.com/emacs-evil/evil/issues/846


wasamasa:
Alright, thanks.

Closes #846

PR #878, opened 2017-07-22
Mention limitation of regex

chrissound:
Mention limitation of regex

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,3 +124,7 @@ modify the syntax-table in all buffers of some mode, e.g.:
 ```
 
 This gives the underscore the word syntax-class in all C-like buffers.
+
+# Limitations
+
+Lookahead and lookbehind regex is not supported.


wasamasa:
This is somewhat misleading to have on its own.  Do you plan on enumerating every single thing Emacs regex lacks?  Wouldn't it be better to point out that Evil uses Emacs regex first and foremost and that evil-search merely translates a few Vim shorthands to the equivalent Emacs regex?

chrissound:
It would be useful to know what functionality is not supported though, rather than having to find out on one's own. 

I'm not really sure what would be an ideal way to list this though. Perhaps a wiki page?

therockmandolinist:
> It would be useful to know what functionality is not supported though

i think the point was that this isn't unsupported functionality on evil's part, it's unsupported _emacs_ functionality (something people should know about emacs regex, not evil regex) so there's no point specifying it here.

On that point though, 
> Wouldn't it be better to point out that Evil uses Emacs regex first and foremost and that evil-search merely translates a few Vim shorthands to the equivalent Emacs regex?

What shorthands are you speaking of here, @wasamasa ?

wasamasa:
@chrissound This kind of approach has the same weakness as blacklists and enumeration of badness.  What I thought of is something along the lines of:

```
- Evil uses Emacs' regular expressions which are close to POSIX RE, but come with Emacs-specific extensions such as syntax table characters and evaluation of elisp code
- This means that equivalent regular expressions are more verbose and extensions such as positive/negative lookahead are not supported
- The `evil-search` module translates certain backslash codes as known from Vim to the corresponding Emacs counterparts, see `evil-transform-vim-style-regexp`

@therockmandolinist See the last note above.

therockmandolinist:
woah, now I know about `evil-ex-search-vim-style-regexp`. I had been wanting this for a while, thanks!

Issue #404, opened 2014-07-04
Dired menus are duplicated


TheBB:
Originally reported by: **Rafael Guerra Paz (Bitbucket: [RafaelGP](https://bitbucket.org/RafaelGP), GitHub: [RafaelGP](https://github.com/RafaelGP))**

----------------------------------------

Here is an screenshot:

![dired](http://clients.tourvista.co.uk/vt/zip/tmp/dired.png)

This is what I do to replicateit:

1. Start Emacs like: C:\emacs\bin\runemacs.exe --quick
2. Write the following code in the scratch buffer and evaluate it:
   (add-to-list 'load-path "~/.emacs.d/elpa/evil-20140625.411")(require 'evil)(evil-mode 1)
3. Do M-x dired and choose a directory
4. The menus: Operate, Mark, Regexp, Immediate and Subdir get duplicated

Emacs version: GNU Emacs 24.4.50.1 (x86_64-w64-mingw32) of 2014-06-09 on KAEL

Evil version: Evil version 1.0-dev (installed though ELPA - Marmalade)

----------------------------------------
- Bitbucket: https://bitbucket.org/lyro/evil/issue/404

PR #879, opened 2017-07-23
Redefine dired keys without overriding map

jojojames:
evil-make-overriding-map will include a second set of menu items
resulting in duplicate menus.

https://github.com/emacs-evil/evil/issues/404

Might not be the best approach but at least fixes the duplicate menus. The keys should be exactly the same as the original.

I'd like to change the keys up a little too to be more evil-friendly. The default integration only adds hjkl but other familiar bindings are not there.

I have something like this to adapt to that instead.

```(add-hook 'dired-mode-hook
          (lambda ()
            (eval-after-load 'evil
              (lambda ()
                ;; Wipe the "g" so `evil' can use "gg".
                (define-key dired-mode-map (kbd "g") nil)
                ;; Wipe "0" so we can use
                ;; `evil-digit-argument-or-evil-beginning-of-line'.
                (define-key dired-mode-map (kbd "0") nil)
                ;; Rather have Visual Line than Mail function.
                (define-key dired-mode-map (kbd "V") nil)
                (evil-define-key 'normal dired-mode-map
                  " " nil
                  "$" 'evil-end-of-line
                  "G" 'evil-goto-line
                  "0" 'evil-digit-argument-or-evil-beginning-of-line
                  "w" 'evil-forward-word-begin
                  "Y" 'dired-copy-filename-as-kill
                  "v" 'evil-visual-char
                  "V" 'evil-visual-line
                  "m" 'dired-mark
                  "u" 'dired-unmark
                  "U" 'dired-unmark-all-marks
                  "c" 'dired-create-directory
                  "n" 'evil-ex-search-next
                  "N" 'evil-ex-search-previous
                  "q" 'quit-window)))))
```


wasamasa:
I'm sorry, but no. Duplicating vanilla keybindings is not OK (think about people who customized those), I wouldn't want to increase the integration level either.  There's both people who want more and less integration, that's why I don't want to increase their levels and let that to projects such as Spacemacs.

jojojames:
I'm not sure why we wouldn't want to provide a default level of integration with default Emacs packages. (Afterall, the file is 'evil-integration'). Sane defaults are always great and making people have to handpick every little thing isn't very DRY.

It might be better to provide a sane set of defaults and let the user configure what they want to change or allow them to opt in/out of the default evil integration instead of having every user repeat/discover the same set of incantations to work with dired/etc properly.

Re: this pull itself, it's OK not to take it but it might be better to have a story on avoiding the duplicate menus with EVIL. IMO, that's much worse than tweaking a couple of binds the user can override anyway.

At this point, it's not (as far as I know) easy to disable the dired 'integration' resulting in the duplicate menus.

PR #851, opened 2017-06-12
Prevent infinite loop in evil-ex-substitute

justbur:
Zero length regexps, particularly "$", may get matched over and over. Also add a
test for this.

Fixes #849

Issue #849, opened 2017-06-12
`:%s/$/ foo/g` hangs Emacs


wasamasa:
# Issue type

- Bug report

# Environment

Emacs version: GNU Emacs 25.2.1 (x86_64-unknown-linux-gnu, GTK+ Version 3.22.10) of 2017-04-22
Operating System: Arch Linux
Evil version: HEAD
Evil installation type: manual
Graphical/Terminal: graphical

# Reproduction steps
- Start Emacs
- Open buffer with text
- `:%s/$/ foo/g`

# Expected behavior
Every line should get " foo" appended

# Actual behavior
Emacs hangs, interrupting the hang with `C-g` reveals that the first line has been changed to `something foo foo foo foo foo foo foo ...`, indicating that something in the replacement code makes this operation not terminate.

# Further notes
Reported on `#evil-mode` by @rgrinberg, CC @justbur.

justbur:
well that's annoying. `(re-search-forward "$")` is special because the end of the match is also the beginning of the match so it matches the same eol over and over. Submitting a PR

rgrinberg:
Thank you!

PR #882, opened 2017-07-26
Fix evil-ex-substitute (#881)

### LINKED ISSUE #881: :substitute doesn't work correctly when replacing a single character X and the buffer contains XX
# Issue type

- Bug report

# Environment

Emacs version: GNU Emacs 25.2.1 (x86_64-apple-darwin13.4.0, NS appkit-1265.21 Version 10.9.5 (Build 13F1911)) of 2017-04-21
Operating System: OS X El Capitan
Evil version: Evil version 1.2.12 (git commit a0b6afcf3993f2edcce39052403220a10e74f362)
Evil installation type: Melpa
Graphical/Terminal: both
Terminal multiplexer: issue can be reproduced with and without tmux

# Reproduction steps
- `make emacs`
- add `""` in the file (2 double-quotes with no space between them)
- run `:%s/"//g`

# Expected behavior
- both the quotes are deleted

# Actual behavior
- one of the quotes is deleted -- the buffer contains `"`

# Further notes
```
fake1="" # the double-quotes on this line will also not be correctly removed
fake2=xx # running :%s/x//g will again reproduce the issue, the issue isn't affecting just quotes
fake3=" " # the quotes on this line will correctly be removed
```

FYI @justbur

justbur:
I'm beginning to not want to be responsible for this function. In dealing with the special case where `:%s/$/ TEST/` can loop infinitely because point never moves past the first "$", I think I introduced the regression in #881. With the line "xxxx" in the buffer `:s/x//` in the old version would skip over every other "x" because it accidentally identified the match as having zero length. 

It's not the prettiest piece of work, but I think I've dealt with the special cases here in a fairly robust way. I added three tests, and a bunch of comments to the code to explain. 

Please review

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -3379,6 +3379,7 @@ resp.  after executing the command."
          (evil-ex-substitute-hl (evil-ex-make-hl 'evil-ex-substitute))
          (orig-point-marker (move-marker (make-marker) (point)))
          (end-marker (move-marker (make-marker) end))
+         zero-length-match
          match-contains-newline
          transient-mark-mode)
     (setq evil-ex-substitute-pattern pattern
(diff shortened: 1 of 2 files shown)


wasamasa:
Heh, I can feel your pain.  I wonder though, which ticket do you believe you introduced this in?  #881 is the one reporting the issue, not a PR with your changes.

justbur:
Oh, sorry. It was in fixing #849 with #851

wasamasa:
Alright, thanks!  Let's hope there won't be many more.  While the code doesn't appear to have surprises lurking in it, it's too big for me to make any definitive judgement on that.

edkolev:
Thank you @justbur for fixing this and fixing it so quickly!

Issue #884, opened 2017-07-28
commit d457729 breaks evil-select-paren behaviour


ninrod:
# Issue type
- Bug report: the following macro does not work from d457729 (which is really a merge of emacs-evil/evil@a5465f3) up to `master`, but works in bc753f4.

# Environment

Emacs version: 25.2.1
Operating System: Centos7
Evil version: Evil version evil-git-41847fd
Evil installation type: manual (git clone)
Graphical/Terminal: graphical
Terminal multiplexer: n/a

# Reproduction steps
- Start Emacs

- verify that the following macro  does not work in d457729, which is a merge of commit:

`a5465f3: bisect fix evil up block -- make it symmetrical forward and backward` by @willghatch 

but it works in the previous commit: bc753f4:

here's the final result of the git bisect:

![image](https://user-images.githubusercontent.com/8352747/28735877-b2441bce-73bd-11e7-9fbf-1029d439ac78.png)

```elisp
;; modified to be able to specify name and use `function'
(defmacro define-and-bind-text-object (name key start-regex end-regex)
  (let ((inner-name (make-symbol (concat "evil-inner-" name)))
        (outer-name (make-symbol (concat "evil-a-" name))))
    `(progn
       (evil-define-text-object ,inner-name (count &optional beg end type)
         (evil-select-paren ,start-regex ,end-regex beg end type count nil))
       (evil-define-text-object ,outer-name (count &optional beg end type)
         (evil-select-paren ,start-regex ,end-regex beg end type count t))
       (define-key evil-inner-text-objects-map ,key #',inner-name)
       (define-key evil-outer-text-objects-map ,key #',outer-name))))

(define-and-bind-text-object "pipe" "|" "|" "|")
```
- now try to `ci|` inside pipes: `| blahblahbl[a]lblaksdfadfkj |`

# Expected behavior
`ci|` works as expected: deletes everything inside the pipes and leaves user in insert mode.

# Actual behavior
`ci|` does not work. 

# Further notes

- this also breaks the promising __[targets.el](https://github.com/noctuid/targets.el/issues/19)__ package, brought to you by @noctuid.

- also see [this](https://github.com/noctuid/evil-textobj-anyblock/issues/9) discussion, where the macro was brought up.

ninrod:
__note__: emacs-evil/evil@a5465f3 can be reverted cleanly from evil git head. This solves the problem.

ninrod:
with the advent of @willghatch's #889, this can be closed.

PR #889, opened 2017-08-01
fix pipe text-object macro

willghatch:
Fixes issue #884.

The code assumed (badly) that the start and end patterns were distinct.  If the start and end patterns are the same it was looping until hitting the end of the buffer.  This fixes the balancing behavior to not try to balance delimiters that are the same on both sides.

### DIFF
--- a/evil-common.el
+++ b/evil-common.el
@@ -1379,7 +1379,8 @@ to reach zero). The behaviour of this functions is similar to
                           (point))))
                (match (match-data t))
                (op (save-excursion
-                     (and (re-search-forward (if forwardp beg end) cl t dir)
+                     (and (not (equal beg end))
+                          (re-search-forward (if forwardp beg end) cl t dir)
                           (or (/= pnt (point))
                               (progn
                                 ;; zero size match, repeat search from
--- a/evil-tests.el
+++ b/evil-tests.el
@@ -8144,6 +8144,42 @@ maybe we need one line more with some text\n")
      "foo unde[f] bar") ;; 'undef' is not an abbrev, shouldn't be expanded
     (setq abbrevs-changed nil)))
 
+(ert-deftest evil-test-text-object-macro ()
+  :tags '(evil abbrev)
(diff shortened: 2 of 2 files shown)


wasamasa:
Wow, this is more than I expected after my suggestion in the linked threads.  Thanks for figuring out the real issue here and providing test cases!

ninrod:
@willghatch, could you verify if this is intended? 

suppose you have the following text: (cursor is represented like this `[]`)

```txt
/blahblahblah[/]
```
if you hit `ci/` wouldn't the text be changed to `//`. ? 

If think this pull might have an off by one error because if point is anywhere else in the previous WORD then it works.

willghatch:
If I have the text:

`/blahblahblah[/]`

and I hit `ci/` I get `/[]/` like you say, but isn't that what I should expect?  When I am at that same original place and hit `ca/` I get `[]`, not `[]/`, so I'm not sure what's going on there for you.  For me things are working as I expect, though my expectations could be wrong slightly.  For instance, if I have:

`/blahblah[/]blahblah/`

and I hit `ci/` I get:

`/blahblah/[]/`, and I could imagine that maybe the proper behavior is to change the previous pair.  But it's an ambiguous case and either choice is just as arbitrary.  I'm not sure what the precident is, though.  The only paren-like thing in normal vim that isn't different on the left and right is `"`, and it determines whether to go left or right by trying to balance them from the start of the file or through the filetype or something, so sometimes it goes left and sometimes right.  But that's not necessarily the thing to do with things like `|` in shell scripts or `/` in file paths and the like.

ninrod:
hum... I'm not getting the same result as you. 
`/blahblah[/]` -> `ci/` gives me `/blahblah[/]`. 

I have `targets.el` installed though. 

is `/` a standard text object that comes with vim or did you customize evil to add it. If so, how did you do it?

willghatch:
I used the same macro that you used to add the `|` object, just adding another use of it with `/`.

ninrod:
yes the macro works perfectly with this pull, thanks. It must be a bug in targets.el.

PR #900, opened 2017-08-10
Add evil-respect-visual-line-mode option

justbur:
This option adds some initial support for adjusting the behavior of some
important commands in visual-line-mode to make them behave more intuitively.

This commit just adjusts the behavior of evil-insert-line and evil-append-line.


justbur:
I have some different thoughts on how to go about doing this

PR #901, opened 2017-08-11
[WIP] Add evil-respect-visual-line-mode option

justbur:
This option adjusts the behavior of commands in visual-line-mode to make them
behave using visual lines.

It works by redefining what the beginning/end of line is when the option is
enabled and visual-line-mode is enabled. Behavior should not change when the
option is disabled.

This affects the behavior of many commands, including

I: evil-insert-line
A: evil-append-line
j: evil-next-line
k: evil-previous-line
c: evil-change
C: evil-change-line
d: evil-delete
D: evil-delete-line
y: evil-yank
Y: evil-delete-yank-line
0: evil-beginning-of-line
$: evil-end-of-line


justbur:
See #902

PR #902, opened 2017-08-12
Add evil-respect-visual-line-mode option

justbur:
Supersedes #901 

Remap commands to use visual version when `visual-line-mode` is active.
These commands are swapped

```
evil-next-line         <-> evil-next-visual-line
evil-previous-line     <-> evil-previous-visual-line
evil-beginning-of-line <-> evil-beginning-of-visual-line
evil-end-of-line       <-> evil-end-of-visual-line
```

**Note**: I previously thought reworking yy, dd, pp, and so on was worth it here, but I changed my mind. I don't think the behavior is that intuitive (at least how I did it), and I think using specific motions like d$ are a better choice.

### DIFF
--- a/evil-common.el
+++ b/evil-common.el
@@ -723,8 +723,10 @@ has already been started; otherwise TARGET is called."
            (setq this-command #'digit-argument)
            (call-interactively #'digit-argument))
           (t
-           (setq this-command #',target)
-           (call-interactively #',target)))))))
+           (let ((target (or (command-remapping #',target)
+                             #',target)))
+             (setq this-command target)
+             (call-interactively target))))))))
 
 (defun evil-extract-append (file-or-append)
   "Return an (APPEND . FILENAME) pair based on FILE-OR-APPEND.
--- a/evil-integration.el
+++ b/evil-integration.el
@@ -553,6 +553,16 @@ Based on `evil-enclose-ace-jump-for-motion'."
      (add-to-list 'evil-motion-state-modes 'ag-mode)
      (evil-add-hjkl-bindings ag-mode-map 'motion)))
 
+;; visual-line-mode integration
+(when evil-respect-visual-line-mode
+  (let ((swaps '((evil-next-line . evil-next-visual-line)
+                 (evil-previous-line . evil-previous-visual-line)
+                 (evil-beginning-of-line . evil-beginning-of-visual-line)
+                 (evil-end-of-line . evil-end-of-visual-line))))
+    (dolist (swap swaps)
+      (define-key visual-line-mode-map (vector 'remap (car swap)) (cdr swap))
+      (define-key visual-line-mode-map (vector 'remap (cdr swap)) (car swap)))))
+
 (provide 'evil-integration)
 
 ;;; abbrev.el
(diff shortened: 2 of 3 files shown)


VanLaser:
I have to ask, do you see this option as a "convenience" setting? Because both Vim and Evil already have the case covered, using the `g`-based movement prefix: `gj`, `gk` ... `g0`, `g^`, `g$`, `gm` etc. - and Evil users can at any time switch they movement keybindings to these, when visual line mode is on. In another words, this patch does not add a new "Vim emulation" feature, but something that is not in Vim. I'm not against it, per se, just asking about the motivation of your patch.

justbur:
Yes, this is not in the category of "vim emulation". I write papers in LaTeX and visual-line-mode is very useful there. I eventually want to be able to work naturally with just visual lines. Unfortunately, the way evil is designed does not make this easy. In vim you could just remap things. Something like this

```
nmap j gj
nmap k gk
nmap 0 g0
nmap $ g$
nmap D dg$
nmap dd g0dg$
...
```

The first commit here fixes a first problem with doing this in emacs. You can't just remap `evil-beginning-of-line` and have that be picked up by the `0` binding because that binding has a wrapper around it that doesn't know about command remapping. 

I added the second commit as something people might find convenient. 

Going further than this and changing `D`, `C`, etc is more work, and I only want to do it if it will be accepted. I might try to do this all as an external package, but it's going to be hacky due to how the code for functions like `evil-delete` and `evil-delete-line` are written. 

Anyway, the first commit in this PR is very reasonable in my opinion. Whether or not the second one is accepted is a matter of taste. I think it falls into the "evil-integration" category, but I can see how others would disagree.

vyp:
My 2 cents is that it will be fine as long as it lives under this option. 👍

VanLaser:
@justbur  I also see no problem with having this, but I admit I don't see it too difficult to remap the keys in evil to their visual equivalents. E.g. even in the most difficult case, by looking in `evil-maps.el`, it's easy to change the role of `0`, even if wrapped:

```emacs lisp
(evil-redirect-digit-argument evil-motion-state-map "0" 'evil-beginning-of-visual-line)
```

And the others are simple remaps (after doing a series of `C-h k j`, `C-h k g j`) e.g.:

```emacs lisp
(define-key evil-motion-state-map "j" #'evil-next-visual-line)
(define-key evil-motion-state-map "k" #'evil-previous-visual-line)
```
etc.

Of course, enabling your proposed option *is* easier, so I wouldn't be against it :-)

ninrod:
I'm all for merging this option in, if it is not enabled by default. Awesome work @justbur.

wasamasa:
Very well.  Considering that it's surprisingly common for Vim people to recommend and use remappings of the `nmap j gj` kind, I'm all in for an option that allows to achieve the effect in a complete manner.

@justbur Do you plan on adding tests or other additional code to this MR?

justbur:
@wasamasa No tests right now, because they are just bindings for the most part. 

I would like to add some more support for operating on visual lines in the future, but I think I'm OK with this for now. I can get away with things like

```elisp
(define-key evil-normal-state-map "D" "d$")
```

which may be the "right way" to do this.

wasamasa:
Alright then.

Issue #775, opened 2017-02-18
Consider derived modes in evil's initial states


blaenk:
I remember finding some code (vaguely remember it was from @wasamasa) that advises the `evil-initial-state` function to assume a match when the current mode is _or derives from_ a mode in the list of initial state modes. For example, if I add `prog-mode` to `evil-normal-state-modes` then any mode that derives from `prog-mode` will be considered a match and therefore start in normal mode.

I'm wondering if this is still necessary, if perhaps this functionality has since been added to evil by way of some opt-in flag and I wasn't aware.

If not, I'm wondering if something like this can be added (again, in an opt-in manner if necessary).

Here's the code:

``` elisp
  (defun my-evil--real-function (fun)
    "Figure out the actual symbol behind a function.
Returns a different symbol if FUN is an alias, otherwise FUN."
    (let ((symbol-function (symbol-function fun)))
      (if (symbolp symbol-function)
          symbol-function
        fun)))

  (defun my-evil--derived-mode-p (mode modes)
    (let ((parent (my-evil--real-function mode)))
      (while (and parent (not (memq parent modes)))
        (setq parent (my-evil--real-function (get parent 'derived-mode-parent))))
      parent))

  (with-eval-after-load 'evil-core
    (defun evil-initial-state (mode &optional default)
      "Return the Evil state to use for MODE.
Returns DEFAULT if no initial state is associated with MODE.
The initial state for a mode can be set with
`evil-set-initial-state'."
      (let (state modes)
        (catch 'done
          (dolist (entry (nreverse (evil-state-property t :modes)) default)
            (setq state (car entry)
                  modes (symbol-value (cdr entry)))
            (when (or (memq mode modes)
                      (my-evil--derived-mode-p mode modes))
              (throw 'done state)))))))
```

wasamasa:
Yes, that code is mine and no, nothing of the sort has been integrated.  Personally, I'd prefer seeing a better approach to this, like by using hooks to set initial states (which works for minor modes, but not for major ones as Evil is activated afterwards...).  The above approach is significantly hackier as seen in https://github.com/Malabarba/paradox/issues/86.

wasamasa:
Closing in favor of #497.

VanLaser:
I have to say I still like this solution the best, and was looking for it as I also remembered seeing this code at Bitbucket. Wouldn't it be possible just to break the (improbable) "endless loop" case by limiting the  to some reasonable number? (assuming I understood the problem)

[...]

Issue #497, opened 2015-03-20
Regarding initial states


TheBB:
Originally reported by: **Vasilij Schneidermann (Bitbucket: [wasamasa](https://bitbucket.org/wasamasa), GitHub: [wasamasa](https://github.com/wasamasa))**

----------------------------------------

I've found out that Evil is basically using [a whitelist of major modes](https://bitbucket.org/lyro/evil/src/2785c9a2cbfef5bcc4e9c6bed6fdc76ce2c2ffbb/evil-vars.el?at=default#cl-527) to decide the initial state for buffers.  This approach is OK, but I wished to write less for configuration, so I wrote a hack that makes it not only compare against modes, but also check for derived modes and aliases in [my Emacs configuration](https://github.com/wasamasa/dotemacs/blob/master/init.org#evil).  That way one can for example mark `text-mode`, `prog-mode` and `fundamental-mode` as normal state modes and have nearly everything else covered because most modes in sufficiently recent Emacs versions derive from them.

However, it occurred to me that this could be done in an even simpler manner, simply by using the respective commands in the hooks these modes have.  Adding `evil-normal-state` to the hooks of the previously mentioned modes seems to have a very similiar effect and gets rid of the extra code I wrote.

Would either alternative be a possible candidate for inclusion into Evil?

----------------------------------------
- Bitbucket: https://bitbucket.org/lyro/evil/issue/497

VanLaser:
Before moving to any PR (as proposed by @wasamasa in [#775](https://github.com/emacs-evil/evil/issues/775#issuecomment-322307014)), I'd like to show a reworked solution and put it before your judgement (tests, advise etc.). So far it touches no function (apart from advising one) and creates a new way of setting up the initial state for a mode and all the modes derived from it, based on the hooks they run (parent hook runs first, derived afterwards, minor modes after that - if I got that right). I hope this is the "other thread" I was supposed to be posting in the first place.

```emacs lisp
;; this is a temporary advice for `evil-initial-state'
(defun evil--around-evil-initial-state (fun &rest args)
  "Make `evil-initial-state' also look for an overriding buffer-local variable
`evil-init-state' and return its value instead."
  (let ((res (apply fun args)))
    (if (and (boundp 'evil-init-state) evil-init-state)
	evil-init-state
      res)))

;; if all goes well, the code above will be the last part of `evil-initial-state' instead
(advice-add 'evil-initial-state :around #'evil--around-evil-initial-state)

[...]

PR #904, opened 2017-08-16
Get evil initial state from parent modes [updated]

VanLaser:
This is basically a modification of the solution proposed by @wasamasa (described in #775 and also discussed in #497) which allows `evil-initial-state` to decide the initial state for buffers based not only on the state set for the current mode (e.g. using `evil-set-initial-state`), but also on those of its parent modes. 

I added a check for unusual situations in the parent -> derived modes relationship (see [Malabarba/paradox#86](https://github.com/Malabarba/paradox/issues/86)), in which a derived mode is set also as a parent of its parent; in such situations, `evil-initial-state` will now raise an error (avoiding the endless loop from the original solution). Another proposed [approach](https://github.com/emacs-evil/evil/issues/497#issuecomment-322514176) was that of using a counter that bails out with an error if the visit from a parent to its parent (and so on) goes above a limit, but I think the current solution is the best because it checks for the actual problem.

I also added an extra optional parameter (`noinherit`) for `evil-initial-state`, allowing for an initial state check that does not consider the parents of the mode, just the mode itself (i.e. in this case, `evil-initial-state` works exactly as before this patch).

### DIFF
--- a/evil-core.el
+++ b/evil-core.el
@@ -313,7 +313,7 @@ initial state for a mode can be set with
 `evil-set-initial-state'."
   (let (state modes)
     (catch 'done
-      (dolist (entry (nreverse (evil-state-property t :modes)) default)
+      (dolist (entry (evil-state-property t :modes) default)
         (setq state (car entry)
               modes (symbol-value (cdr entry)))
         (when (or (memq mode modes)
--- a/evil-core.el
+++ b/evil-core.el
@@ -293,16 +293,18 @@ Returns a different symbol if FUN is an alias, otherwise FUN."
         symbol-function
       fun)))
 
-(defun evil--derived-mode-p (mode modes)
+(defun evil--get-parents (mode)
+  "Returns the parents of the MODE in a bottom-up list,
+including (and starting with) the mode itself."
   (let ((parent (evil--real-function mode))
-	visited-modes)
-    (while (and parent (not (memq parent modes)))
-      (push parent visited-modes)	; remember already visited parents
-      (setq parent			; get the next parent
+	modes)
+    (while parent
+      (push parent modes)
+      (setq parent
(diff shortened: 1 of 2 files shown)


evil-core.el
```diff
@@ -285,17 +285,39 @@ See also `evil-initial-state'."
...
-The initial state for a mode can be set with
+Returns DEFAULT if no initial state is associated with either
+MODE, or any of its parents. If NOINHERIT is not nil, returns
+DEFAULT if no initial state is associated with this mode. The
+initial state for a mode can be set with
 `evil-set-initial-state'."
   (let (state modes)
     (catch 'done
-      (dolist (entry (evil-state-property t :modes) default)
+      (dolist (entry (nreverse (evil-state-property t :modes)) default)
```
wasamasa: I'm not sure why I did add a `nreverse` here, IIRC it was so that my default state doesn't kick in yet.

VanLaser: It's true that I didn't follow that to the end, I just assumed you had a good reason for it. The test passes (on my PC) even without it. Should I take it out?

wasamasa: Yeah, if the need still arises, it would be better to make the order modes are checked customizable.

VanLaser:
Ok, please don't pull it yet; I thought about that order and I think I'll try to rework the logic, so that the code will first try to get the list of the parents (once) and then it will check upwards for the initial state of each parent, until I find one (or not). In any case, the initial state would come from the closest parent in the hierarchy (e.g. "father", not "grandfather"), rather than from the order given by the `(evil-state-property :modes)`. Since a mode can't be set to two initial states at the same time (I suppose!), that order shouldn't count in this case. I'll probably make a new clear pull request (I'm not too used with this github workflow).

VanLaser:
Should be OK now.

PR #905, opened 2017-08-17
Slightly cleaner download and install step

Khalian:


wasamasa:
Interesting, I never noticed that the README suggests `~/.emacs.d/evil`.  Thanks!

PR #913, opened 2017-08-31
Eliminate a few byte compiler warnings

YorkZ:


evil-ex.el
```diff
@@ -40,6 +40,7 @@
 ;; substitution command (":s/foo/bar"), the handler incrementally
 ;; highlights matches in the buffer as the substitution is typed.
 
+(require 'comint)
```
wasamasa: Please remove the `(require 'comint nil t)` further below.

evil-commands.el
```diff
@@ -550,7 +550,7 @@ and jump to the corresponding one."
        ((< open close) (goto-char open-pair))
        (t (goto-char close-pair)))))))
 
-(eval-when-compile
+(eval-and-compile
```
wasamasa: I don't quite see how this makes a difference to just putting `(require 'flyspell)` there.  If you go for that route, you might as well remove the below `(require 'flyspell)`.

evil-ex.el
wasamasa: That being said, you could give the `(require 'shell nil t)` the same treatment and add `(require 'shell)` at the top after removing it.

evil-commands.el
YorkZ: This eliminate the ``Warning: the function ‘flyspell-overlay-p’ might not be defined at runtime.`` byte compilation warning message while running ``make``. But yes, you are right, it's better to simply use ``(require 'flyspell)`` at the top level and remove the the below ``(require 'flyspell)``. I will do this.

evil-ex.el
YorkZ: Adding ``(require comint)`` was to eliminate the

    Warning: the function ‘comint-completion-at-point’ is not known to be defined.

byte compiler warning message. 

The reason for requiring ``comint`` was because function ``comint-completion-at-point`` is defined in this file. ``(require shell)`` works too because ``shell.el`` included ``comint.el``. However, I personally prefer requiring the smallest possible dependency. If you don't agree with me, feel free to change to ``(require shell)``.

YorkZ:
Hi Vasilij,

I've commented your review, and have revised my commits as per your suggestion. Probably because I amended my commits and force pushed, you may need to click the ``Show outdated`` links to read my comments. Please let me know if you prefer not to amend committing followed by force push.

Thanks,

York

evil-ex.el
wasamasa: That's fine, my top concern is making sure there are no duplicate imports which would confuse a future reader.  With this change of yours, both `(require 'shell nil t)` and `(require 'comint nil t)` have become unnecessary and should be removed as well.

wasamasa:
Sure, that's fine.  I've left another comment.

evil-ex.el
YorkZ: No, all my changes address different byte compilation warnings. `(require comint)` targets the warning:

    Warning: the function ‘comint-completion-at-point’ is not known to be defined

You may need to run `make clean` before running `make` to see the previous warning messages.

wasamasa: I'm aware of that, just pointing out that `comint` is required twice (and `shell` once). You don't need that second import.

YorkZ: Yes, you are right, we don't need to require `comint` elsewhere in this file. My apology for missing your point. I've deleted the two function level `require`.

YorkZ: I've also changed to requiring `shell` instead of `comint` as per your suggestion. Thanks.

wasamasa:
Thanks!

PR #914, opened 2017-09-01
Fix term-mode undo issue

toctan:
Currently Evil breaks `term-char-mode` undo by enabling `undo-tree-mode`, see
https://github.com/syl20bnr/spacemacs/issues/4063#issuecomment-162913414

`turn-on-undo-tree-mode` checks the variable `undo-tree-incompatible-major-modes`
before enabling `undo-tree-mode`, see https://github.com/emacs-evil/evil/blob/master/lib/undo-tree.el#L2618

### DIFF
--- a/evil-integration.el
+++ b/evil-integration.el
@@ -221,7 +221,7 @@ some buffer, but only if `global-undo-tree-mode' is also
 activated."
        (when (and (boundp 'global-undo-tree-mode)
                   global-undo-tree-mode)
-         (undo-tree-mode 1)))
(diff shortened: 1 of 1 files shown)


wasamasa:
Hm, I don't exactly see a change in behavior with and without this PR, but I agree that undo-tree-mode shouldn't be enabled in ansi-term.  Thanks!

toctan:
@wasamasa Try use `C-_` instead of `C-/` in `ansi-term` to undo with & without this PR.

PR #915, opened 2017-09-02
Fix register `#'

YorkZ:
Because the `other-buffer` function returns the most recently selected buffer which it not necessarily the buffer being displayed in the "next window".
Therefore, we should use `(window-buffer (next-window))` instead.

### DIFF
--- a/evil-common.el
+++ b/evil-common.el
@@ -2041,7 +2041,7 @@ The following special registers are supported.
                                          evil-ex-current-buffer))
                   (user-error "No file name")))
              ((= register ?#)
-              (or (with-current-buffer (other-buffer) (buffer-file-name))
+              (or (buffer-file-name (window-buffer (next-window)))
                   (user-error "No file name")))
              ((eq register ?/)
               (or (car-safe


VanLaser:
From what I understand in Vim (reading both `:h alternate-file` and `:h quote#`) each window remembers the previous file associated with the buffer edited in that window, and keeps it under register `#`. Is that what your modification does?

YorkZ:
Thanks VanLaser for pointing this out. I haven't been using Vim since I started using Evil six years ago. Therefore, I didn't know that the register `#` in Vim stores the file name of the previous file visited in the current Window. If that's the case, and people wants the exactly same behaviour as Vim, my apology and I'm happy to withdraw this PR.

Personally, I don't see why one would find it more useful to get the file name of the previous file visited in the same Window, than to get the file name currently visited by the other Window next to the current one. However, that just me, doesn't matter. So please confirm, and I will withdraw this PR.

Thanks,

York

wasamasa:
The general guide line for PRs is to increase the Vim compatibility, other goals are fixing bugs and enhancing existing functionality to allow surpassing Vim.

YorkZ:
@wasamasa I understand the general guide line. Would you confirm that the `#` register in Vim is for getting the name of the file which was visited by the current Window immediately before switching to visiting the current file?

wasamasa:
I find Vim's help pretty instructive:

> 6. Alternate file register "#
> Contains the name of the alternate file for the current window.  It will
> change how the |CTRL-^| command works.
> This register is writable, mainly to allow for restoring it after a plugin has
> changed it.  It accepts buffer number: >
>     let altbuf = bufnr(@#)
>     ...
>     let @# = altbuf
> It will give error |E86| if you pass buffer number and this buffer does not
> exist.
> It can also accept a match with an existing buffer name: >
>     let @# = 'buffer_name'
> Error |E93| if there is more than one buffer matching the given name or |E94|
> if none of buffers matches the given name.

The error conditions are not about the visibility of the buffer, so I'd argue restricting the `#` register this way seems dubious.

VanLaser:
@YorkZ I hope I didn't appear too annoying or something with my reply - but from what I've read in the old vim emulation mailing list, Evil really tries to emulate Vim as close as possible :) 

To add to what @wasamasa said, I think the idea is that `#` is a simple history for the "last file edited here". Pulling a buffer from the next window into the current one is not a bad idea IMHO - perhaps it just doesn't need to be associated to the `#` register (this doesn't mean it can't be added somehow as _additional_ functionality).

YorkZ:
@VanLaser you definitely didn't appear as annoying. If Vim has another register to store the file name of `next-window`, it would be a great time to adapt this PR and add that ability to Evil. If not, I'm not sure whether most people in the Evil community would be interested in adding the feature in Evil. I personally have found this modified behavior very useful and have been using it for a few years. But again, that's just me.

YorkZ:
Closed to keep the behavior of `#` register the same as Vim.

PR #918, opened 2017-09-04
Add a dir-locals file

jgkamat:
This PR adds a trivial dir-locals file which overrides `indent-tabs-mode` to nil in emacs-lisp files.

Let me know if you think anything else should be added (fill-column, indention level, etc).


wasamasa:
I can't think of anything else and believe avoiding that bad default is a good start.  Thank you!

Issue #847, opened 2017-06-06
Abort macro recording (changed buffer)


greboide:
function evil-abort-macro is a bug and extremely evil even for evil ;), at the very least this should be configurable and off by default, there is no point in avoiding a macro changing buffer!!!

TheBB:
Some details would be appreciated.

edkolev:
You can use this doc for guidance on what details are needed https://github.com/emacs-evil/evil/blob/master/CONTRIBUTING.md

GitHub is supposed to ask you to fill it when opening an issue, please don't ignore it.
If it didn't ask you, maybe there's some bug with how it's set up (typo in the file name or something)

greboide:
sure, emacs by default allows us to change buffer when performing a macro, i believe that vim does not has this feature, maybe its a design issue with its macro, maybe its by design. The original author mimicked this vim behavior with this function prohibiting us to change buffer when executing a macro, but there is no purpose on this, besides it uses a poor implementation that hooks on post-command-hook that is not a good practice imho. Emacs has this feature of allowing macros to be able to do pretty much anything i dont think its wise to limit it purposefully. If you like i can write a PR to solve this issue, i even have a variable name to get this option: extremely-evil-abort-macro

edkolev:
It took me a while to get it, but I finally did:

Evil aborts macro recording when buffer is changed. Vim doesn't abort, neither does emacs (as far as I understand).

Git blaming shows that the `evil-abort-macro` was added because of this issue https://github.com/emacs-evil/evil/issues/568

greboide:
hmm i see, so the original issue was vim macros were being restarted after changing buffers? That is odd behavior, but instead of fixing it they added a workaround that prevents buffer changing, i think this should be properly fixed...

greboide:
but just switching off evil macro abort i cant reproduce the original issue...

expez:
The following pattern is extremely common and no longer works:

1. Grep for something that needs to be fixed in multiple places
2. Jump to first occurrence
3. Record macro while fixing the problem
4. Ends macro after jumping to next occurrence
5. Inspect next occurrence and conditionally replay the macro or jump again.

The macro recording ends while jumping between the grep buffer and the buffer to modify :/

edkolev:
The fix should be quite simple though, someone just needs to carefully remove parts of the code around the `evil-abort-macro` function.

edkolev:
#927 is merged, this can be closed.

[...]

Issue #568, opened 2015-08-24
Macro doesn't apply to multiple buffers. Vim does.


TheBB:
Originally reported by: **ReneFroger (Bitbucket: [ReneFroger](https://bitbucket.org/ReneFroger), GitHub: [ReneFroger](https://github.com/ReneFroger))**

----------------------------------------

In Vim, start in the buffer A with recording a marco with `qq`.
In buffer B, I end the macro with `q`. The macro is saved. 

In Evil, I start in the buffer A with recording a marco with `qq`.
In buffer B, I try to end the macro with `q`. But instead this default behaviour, a new marco will be started, which differs from Vim. 

----------------------------------------
- Bitbucket: https://bitbucket.org/lyro/evil/issue/568

PR #927, opened 2017-09-29
Don't abort macro recording when current buffer is changed

edkolev:
Refs #847

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -2021,15 +2021,6 @@ The return value is the yanked text."
 (defvar evil-macro-buffer nil
   "The buffer that has been active on macro recording.")
 
-(defun evil-abort-macro ()
-  "Abort macro recording when the buffer is changed.
-Macros are aborted when the the current buffer
-is changed during macro recording."
-  (unless (or (minibufferp) (eq (current-buffer) evil-macro-buffer))
-    (remove-hook 'post-command-hook #'evil-abort-macro)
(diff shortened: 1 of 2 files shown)


edkolev:
This PR removes `evil-abort-macro` which is aborts macro recording when the user switches to another buffer. Aborting on buffer change limits the usefulness of macros.

This function seems to be added in the first place to address this issue https://github.com/emacs-evil/evil/issues/568 however it doesn't seem relevant anymore.

wasamasa:
Code looks good, what I'm mostly wondering about is the concerns voiced in the original ticket.  I know that keyboard macros are just keys, but if it's possible to enter undefined state, there'd need to be more precautions when recording them (such as resetting some variables upon a buffer change?).

edkolev:
I played a bit with switching buffers while also recording a macro, plus leaving a buffer in insert state. However, I can't come up with a situation where evil misbehaves.

@wasamasa could you give me some guidelines what to try out to possibly break evil? If I can't reproduce evil entering an undefined state, I most probably won't be able to figure out what variables to reset.

wasamasa:
Not really, no.  I guess we'll have to do it like with the search thing and fix it if people open issues for it.

edkolev:
OK, I'll be keeping a close eye on any reported issues.

wasamasa:
Alright, thanks.

PR #929, opened 2017-10-02
Update README: goto-chg.el URL

agsdot:


wasamasa:
Thanks!

agsdot:
No problem. You're welcome.

ninrod:
in the same vein, is undo-tree.el under version control somewhere?

wasamasa:
It has always been versioned with Git, see http://www.dr-qubit.org/undo-tree.html.

PR #931, opened 2017-10-04
Handle empty buffer when shifting left/right

### LINKED ISSUE #930: Empty buffer or on last empty line, insert, C-t or C-d, Wrong type argument: number-or-marker-p, nil
### Issue type
- Bug report

### Environment

Emacs version: `GNU Emacs 26.0.50 (build 4, x86_64-w64-mingw32) of 2017-08-07`
- To make sure that it wasn't an Emacs 26 issue, I also tested it in: 
`GNU Emacs 25.3.1 (x86_64-w64-mingw32) of 2017-09-12`, and it shows same results.

Operating System: `Windows 10 Version 1607 (OS Build 14393.1715)`
Evil version: `evil-git-0e06bbaf` (directory name: `evil-20171002.135`)
Evil installation type: MELPA (installed/updated in Spacemacs)
Graphical/Terminal: Graphical

### Reproduction steps
- Start `runemacs --quick`
- `M-x` `package-initialize`
- `M-:` `(require 'evil)`
- `M-x` `evil-mode`
- With the cursor in an empty buffer, or on the buffers last line (that's empty): press `C-t`

### Expected behavior
The line should shift to the right, like it does on any other line (whether it's empty or not).

### Actual behavior
The cursor doesn't move, and the messages buffer shows:
`evil-shift-right-line: Wrong type argument: number-or-marker-p, nil`

### Possible cause
I followed the execution in the `evil-shift-right` function, and noticed that the while condition 
`(< (point) end)` doesn't evaluate to `t`, when the cursor is on the buffers last line and it's empty.

The `first-shift` variable gets assigned inside the while loop, but it's used after the while loop in the last line:
`(move-to-column (max 0 (+ pnt-indent first-shift))))))`

In this case `first-shift` has the value `nil`, and that makes the `max` function return: 
`wrong-type-argument number-or-marker-p nil`.

I'm not sure how this should be solved, so I'll leave it to the professionals :)

### Further notes
While looking through the `evil-shift-right` function, I noticed that the count variable is set before the let expression:
```
  (interactive "<r><vc>")
  (setq count (or count 1))
  (let ((beg (set-marker (make-marker) beg))
```
does this make it a global variable?
Maybe it should be defined in the let expression instead.

#### Note 2
Here's a description of how I discovered that the issue might be occurring in a `max` function.

It might not be relevant to this issue, but it's something I noticed, and someone else might find it interesting.

I was going to test this issue in Emacs 25.3, and repeated the steps above, but when I called 
`M-x` `evil-mode`, then this message appeared in the minibuffer:
`Invalid byte opcode: op=183, ptr=109`

It might be caused by a package that's compiled in Emacs 26, will make it backwards incompatible, because when all the `.elc` files in the `evil-20171002.135` directory, are deleted and Emacs 25.3 is restarted. Then the invalid byte opcode message doesn't appear again. (byte recompiling the evil directory in Emacs 25.3, also makes calling `evil-mode` work as expected).

Without the `.elc` files in the evil directory, then the steps above shows `max` before the wrong type message:
`max: Wrong type argument: number-or-marker-p, nil`
instead of the `evil-shift-right-line` function name.

Both error messages seem to be useful, when the evil directory has been compiled, then the error message seems to show the function name (`evil-shift-right-line`) that was called, but when there are no `.elc` files, then it seems to show the function name (`max`) that output the error message.

wasamasa:
Closes #930

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -1812,6 +1812,12 @@ See also `evil-shift-left'."
             (indent-to new-indent 0))
           (delete-region (point) (progn (skip-chars-forward " \t") (point)))
           (forward-line 1))))
+    ;; in case we're in an empty buffer first-shift is still unchanged
+    (unless first-shift
+      (if (< count 0)
+          (setq first-shift 0)
+        (setq first-shift (* count evil-shift-width))
+        (indent-to first-shift)))
     ;; assuming that point is in the first line, adjust its position
     (if (called-interactively-p 'any)
         (evil-first-non-blank)


duianto:
Confirmed. This solves #930, `C-t` works in an empty buffer or on a buffers last empty line.

PR #934, opened 2017-10-06
Extend error message when folding isn't supported

### LINKED ISSUE #932: misleading error message on `za` without running a supported folding minor mode
Could we rename 
https://github.com/emacs-evil/evil/blob/ca6e155878c4a398fa4c297ea8b6a329cfc40ca9/evil-commands.el#L2696
to say something like

 "Folding is not supported for any of these major/minor modes. You may want to try M-x hs-minor-mode RET first, or see variable `evil-fold-list'."

I at first thought evil had some built-in folding that happened to not support any of the stuff I tried. A slightly improved message would make this more discoverable.

edkolev:
Fix #932

Sample error message:
`Folding is supported for these major/minor modes: hs-minor-mode, hide-ifdef-mode, outline-mode, origami-mode`

Let me know if the message can be improved.

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -2693,7 +2693,8 @@ Handler errors will be demoted, so a problem in one handler will (hopefully)
 not interfere with another."
   (if (null list)
       (user-error
-       "Folding is not supported for any of these major/minor modes")
+       "Enable one of the following modes for folding to work: %s"
+       (mapconcat 'symbol-name (mapcar 'caar evil-fold-list) ", "))
     (let* ((modes (caar list)))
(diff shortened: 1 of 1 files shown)


wasamasa:
Hm, this doesn't make clear it's an error and is merely informational.  Why not something like "Enable one of the following modes for folding to work: ..."?

edkolev:
The message is now modified, here's a sample: `Enable one of the following modes for folding to work: hs-minor-mode, hide-ifdef-mode, outline-mode, origami-mode, vdiff-mode`

wasamasa:
Alright, thanks.

PR #937, opened 2017-10-09
Fixed evil-yank-line-handler and evil-yank-block-handler when yank-excluded-properties is t

jeremies:
Some major modes in emacs have variable yank-excluded-properites as t that means to discard all text properties when yanking. With the old evil-yank-line-handler function this case is not correctly implemented, only discards text properties when yank-excluded-properties is a list. With the old function there is a bug that with web-mode and evil-mode activated in a buffer, when you yank whole line with `yy` and then paste the yanked line, if you copy a piece of text of the yanked line and paste it, the pasted text appears in a new line and truncated. With this commit this bug is fixed.

### DIFF
--- a/evil-common.el
+++ b/evil-common.el
@@ -2478,8 +2478,10 @@ The tracked insertion is set to `evil-last-insertion'."
   "Inserts the current text linewise."
   (let ((text (apply #'concat (make-list (or evil-paste-count 1) text)))
         (opoint (point)))
-    (remove-list-of-text-properties
-     0 (length text) yank-excluded-properties text)
+    (if (eq yank-excluded-properties t)
+        (set-text-properties 0 (length text) nil text)
+      (remove-list-of-text-properties 0 (length text)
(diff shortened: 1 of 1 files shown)


wasamasa:
Thanks, this looks good!  I've checked the sources and found that there's one more place in the code using `remove-list-of-text-properties` in combination with `yank-excluded-properties`, it's in `evil-yank-block-handler`.  Perhaps it would be better to write a helper function that removes yank properties and use it for both.

evil-common.el
```diff
@@ -2474,12 +2474,18 @@ The tracked insertion is set to `evil-last-insertion'."
       (unless (eq register ?_)
         (kill-new text)))))
 
+(defun evil-remove-yank-excluded-properties (text)
+  "Removes `yank-excluded-properties' from text."
```
wasamasa: Please use allcaps for references to function arguments

wasamasa:
Thanks!

PR #942, opened 2017-10-15
Remove reference to missing arguments FORWARD and BACKWARD from doc

### LINKED ISSUE #838: fix docstring of evil-select-inner-object
there is no forward or backward named argument for the function `evil-select-inner-object`, 
so, if I'm not mistaken, they should not appear on the [docstring](https://github.com/emacs-evil/evil/blob/master/evil-common.el#L2942-L2951).

```elisp
(defun evil-select-inner-object (thing beg end type &optional count line)
  "Return an inner text object range of COUNT objects.
If COUNT is positive, return objects following point; if COUNT is
negative, return objects preceding point.  FORWARD is a function
which moves to the end of an object, and BACKWARD is a function
which moves to the beginning.  If one is unspecified, the other
is used with a negative argument.  THING is a symbol understood
by thing-at-point. BEG, END and TYPE specify the current
selection. If LINE is non-nil, the text object should be
linewise, otherwise it is character wise."
```

barrucadu:
The docstrings for `evil-select-inner-object` and `evil-select-an-object` reference the arguments FORWARD and BACKWARD, which don't exist.  This PR just removes them.

Fixes #838

### DIFF
--- a/evil-common.el
+++ b/evil-common.el
@@ -2946,12 +2946,10 @@ This can be overridden with TYPE."
 (defun evil-select-inner-object (thing beg end type &optional count line)
   "Return an inner text object range of COUNT objects.
 If COUNT is positive, return objects following point; if COUNT is
-negative, return objects preceding point.  FORWARD is a function
-which moves to the end of an object, and BACKWARD is a function
-which moves to the beginning.  If one is unspecified, the other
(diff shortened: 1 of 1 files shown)


wasamasa:
Thanks!

PR #943, opened 2017-10-15
Add evil-paste-{after,before}-adjusting-indent

kljohann:
This is an attempt to finish a patch I sent to the vim-emulation mailing list in 2014.
The intention is to add the corresponding functionality available in vim:
- `]p` is like `p` but adjusts indentation to the current line
- `]P`/`[p`/`[P` is like `P` but adjusts indentation to the current line

I just decided to turn this into a PR now so that I do not forget about it and we can discuss the implementation. What currently is missing from this change is support for `evil-paste-pop`.


kljohann:
Well, I only just saw 896, so I will compare our implementations and see what I can contribute to that PR :)

kljohann:
I think the approach in this PR is more similar to the original vim functionality, since it just does relative re-indentation of the pasted block instead of calling `evil-indent` (which re-indents the region according to the rules of the major-mode). So I would prefer this behavior over #896.

edkolev:
If the behaviour is identical to what #896 does, I think 896's approach is better since it involves less code.

The only thing the other PR lacks is tests. My guess is if you add tests and maybe simplify the code to use `evil-indent`, this PR will be easily merged. Making `C-p`/`C-n` work after `[ p` can be handled later if at all (vim doesn't have this functionality).

evil-commands.el
```diff
@@ -2011,6 +2011,65 @@ The return value is the yanked text."
   (interactive)
   (evil-paste-from-register ?.))
 
+(defun evil--get-common-indent-in-region (start end)
+  ""
```
wasamasa: Why bother putting a docstring there if it's empty?

evil-commands.el
```diff
@@ -2011,6 +2011,65 @@ The return value is the yanked text."
...
+        (unless (eolp)
+          (setq indentation
+                (if indentation
+                    (min (current-column) indentation)
+                  (current-column))))
+        (forward-line 1)))
+    (or indentation 0)))
+
+(defun evil--adjust-indent-afer-paste (&optional command)
+  ""
```
wasamasa: Same as above.  An empty docstring is worse than none.

wasamasa:
@kljohann I don't believe you need docstrings for functions with double dashes, but if you insist, you can add them. Other than the lack of tests this looks pretty good!

kljohann:
> If the behaviour is identical to what #896 does, I think 896's approach is better since it involves less code.

Here is a simple example that works using this PR but where #896 does not reproduce vim's behavior:

Let's assume we copied
```python
for x in [1, 2, 3]:
    if True:
        pass
else:
    print("x")
```

Ant try to paste it using `]p` (paste after) here:
```python
if False:
    pass # ← here
```

The correct result (achieved using this PR) would be:
```python
if False:
    pass
    for x in [1, 2, 3]:
        if True:
            pass
    else:
        print("x")
```

However #896 produces the following:
```python
if False:
    pass
for x in [1, 2, 3]:
    if True:
        pass
else:
    print("x")
```
which is not what we want.

VanLaser:
If this is closer to vim, IMHO it should be preferred. I guess in modes like python where indentation can't be guessed by the major, this PR does "less" (which is better) because "it just does relative re-indentation of the pasted block instead of calling evil-indent (which re-indents the region according to the rules of the major-mode)." (to quote the OP). I'm speaking just from these comments, haven't looked how the code is written.

wasamasa:
I'm generally in for more correct behavior, so I prefer this PR over #896.

kljohann:
When implementing tests I ran into an issue with the `yank-handler` argument to `evil-paste-{before,after}`  (something similar has also mentioned by @VanLaser in [the other issue](https://github.com/emacs-evil/evil/pull/896#issuecomment-321523829)):
While they check whether the argument is present and callable, they defer the actual pasting to `insert-for-yank` from `subr.el`, which only looks at the text properties to determine the yank handler.

However, I have to override the yank handler to enforce line-wise pasting and correctly reproduce the following behavior in vim:
```
uiae
    aaabbb
  xxx
```
blockwise selection of the first two lines, placing the cursor on the first b and pressing `]p` should result in
```
uiae
    aaabbb
    uiae
        aaabbb
  xxx
```
to correctly mirror vim's behavior. Thus we need to force line-wise pasting.

wasamasa:
Hm, the reason this strategy works out for the existing commands (like `evil-yank-lines`) is because they put the yank-handler property on the text to be yanked.  Perhaps you'll have to do the same, although it makes me wonder what exactly the argument is for.

kljohann:
I came up with a work-around involving `(propertize text 'yank-handler (list yank-handler))` in `evil-paste-{after,before}` to pass the yank-handler to `insert-for-yank` (overriding any other yank-handler that may be set via properties ← I'm not sure whether this is okay for all cases) and conditionally adding a trailing newline in `evil--paste-and-adjust-indent-yank-handler` (as evil-yank-line-handler assumes that the text to be yanked ends in a newline).

kljohann:
@wasamasa Can you take another look?

evil-commands.el
```diff
@@ -2021,6 +2023,58 @@ The return value is the yanked text."
...
+  "Return the minimum indent (in columns) between START and END ignoring empty lines."
+  (let (indentation)
+    (save-excursion
+      (goto-char start)
+      (or (bolp) (forward-line 1))
+      (while (< (point) end)
+        (skip-chars-forward " \t")
+        (unless (eolp)
+          (setq indentation
+                (if indentation
```
dzop: I think it would be cleaner to set `indentation` to 0 in the `let` form above and use `max` to get rid of this `if` branch. Then you can also remove the `or` in the return value and just return `indentation`.

NyiNyiNai:
Ok

evil-commands.el
kljohann: But would it still compute the same thing? The goal is to return the _minimum_ indentation, ignoring lines that contain only whitespace. Setting `indentation` to `most-positive-fixnum` and using `min` would work, but is it much more readable? I guess it depends on what should be returned if the region consists only of whitespace (currently `0` is returned).
```elisp
  (let ((indentation most-positive-fixnum))
    (save-excursion
      (goto-char start)
      (or (bolp) (forward-line 1))
      (while (< (point) end)
        (skip-chars-forward " \t")
        (unless (eolp)
          (setq indentation
                (min indentation (current-column))))
        (forward-line 1)))
    (if (= indentation most-positive-fixnum)
        0
      indentation))
```
(If you prefer this version, just say so then I'll gladly change it. :slightly_smiling_face:)

dzop: Whoops your right, I wasn't paying attention to the loop. Sorry for the noise.

No it's not much more readable with the newer version, it just moves the if branch somewhere else. Although it is slightly more efficient as it moves the branch outside of the loop.

Issue #897, opened 2017-08-09
Add option for `v$` to exclude new line?


quicknir:
# Issue type
- Enhancement request

# Environment

Emacs version: 25.1
Operating System: Name and version, e.g. Windows 10, macOS Sierra, Debian Jessie 
Evil version: evil-git-bff206a
Evil installation type: unsure (spacemacs)
Graphical/Terminal: graphical
Tested in a [clean environment](https://github.com/emacs-evil/evil/blob/master/CONTRIBUTING.md#bug-reports): No

# Reproduction steps
- Start Emacs
- Go to a line with some text
- try to do `y$` to copy some text, and paste
- it doesn't include newline
- try to do `v$y` to copy some text, and paste
- it does include newline

This isn't a bug report because vim behaves the same way. That said, I find this behavior to be unintuitive, surprising, and strictly less useful than if `v$` behaved consistently with `y$`. It would be amazing if there was an option added that enabled this. I would think that this shouldn't be too bad to implement, because the behavior of `$` is *currently* inconsistent depending on which operator it is used with. I would assume then that there must be some kind of special treatment of `$` when preceded by `v`. All that's needed is an option to disable that special treatment.

wasamasa:
Isn't this about visual mode being inclusive?  Wouldn't something like `:set selection=exclusive` make more sense?

quicknir:
@wasamasa Selection exclusive is an option with much more far reaching consequences, and having played with that option in vim that's definitely not what I want. And no, this is not about visual mode being inclusive. If you try this with any other motion, you will not observe the same thing. `yfx` and `vfxy` will yank exactly the same thing. The same is true for `w`, `e`, `t`, etc almost any motion. Why should `$` be special?

quicknir:
Also as a follow up I've put this change into effect locally and it seems to work, I'll report here if I see any unexpected behavior.

wasamasa:
Thanks, but no thanks.

ninrod:
am I missing something? Why not just use `vg_`?

quicknir:
@ninrod That's not the same thing, if there's trailing whitespace it will move before it. Also, `vg_` is not something I ever need to use, I don't see a reason to use more muscle memory on `vg_`. Flipping the question around: when is the last time you found `v$` grabbing the newline useful? I've never wanted it to do that so I find it a pure win. Obviously others may disagree, but that's why I proposed it as an option (and a non-default one at that).

ninrod:
[...]

PR #948, opened 2017-10-22
Allow user to control whether v$ takes newline

quicknir:
The default is set so that existing behavior is unaffected. It is true that the existing behavior is consistent with vim, but vim's behavior here is quite self inconsistent. If you do `d$` or `D`, or `y$` or `Y`, you don't grab the newline. But if you do `v$`, you do. So `v$d` is different from `d$`, which is weird. This is also very annoying when using surround, as wanting to surround something to the end of the line via `v$S)` is very common.

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -174,7 +174,7 @@ If COUNT is given, move COUNT - 1 lines downward first."
   (when evil-track-eol
     (setq temporary-goal-column most-positive-fixnum
           this-command 'next-line))
-  (unless (evil-visual-state-p)
+  (unless (and  (evil-visual-state-p) evil-v$-gets-eol)
     (evil-adjust-cursor)
     (when (eolp)
       ;; prevent "c$" and "d$" from deleting blank lines
--- a/evil-vars.el
+++ b/evil-vars.el
@@ -216,6 +216,11 @@ a line."
   :type 'boolean
   :group 'evil)
 
+(defcustom evil-v$-gets-eol t
+  "Whether when v$ is performed, the new line character is taken as well"
+  :type 'boolean
+  :group 'evil)
+
 (defcustom evil-respect-visual-line-mode nil
   "Whether to remap movement commands when `visual-line-mode' is active.
 This variable must be set before evil is loaded. The commands


quicknir:
See https://github.com/emacs-evil/evil/issues/897.

quicknir:
I have no idea why this check is failing, or what it means?

edkolev:
> This is also very annoying when using surround, as wanting to surround something to the end of the line via v$S) is very common

I barely use visual mode at all, I wouldn't call it common :) Having followed /r/vim for some time, I would say some people use it regularly, some people don't.

You could use `ys $ )` to achieve the behaviour you need.

As mentioned in #897, maybe it would be a better idea to add an option like `evil-selection` which could either be `inclusive/exclusive` instead of just targeting the `$` motion.

wasamasa:
I hardly use visual mode either beyond the basic motions for extending the region. Considering that `ys$)` does the right thing and it's just you wanting this behavior, I don't see any reason to add this option.

quicknir:
Closing this, at this point, feels awfully heavy handed. @edkolev I didn't say such usage was the majority; if some people use it often and some don't it is still common, i.e. not rare.

Suggesting `ys$)` doesn't say anything about the value of this patch (nor does saying that "it's just me" wanting it). As I mentioned in the issue, the fundamental point is that `$` behaves inconsistently; `ys$)` may do the "right thing" in this particular case, but `$` does the wrong thing in general.

One of the key ideas in vim is that you can combine operators and motions as you please. If a motion did something different when used with each operator, then this whole scheme would be worthless. The fact that operators are consistent, is the reason why `v<motion><operator>` does the same thing as `<operator><motion>`, in 99.9% of cases.

The fact that `$` behaves differently depending on the operator, treating `v` differently from all other operators, is in my view just a bug in `$` ( @edkolev , this is also why the suggestion of changing selection to inclusive exclusive is *not* a good one. other motions are fine, it's a problem specifically with `$`, which is why we need a solution that targets `$` specifically). Most likely vim maintainers would not admit this is a bug, but it is clearly an inconsistency with no real benefit. It is similar in many respects to the mapping of `Y` to `yy`; evil provides an explicit variable to handle this even though vim does not, because it makes more sense, and you may as well make such things more logical.

This patch is tiny, backwards compatible, and allows enabling behavior that makes evil's behavior strictly more internally consistent (albeit less consistent with vim). I don't see any substantial downside and definitely nothing tha thas been presented here.

ninrod:
am I missing something, why not just use `vg_`?

nixmaniack:
Can we please reconsider this patch? I already [commented on related thread](https://github.com/emacs-evil/evil/issues/897#issuecomment-359352813) but thought of voicing on this as patch is already there as a PR.

Issue #926, opened 2017-09-26
Suggestion: autoload ex commands


edkolev:
# Issue type

- Question / Suggestion

# Further notes

I've been thinking about autoload-ing ex commands, for example `evil-ex-global`. This is how I see things would work:
1. in evil-maps.el, the `:global` command will be registered `(evil-ex-define-cmd "g[lobal]" 'evil-ex-global)`
2. (new file) evil-ex-commands.el will define `evil-ex-global` with an autoload cookie; Note: this new file will not be required by evil-maps.el
3. in evil-ex.el, when the minibuffer completion is active, function `evil-ex-argument-completion-at-point` will check if the ex command `:global` has an autoloaded binding with `autoloadp` and, if yes, force the autoload with `autoload-do-load`

Notes:
- Step 3. is necessary because without it evil-ex-global's interactive statement `(interactive "<r><g/><!>")` wouldn't be processed
- Step 3. could be modified to force the autoload only if the binding (`evil-ex-normal` in this example) starts with "evil"; this will prevent forcing autoload of non-evil commands, e.g. `:checkdoc`
- I'm not sure exactly how this will work if the end user doesn't generate autoloads (using `borg` or some other means of accumulating packages)

So my question is: would you consider this a good addition to evil? I'm willing to contribute this change

wasamasa:
Not sure if I'm not reading this properly, but I'm missing out on the why behind this addition.  Autoloads are intended to be used on the main entry points of a package and in the case of a major/minor mode it would precede that definition.  Once the mode is enabled, everything else is loaded alongside with it.  How would adding more autoloads improve things?

edkolev:
The reason I got thinking about this in the first place was that quite a few ex commands are defined (like :sort, :normal, etc) when `(evil-mode)` is executed but they might end up never being used by the user. I thought that it would make sense to delay the loading of the ex commands until used.

[...]

PR #949, opened 2017-10-24
If an autoload, load ex function so completion can be provided

edkolev:
With this PR evil's ex completion will load an autoloaded function before providing completion.

For example, the following scenario will work:
1. external package [evil-expat](https://github.com/edkolev/evil-expat) will define ex command `:rename NEW-NAME` (which moves/renames the current file and its buffer) with an autoload cookie like  so:
```emacs-lisp

(evil-define-command evil-expat-rename (bang new-name)
  "Rename the current file and its buffer to NEW-NAME..."
  (interactive "<!><f>")
  ...)

;;;###autoload
(eval-after-load 'evil
  '(progn
     (evil-ex-define-cmd "rename" 'evil-expat-rename)
     (autoload 'evil-expat-rename "evil-expat" nil t)))
```
2. the evil-expat-autoloads.el file is loaded by `(package-initialize)`
3. when the user enters `:rename /tm<TAB>` the evil-expat package will be loaded and the prompt will be completed to `:rename /tmp/`

Without this PR, entering `:rename /tm<TAB>` would result to `:rename /tm`, without completing `/tm` to `/tmp`.

Please let me know if the PR can be improved.

I've previously mentioned autoloading of ex commands in #926 but then I didn't know enough about autoloading to have meaning discussion.

### DIFF
--- a/evil-ex.el
+++ b/evil-ex.el
@@ -458,6 +458,13 @@ in case of incomplete or unknown commands."
 (defun evil-ex-argument-completion-at-point ()
   (let ((context (evil-ex-syntactic-context (1- (point)))))
     (when (memq 'argument context)
+      ;; if it's an autoload, load the function; this allows external
+      ;; packages to register autoloaded ex commands which will be
+      ;; loaded when ex argument completion is triggerred
+      (let ((binding-definition (symbol-function (evil-ex-binding evil-ex-cmd))))
+        (when (autoloadp binding-definition)
+          (autoload-do-load binding-definition)))
+
       (let* ((beg (or (and evil-ex-argument
                            (get-text-property 0 'ex-index evil-ex-argument))
                       (point)))


wasamasa:
Ah, I think I finally understand what you meant in the other conversation, it's about the phenomenon that you can't do any kind of introspection on a yet to be loaded package unless you do some extra trickery to automatically load the information you need.  I previously assumed you want to modify commands Evil provides (as opposed to external commands loading up Evil) so that they can be used as an alternative entry point.  IIRC Emacs recently added extra support for this in its help system.

evil-ex.el
```diff
@@ -458,6 +458,11 @@ in case of incomplete or unknown commands."
 (defun evil-ex-argument-completion-at-point ()
   (let ((context (evil-ex-syntactic-context (1- (point)))))
     (when (memq 'argument context)
+      ;; if it's an autoload, load the function so ex completion can be provided
```
wasamasa: I'd extend this comment by an explanation how this situation happens in the first place as it's not immediately obvious.  Something like a one-sentence version of your original post would work.

edkolev:
Thanks, the comment is now updated. Let me know if it should be further clarified.

wasamasa:
Looks good now, I'll have to do some testing and then it can be merged.

wasamasa:
FWIW, I've installed evil-expat from its file, added the check to my local installation of Evil, but can't reproduce your scenario.  Printf-debugging this reveals that the `binding-definition` isn't considered an autoload which makes this fail.

edkolev:
Could you confirm the evil-expat-autoloads.el file has been generated and loaded?

This is what I get when testing in the scratch buffer:
- `(evil-ex-binding "rename")` => `evil-expat-rename`
- `(symbol-function (evil-ex-binding "rename"))` => `(autoload "evil-expat" nil t nil)`
- `(autoloadp (symbol-function (evil-ex-binding "rename")))` => `t`

Here's an alternative way to reproduce the scenario:
- create file /tmp/ttt/test.el with contents:
```emacs-lisp

(evil-define-command test (new-name)
  (interactive "<f>")
  (message "%s" new-name))

;;;###autoload
(eval-after-load 'evil
  '(progn
     (evil-ex-define-cmd "test" 'test)
     (autoload 'test "test" nil t)))
```
- run M-x update-file-autoloads, specifying test-autoloads.el
- the file test-autoloads.el should have contents something like this:
```
;;; test-autoloads.el --- automatically extracted autoloads
;;
;;; Code:


;;;### (autoloads nil "test" "test.el" (23024 20443 0 0))
;;; Generated autoloads from test.el

(eval-after-load 'evil '(progn (evil-ex-define-cmd "test" 'test) (autoload 'test "test" nil t)))

;;;***

(provide 'test-autoloads)
;; Local Variables:
;; version-control: never
;; no-byte-compile: t
;; no-update-autoloads: t
;; coding: utf-8
;; End:
;;; test-autoloads.el ends here
```
- load the test-autoloads.el file with M-x eval-buffer
- enter `:test /tm<TAB>`

wasamasa:
Ah my bad, I messed up while testing.  I can reproduce the desired effect now, `(autoloadp (symbol-function (evil-ex-binding "rename")))` returns `t` as it should and arguments are completed as expected.  Thank you for the contribution!

edkolev:
Thanks to you too!

wasamasa:
I've reached out to @TheBB to give you commit access so that you don't have to work on a fork anymore, but can instead just create branches on this project and submit PRs for them.  Feel free to commit directly to master for trivial stuff like typos, but please keep the existing workflow with reviewing PRs for anything else.

edkolev:
Great, thank you, I'll keep the workflow.

PR #950, opened 2017-10-25
Extended docstring for evil-find-char.

triklsbg:
Documented the dependency from 'evil-cross-lines' custom variable.


wasamasa:
Thanks!

PR #951, opened 2017-10-27
Open url via evil-ret

amosbird:


wasamasa:
Why is this needed in addition to the previous code that allows clicking widgets and buttons? If you wish to make URLs clickable, use `goto-address-mode`.  Heck, you're using `goto-address-at-point` from that package...

If I search for "vim open link", I get Stackoverflow answers telling me to use `gx` which isn't implemented yet.  You could implement this with `browse-url-at-point`.

amosbird:
Hi, I just tried enabling `goto-address-mode` but it still doesn't make `evil-ret` open url at point.

wasamasa:
Fix it then and make a different PR for that.  I don't see any reason to merge this PR in its current form.

PR #956, opened 2017-10-30
Make evil-ex-substitute ignore read only matches.

amosbird:
This patch makes evil recognize `query-replace-skip-read-only` variable and ignore read only matches while substituting. This is useful in wgrep and wdired mode.

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -3383,83 +3383,85 @@ resp.  after executing the command."
           (goto-char beg)
           (catch 'exit-search
             (while (re-search-forward evil-ex-substitute-regex end-marker t)
-              (let ((match-str (match-string 0))
-                    (match-beg (move-marker (make-marker) (match-beginning 0)))
-                    (match-end (move-marker (make-marker) (match-end 0)))
-                    (match-data (match-data)))
-                (goto-char match-beg)
-                (setq match-contains-newline (string-match-p "\n" match-str))
-                (setq zero-length-match (= match-beg match-end))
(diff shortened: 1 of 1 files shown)


wasamasa:
@justbur What do you think of this one? The only thing I can come up with is that this check might be better placed right before the replacement or inside the replacement function.

justbur:
I don't have strong feelings about this, but this appears to me to be the right place to check this (on the initial search instead of the replace). 

However, I think it's a little simpler and easier to understand if the condition checked as part of the `while` statement like this

```elisp
(while (and (re-search-forward evil-ex-substitute-regex end-marker t)
            (not (and query-replace-skip-read-only
                      (text-property-any (match-beginning 0) (match-end 0) 'read-only t))))
...)
```

amosbird:
@justbur hmm that won't work as expected. The while loop will exit prematurely.

justbur:
Yes of course. Then can you use when instead of if at least?

amosbird:
OK.

amosbird:
@justbur Hi, what do you think of this :) Can it be merged ?

justbur:
@amosbird It looks ok to me, but I haven't been merging PRs for this repo.

amosbird:
@wasamasa so can we merge this :) ?

wasamasa:
Yeah, it's fine now (save the merge conflict, but I resolved it).

justbur:
There's a parentheses problem now. See #986

PR #959, opened 2017-11-04
CI: use "emacs-travis" instead of EVM

edkolev:
This PR is an attempt to fix the CI segfaults that we've been dealing with.

The main change is that EVM (and its binaries) is not used. Instead, the emacs binaries from this repo are used https://github.com/npostavs/emacs-travis

These are the same binaries used by magit since this PR https://github.com/npostavs/emacs-travis/releases
For reference, here's magit's .travis.yml https://github.com/magit/magit/blob/master/.travis.yml

I'm not sure if the segfault-ing will be completely fixed, however, I've had about 10 successful successive CI runs here
https://github.com/edkolev/evil/commits/ci-experiment

Some notes:
- the CI jobs run for roughly the same time, ~1 min
- this PR adds a check for emacs 25.2 and emacs "master"
- emacs 25.1 is still in the "allow_faulires" group - I've seen it fail with this CI setup; I think it failed 3 times out of the 10 runs

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,23 @@
 language: generic
 sudo: false
 
+env:
+  global:
+    - CURL="curl -fsSkL --retry 9 --retry-delay 9"
+  matrix:
+  - EMACS_VERSION=24.3
+  - EMACS_VERSION=24.4
+  - EMACS_VERSION=24.5
+  - EMACS_VERSION=25.1
+  - EMACS_VERSION=25.2
+  - EMACS_VERSION=master
+
+matrix:
+  allow_failures:
+    - env: EMACS_VERSION=master
+    # Emacs 25.1 segfaults when running make test
+    - env: EMACS_VERSION=25.1
+
 # NOTE: flyspell test
 addons:
   apt:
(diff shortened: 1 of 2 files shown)


wasamasa:
Thank you! I hope this will make the CI less of an impediment in our workflow. I've already considered digging through the commits between 25.1 and 25.2 so that I could hopefully find a patch that makes the segfaults for 25.1 go away, that might fix the issue completely.

wasamasa:
Hm, seems it's not quite bullet-proof yet: https://travis-ci.org/emacs-evil/evil/jobs/294445092

wasamasa:
Blech, unlike with the previous ones, `addr2line` only gives me question marks. Perhaps we'd be better off with our own builds that contain debug symbols...

edkolev:
> Hm, seems it's not quite bullet-proof yet: https://travis-ci.org/emacs-evil/evil/jobs/294445092

This seems to be an old CI job, executed with EVM - I see it's configuration has EVM in it https://travis-ci.org/emacs-evil/evil/jobs/294445092/config

I don't think we can make Travis run for old commits, while using the latest .travis.yml.
I think we need to wait and see if the CI jobs executed from now on will segfault or not.

wasamasa:
Alright, we'll see then and if segfaults still happen, I'll look into replacing the binaries or something.  For fun I set up an Ubuntu Trusty VM and tried reproducing the segfaults with 24.5 there, but had no luck, no matter whether I used the binaries of the old CI, the new CI or freshly built ones.  I suspect there's more going on in Travis environment than that.

PR #961, opened 2017-11-06
Create interactive autoloads for evil(-local)-mode

### LINKED ISSUE #733: Won't be autoloaded evil-local-mode
Originally reported by: **pluser NA (Bitbucket: [pluser](https://bitbucket.org/pluser), GitHub: [pluser](https://github.com/pluser))**

When installed from MELPA, evil-mode is registered to autoload, but evil-local-mode is not.
So I evaluate (evil-local-mode t) without loading evil package explicitly, and an error is return.

```
#!lisp

;;;###autoload (autoload 'evil-mode "evil")
```
in evil-core.el

In my humble opinion, it may be good idea to change it to

```
#!lisp

;;;###autoload (progn (autoload 'evil-mode "evil" nil t) (autoload 'evil-local-mode "evil" nil t))
```

or something.

- Bitbucket: https://bitbucket.org/lyro/evil/issue/733

fice-t:
Fixes #733

This lets you just `M-x evil(-local)-mode` after `(package-initialize)`.


wasamasa:
FWIW, I disagree with auto-loading `evil-local-mode`. It's not supported to use Evil locally as that skips the global initialization parts.  If you run into bugs with such a configuration, they won't be fixed, so autoloading that part would encourage more people into doing that.

The other change is fine though.

fice-t:
That would explain why I was having trouble with `ESC` in `evil-local-mode` (I was using for it since using `evil-mode` seems to override `edebug`'s bindings).

Is it not possible to have an `evil-mode` equivalent that's constrained to a single buffer? That might be useful in some cases.

wasamasa:
I'm not sure what edebug has to do with this. The integration for this case is as simple as `(add-hook 'edebug-mode-hook 'evil-normalize-keymaps)`.

It's not about it being possible or not, it's about how the whole thing is engineered. If you work around that, you're on your own and will not get support.  That's all there is to it.

fice-t:
> I'm not sure what edebug has to do with this.

I just mentioned that since it was my use-case for using `evil-local-mode` directly. I don't care about it otherwise, I suppose.

> The integration for this case is as simple as (add-hook 'edebug-mode-hook 'evil-normalize-keymaps).

This works, but shouldn't this integration be done automatically? 

I see some mention of edebug in `evil-intercept-maps`, but I'm not sure what "Keymaps that should intercept Evil maps." means, and how this is different from `evil-overriding-maps`. 

Actually, adding `(edebug . nil)` to `evil-overriding-maps` fixes the issue for me.

Anyway, I removed the `evil-local-mode` autoload. Should #733 be closed as `wontfix` instead?

wasamasa:
I believe that edebug issue belongs into a separate thread, especially if it's linked to the seemingly useless intercepting keymap.  Perhaps you'll figure out if it's needed after all, broke at some point or can be deprecated.  From a quick look it seems like both intercepting and overriding keymaps are handled the same, with intercepting ones coming first, the only difference being how a map is detected as being intercepting/overriding.

I'll close #733 after merging this one.

PR #964, opened 2017-11-12
Decode mouse control sequences

### LINKED ISSUE #960: Visual mode starts erratically and cursor jumps when xterm-mouse-mode is used in a text terminal
# Issue type

- Bug report

# Environment

Emacs version: GNU Emacs 25.3.1
Operating System: FreeBSD 11 and also macOS 13 HighSierra (might be a non-OS specific issue)
Evil version: Evil version 1.2.13
Evil installation type: MELPA
Graphical/Terminal: iTerm
Tested in a `make emacs` session (see [CONTRIBUTING.md](https://github.com/emacs-evil/evil/blob/master/CONTRIBUTING.md#bug-reports)): YES

# Reproduction steps
- Start Emacs in a text terminal (not GUI) that supports mouse events. I use iTerm. Ensure mouse events are not disabled in the terminal.
- `M-x <RET> xterm-mouse-mode` to enable mouse clicks being received.
- Type some text into a buffer.
- Click repeatedly within the text with a mouse.
 
# Expected behavior
Cursor should jump to the place mouse was clicked. Visual mode should not turn on. In other words, it should work the way it works in GUI, or in text, when evil mode is off, like after pressing `C-z` and continuing to click.

# Actual behavior
The result you will see is that *visual* mode will turn itself on and off whilst selecting parts of each clicked line starting at the the left edge of the line and ending at click point. However, cursor will jump to the beginning of the line. The resulting selection has no relation to what was clicked, it seems erratic. Further, *Messages* shows errors being logged while clicking: 
```
[ < is undefined
user-error: No previous search
```

evil-emacs-xterm-mouse-mode-text-issue

# Further notes
If evil-mode is then switched off using `C-z` the issue disappears. When it is switched back on, erratic behaviour resumes.

I have also reported it on the Emacs bug list, but it was suggested that I should report it here first. See my report over there: https://debbugs.gnu.org/cgi/bugreport.cgi?bug=29143

Thank you for your help. Evil is my main editor nowadays and I appreciate your work. If I can help in any way, please let me know.

fice-t:
`xterm-mouse-mode' uses control sequences for mouse events which need to be
decoded, otherwise each character in the sequence will be executed as a command.

Fixes #960.

This change came about from [Emacs Bug#29150](https://debbugs.gnu.org/cgi/bugreport.cgi?bug=29150).

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -4202,7 +4202,7 @@ DO-MOUSE-DRAG-REGION-POST-PROCESS should only be used by
     ;; Track the mouse until we get a non-movement event.
     (track-mouse
       (while (progn
-               (setq event (read-event))
+               (setq event (read-key))
                (or (mouse-movement-p event)
                    (memq (car-safe event) '(switch-frame select-window))))
         (unless (evil-visual-state-p)


wasamasa:
Thanks!

Issue #952, opened 2017-10-27
Evil uses obsolete functions: x-set-selection and find-tag


diamond-lizard:
# Issue type

- Bug report

# Environment

Emacs version: GNU Emacs 26.0.90 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.22.16)
Operating System: Gentoo Linux amd64
Evil version: evil-git-4bae299
Evil installation type: manual
Graphical/Terminal: terminal
Tested in a `make emacs` session (see [CONTRIBUTING.md](https://github.com/emacs-evil/evil/blob/master/CONTRIBUTING.md#bug-reports)): not applicable

# Reproduction steps
`make compile`

# Expected behavior
compile without warnings

# Actual behavior
`make compile` warns about obsolete functions:
```
Compute dependencies
emacs --batch -Q -L . -L lib -f batch-byte-compile evil-vars.el
emacs --batch -Q -L . -L lib -f batch-byte-compile evil-digraphs.el
emacs --batch -Q -L . -L lib -f batch-byte-compile evil-common.el

In evil-get-register:
evil-common.el:2013:35:Warning: ‘x-get-selection’ is an obsolete function (as
    of 25.1); use ‘gui-get-selection’ instead.

In evil-set-register:
evil-common.el:2103:6:Warning: ‘x-set-selection’ is an obsolete function (as
    of 25.1); use ‘gui-set-selection’ instead.
evil-common.el:2106:9:Warning: ‘x-set-selection’ is an obsolete function (as
    of 25.1); use ‘gui-set-selection’ instead.
emacs --batch -Q -L . -L lib -f batch-byte-compile evil-core.el
emacs --batch -Q -L . -L lib -f batch-byte-compile evil-states.el
emacs --batch -Q -L . -L lib -f batch-byte-compile evil-repeat.el
emacs --batch -Q -L . -L lib -f batch-byte-compile evil-macros.el
emacs --batch -Q -L . -L lib -f batch-byte-compile evil-types.el
emacs --batch -Q -L . -L lib -f batch-byte-compile evil-ex.el
emacs --batch -Q -L . -L lib -f batch-byte-compile evil-search.el
emacs --batch -Q -L . -L lib -f batch-byte-compile evil-command-window.el
emacs --batch -Q -L . -L lib -f batch-byte-compile evil-jumps.el
emacs --batch -Q -L . -L lib -f batch-byte-compile evil-commands.el

In evil-jump-to-tag:
[...]

PR #971, opened 2017-11-18
Fix compile warning for deprecated x-set-selection, x-get-selection

edkolev:
Refs #952 

This PR addresses these compile warnings:
```
emacs --batch -Q -L . -L lib -f batch-byte-compile evil-common.el

In evil-get-register:
evil-common.el:2012:20:Warning: ‘x-get-selection’ is an obsolete function (as
    of 25.1); use ‘gui-get-selection’ instead.

In evil-set-register:
evil-common.el:2100:9:Warning: ‘x-set-selection’ is an obsolete function (as
    of 25.1); use ‘gui-set-selection’ instead.
evil-common.el:2102:9:Warning: ‘x-set-selection’ is an obsolete function (as
    of 25.1); use ‘gui-set-selection’ instead.
```

I haven't yet figured out how to fix the `find-tag` compile errors.

### DIFF
--- a/evil-common.el
+++ b/evil-common.el
@@ -47,6 +47,13 @@ window commands not available.")
 
 ;;; Compatibility with different Emacs versions
 
+;; x-set-selection and x-get-selection have been deprecated since 25.1
+;; by gui-set-selection and gui-get-selection
+(defalias 'evil-get-selection
+  (if (fboundp 'gui-get-selection) 'gui-get-selection 'x-get-selection))
+(defalias 'evil-set-selection
+  (if (fboundp 'gui-set-selection) 'gui-set-selection 'x-set-selection))
+
 (defmacro evil-called-interactively-p ()
   "Wrapper for `called-interactively-p'.
 In older versions of Emacs, `called-interactively-p' takes
@@ -2010,7 +2017,7 @@ The following special registers are supported.
                   (setq request-type (list request-type)))
                 (while (and request-type (not text))
                   (condition-case nil
-                      (setq text (x-get-selection what (pop request-type)))
+                      (setq text (evil-get-selection what (pop request-type)))
                     (error nil)))
                 (when text
                   (remove-text-properties 0 (length text) '(foreign-selection nil) text))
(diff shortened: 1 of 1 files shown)


wasamasa:
Thanks! While this works, it's not ideal yet.  If someone were to load this in an old Emacs, they'd get the impression `gui-set-selection` is a thing in it which is wrong. A better alternative would be to define a differently named wrapper, like with `(defalias 'evil-get-selection (if (fboundp 'gui-get-selection) 'gui-get-selection 'x-get-selection))`.

edkolev:
Sure, the commit is now amended.

I removed `eval-and-compile` since I didn't get warnings without it.

I tried adding `#` sign like so: 
```
(if (fboundp 'gui-get-selection) #'gui-get-selection #'x-get-selection)
```

but that would yield compile errors with and without `eval-and-load`.

edkolev:
Regarding the `find-tag` warnings - I'm not sure the fix would be as simple because `xref-find-definitions` isn't a drop-in replacement for `find-tag`.

With `xref-find-definitions` the definition of `evil-jump-to-tag` would be something like:
```emacs-lisp
(evil-define-motion evil-jump-to-tag (arg)
    "Jump to tag under point.
If called with a prefix argument, provide a prompt
for specifying the tag."
    :jump t
    (interactive "P")
    (let ((xref-prompt-for-identifier arg))
      (call-interactively #'xref-find-definitions)))
```

While the definition with `find-tag` is:
```emacs-lisp
(evil-define-motion evil-jump-to-tag (arg)
  "Jump to tag under point.
If called with a prefix argument, provide a prompt
for specifying the tag."
  :jump t
  (interactive "P")
  (if arg (call-interactively #'find-tag)
    (let ((tag (funcall (or find-tag-default-function
                            (get major-mode 'find-tag-default-function)
                            #'find-tag-default))))
      (unless tag (user-error "No tag candidate found around point"))
      (find-tag tag))))
```

wasamasa:
I've experimented a bit more and haven't been able to silence the byte-compiler for the `find-tag` case. Emacs itself circumvents this problem by ditching out code using obsolete APIs which we can't really do. Anyway, your code is good to go now and an improvement over the current status quo.

Issue #969, opened 2017-11-17
Replace associative with object style notation


earthIsAPrisonIfYouNotAwareItCanNotOut:
# Issue type

- Question

# Environment

Emacs version: 25.3
Operating System: osx
Evil version: latest version
Evil installation type: MELPA
Graphical/Terminal: graphical
Tested in a `make emacs` session (see [CONTRIBUTING.md](https://github.com/emacs-evil/evil/blob/master/CONTRIBUTING.md#bug-reports)): no

# Reproduction steps
- Start Emacs
- type `$a = $some_array['key_name'];`
- type `:%s/\['\(.\{-}\)'\]/->\1/g`

# Expected behavior

$a = $some_array->key_name;

# Actual behavior

$a = $some_array['key_name'];

# Further notes
I think the `{-}` in ex mode is something like a match group
how to express in evil?

wasamasa:
I'd be inclined to use `da}` and insert the key manually (and maybe record a macro). If you need to do it with regex, two typical ways to do it would be `%s/\[\(.*?\)\]/->\1/g` and `%s/\[\([^][]*\)\]/->\1/g`, with the latter one being more efficient as it doesn't require backtracking.

earthIsAPrisonIfYouNotAwareItCanNotOut:
> with the latter one being more efficient as it doesn't require backtracking.
anything detail appreciate

wasamasa:
http://www.regular-expressions.info/catastrophic.html

earthIsAPrisonIfYouNotAwareItCanNotOut:
# Reproduction steps
- Start Emacs
- type `ABC`
- type `:%s/\([^][]*\)/\1->/g`


# Expected behavior

ABC->

# Actual behavior

ABC->->

# Further notes
the `[^][]*` can be replaced with [\d\D]?

wasamasa:
@justbur Could you please take a look at why the above does two instead of one replacement?

@earthIsAPrisonIfYouNotAwareItCanNotOut No, it means matching all characters except `]` and `[`. This doesn't make any sense for your example because there's no brackets involved. If you want to put something at the end of line, I'd use `:s/$/->`.

earthIsAPrisonIfYouNotAwareItCanNotOut:
sorry to trouble you again
there is something wrong with `:s/$/->`
the prompt popup can be removed
how to express "matching all possible characters" in evil (like `[\d\D]*`?

[...]

PR #972, opened 2017-11-20
Fix detection of newlines in evil-ex-substitute

justbur:
This addresses part of the problem in #969. I'll look at the rest later.

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -3439,7 +3439,11 @@ resp.  after executing the command."
                                         (not case-replace)))
                   (setq evil-ex-substitute-last-point (point)))
                 (goto-char match-end)
-                (cond ((and (not whole-line)
+                (cond ((>= (point) end-marker)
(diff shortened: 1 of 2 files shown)


wasamasa:
Hm, this doesn't improve things for me.  In a buffer with just "ABC" and no trailing newline I still get "ABC->->" after running `:%s/\([^][]*\)/\1->/g`. If I add a trailing newline, I get "ABC\n->->".

justbur:
This corrects a different problem that I noticed while checking #969. The regex `[^][]*` is matching newlines without them being explicitly specified in the match string. This change detects this situation correctly. 

I added another commit to fix the issue you were referring to

wasamasa:
Alright, thanks. Would you consider this done or do you reckon regression tests could be added? IIRC you stated last time there isn't much point to adding them.

justbur:
Added a simple test for the regexp that matches newlines without containing one. I think it's done now

PR #975, opened 2017-11-21
Remove use of caddr in evil-test-helpers.el

justbur:
Also require cl-lib

### DIFF
--- a/evil-test-helpers.el
+++ b/evil-test-helpers.el
@@ -153,7 +153,7 @@ raised.  Remaining forms are evaluated as-is.
                                    ',visual-start ',visual-end))
                                 ((eq (car-safe form) 'file)
                                  `(evil-test-file-contents ,(cadr form)
-                                                           ,(caddr form)))
+                                                           ,(car (cddr form))))
                                 ((or (stringp (car-safe form))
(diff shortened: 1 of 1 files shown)


wasamasa:
Hum, wouldn't that also make the `(eval-when-compile (require 'cl))` in `evil-common.el` obsolete?  Other than that this looks good.

justbur:
The test helpers don't get compiled usually, right? They're excluded from the melpa download for example.

wasamasa:
Sure, it's just that I'd rather eliminate `cl` everywhere and from what I can tell, that particular line doesn't do anything because I cannot spot any of its macros in the file.

justbur:
Well `cl` is included in `evil-jumps`. 

The issue is that `caddr` is not defined outside of `cl`. I can use emacs primitives instead of `cl-caddr`.

wasamasa:
No, it's `cl-lib.el` which is fine. `cl` not so much, I'd like to avoid its use in our own code if possible.

justbur:
Oh I think we’re talking about different things. Yes it seems that cl can be removed. Sorry I misinterpreted what you said.

justbur:
@wasamasa I changed it to only use primitives

PR #976, opened 2017-11-21
Generalize use of KEYMAP in evil-define-key

justbur:
In evil-define-key, allow keymap to be 'global, 'local or some other quoted
symbol, which is assumed to be a minor mode.

Using 'global, corresponds to making a global binding in the corresponding
global evil keymap (similar to evil-global-set-key). Using 'local, corresponds
to making a local binding in the corresponding local evil keymap (similar to
evil-local-set-key).

Any other quoted symbol is assumed to be the name of a minor mode, and the
binding is dispatched to evil-define-minor-mode-key.

Make evil-define-key* support global and local symbols.

Allow a list of states in evil-define-minor-mode-key.

This makes evil-define-key serve as a general keybinding function.

### DIFF
--- a/evil-core.el
+++ b/evil-core.el
@@ -960,15 +960,33 @@ to specify multiple bindings at once:
       \"b\" 'foo)
 
 If foo-map has not been initialized yet, this macro adds an entry
-to `after-load-functions', delaying execution as necessary."
+to `after-load-functions', delaying execution as necessary.
+
+KEYMAP may also be a quoted symbol. If the symbol is global, the
+global evil keymap corresponding to the state(s) is used, meaning
+the following are equivalent:
+
+    (evil-define-key 'normal 'global \"a\" 'bar)
+
+    (evil-global-set-key 'normal \"a\" 'bar)
+
+The symbol local may also be used, which corresponds to using
+`evil-local-set-key'. If a quoted symbol is used that is not
+global or local, it is assumed to be the name of a minor mode, in
+which case `evil-define-minor-mode-key' is used."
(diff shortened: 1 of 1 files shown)


wasamasa:
Thanks! Before merging, can you tell me how to test the local part of this?

justbur:
It's just a buffer local keymap for each state. You might have to invoke `evil-normalize-keymaps` before the binding works (or simply change states).

wasamasa:
Ah indeed, `(evil-normalize-keymaps)` did it.

ninrod:
@justbur, this seems huge. What does this mean for normal evil users? This should facilitate keybind management per mode, yes? Should we add an entry in readme.md to advertise this facility? 

thanks!

TheBB:
To be clear, this is not a new feature, just a different interface.

justbur:
@ninrod. It's more that I am consolidating functionality in `evil-define-key`. It makes `evil-define-key` able to handle most key binding tasks.

ninrod:
@justbur, I see. Just so I can understand: what the user could not do with `evil-define-key` which he can do now?

justbur:
@ninrod There are a couple of things that are now possible that weren't before. For example, you could not do this in one call with `evil-define-key`. 

```elisp
(evil-define-key '(normal visual) 'global
  ...)
```

You would have had to do

```elisp
(evil-define-key nil evil-normal-state-map
  ...)

(evil-define-key nil evil-visual-state-map
  ...)
```

and even then using `nil` as a state was not documented (it is now).

PR #978, opened 2017-11-22
Some more improvements for evil-define-key

justbur:
Mention the effect of a `nil` state in docstring and give a meaning to using `nil` with `'global` or `'local`.

Handle errors in `evil-define-key` by wrapping definitions in `condition-case-unless-debug`.

### DIFF
--- a/evil-core.el
+++ b/evil-core.el
@@ -944,18 +944,26 @@ A return value of t means all states."
 
 (defmacro evil-define-key (state keymap key def &rest bindings)
   "Create a STATE binding from KEY to DEF for KEYMAP.
-STATE is one of `normal', `insert', `visual', `replace',
-`operator', `motion', `emacs', or a list of one or more of
-these. The remaining arguments are like those of
+STATE is one of normal, insert, visual, replace, operator,
+motion, emacs, or a list of one or more of these. Omitting a
+state by using nil corresponds to a standard Emacs binding using
+`define-key'. The remaining arguments are like those of
 `define-key'. For example:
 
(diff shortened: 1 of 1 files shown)


wasamasa:
Thanks! I'm currently testing the new functionality and stuff like `(evil-define-key '(motion visual) nil "SPC" 'butterfly)` doesn't work.  If I look at the expansion, it defines an `evil-define-key-in-nil` function which looks wrong to me, perhaps it would be better to use `make-symbol` instead.

justbur:
Agree we could improve the function name, but the problem with your example is you need to wrap that key with `kbd`.

justbur:
Oh and nil is not a valid keymap

wasamasa:
Oh duh. I guess what I want is `(evil-define-key* '(motion visual) 'global (kbd "RET") 'butterfly)` which works as expected, however `(evil-define-key '(motion visual) 'global (kbd "RET") 'butterfly)` doesn't.  What's going wrong here?

justbur:
> however `(evil-define-key '(motion visual) 'global (kbd "SPC") 'butterfly)` doesn't. What's going wrong here?

Nothing as far as I can tell. That works for me.

wasamasa:
Sorry, I didn't apply the patch correctly. Anyway, the only thing missing now is that this added one more warning to the byte-compilation output.  You're referencing `local-set-key`, but didn't adapt it correctly as you're setting an unbound `map` argument.

justbur:
@wasamasa Thanks, I overlooked that. I think it's correct now. Note that I did an interactive rebase and force pushed.

wasamasa:
Alright.

PR #979, opened 2017-11-22
Remove cl

justbur:
@wasamasa Decided to take it one step further and remove cl. It appears correct to me, but please check.


wasamasa:
Alright.

PR #980, opened 2017-11-23
add travis, melpa and GPLv3 badges

ninrod:
this only touches 3 lines of the readme adding some standard badges.


wasamasa:
Thanks! Can you also add a MELPA Stable badge? Now that Evil can be properly installed from there, it would be nice to indicate that.

ninrod:
@wasamasa, sure thing, stable channel badge added.

wasamasa:
Alright.

Issue #709, opened 2016-09-08
Binding a key in lisp-mode-map with evil-define-key can bind it in emacs-lisp-mode-map


TheBB:
Originally reported by: **Anonymous**

----------------------------------------

Because the `evil-get-auxiliary-map` used in `evil-define-key*` doesn't set `ignore-parent` to a non-nil value, using `evil-define-key` with `lisp-mode-map` can actually end up binding the keys in `lisp-mode-shared-map`. This results in key bindings specified for `lisp-mode-map` ending up in `emacs-lisp-mode-map`.

This behavior doesn't seem very intuitive to me. Are there a lot of cases where having `ignore-parent` as nil here makes more sense? If not, I think it would be better to set it to `t` in `evil-define-key*`. Otherwise, it could potentially be nice to add a disclaimer in the docstring that the parent keymap could end up being used.

----------------------------------------
- Bitbucket: https://bitbucket.org/lyro/evil/issue/709

noctuid:
Is there any opposition to this? If not, this is a very simple change, and I'd be happy to make a pull request if it would be accepted. It's really annoying to, for example, bind a key in `image-mode-map`, have that same key end up being bound in `message-mode-map`, and then manually have to put `evil-get-auxiliary-map` beforehand in my config to prevent this. I've had this problem with several other keymaps as well.

wasamasa:
No objection here. I recall a more recent ticket about the same issue where I asked for clarification, if I find it again I'd close it in favor of this one.

noctuid:
This can be closed now.

wasamasa:
OK.

PR #982, opened 2017-11-23
Always create new auxiliary map in evil-define-key*

noctuid:
If the parent keymap already has an auxiliary keymap, ignore it. For example, if the user attempted to bind a key in emacs-lisp-mode-map before this commit, evil-define-key* could define the key in the corresponding auxiliary keymap in lisp-shared-mode-map instead of creating a new auxiliary keymap in emacs-lisp-mode-map. This resulted in keys bound in emacs-lisp-mode-map with evil-define-key potentially being bound in other keymaps that inherit from lisp-shared-mode-map (e.g.
lisp-mode-map).

Addresses #709.

### DIFF
--- a/evil-core.el
+++ b/evil-core.el
@@ -1011,7 +1011,7 @@ its behavior more predictable."
                     ((eq keymap 'local)
                      (evil-state-property st :local-keymap t))
                     (t
-                     (evil-get-auxiliary-keymap keymap st t))))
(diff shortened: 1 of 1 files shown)


wasamasa:
Looks good, but I suspect I'll have to apply this one after the revamp of `evil-define-key`.

PR #984, opened 2017-11-27
Add evil-use-derived-mode-state

justbur:
This variable allows evil-initial-state-for-buffer to use the initial state of a
parent mode if one is not found for the current major mode. For example,

(setq evil-use-derived-mode-state t)
(evil-set-initial-state 'special-mode 'motion)

makes motion state the default for all modes derived from special mode.

### DIFF
--- a/evil-core.el
+++ b/evil-core.el
@@ -283,6 +283,11 @@ See also `evil-initial-state'."
               (when (setq mode (evil-initial-state mode))
                 (throw 'done mode)))))
         (evil-initial-state major-mode)
+        ;; see `help-fns--parent-mode'
+        (when (and evil-use-derived-mode-state
+                   (symbolp major-mode)
+                   (get major-mode 'derived-mode-parent))
+          (evil-initial-state (get major-mode 'derived-mode-parent)))
         default)))
 
 (defun evil-initial-state (mode &optional default)
--- a/evil-core.el
+++ b/evil-core.el
@@ -283,11 +283,19 @@ See also `evil-initial-state'."
               (when (setq mode (evil-initial-state mode))
                 (throw 'done mode)))))
         (evil-initial-state major-mode)
-        ;; see `help-fns--parent-mode'
-        (when (and evil-use-derived-mode-state
-                   (symbolp major-mode)
-                   (get major-mode 'derived-mode-parent))
-          (evil-initial-state (get major-mode 'derived-mode-parent)))
+        ;; similar to `derived-mode-p'
+        (when evil-use-derived-mode-state
+          (let ((mode major-mode)
+                (checked-modes (list major-mode))
+                state)
+            (while (and (not state)
+                        (symbolp mode)
+                        (setq mode (get mode 'derived-mode-parent))
+                        ;; in case there's a cycle
+                        (not (memq mode checked-modes)))
+              (push mode checked-modes)
+              (setq state (evil-initial-state mode)))
+            state))
         default)))
 
 (defun evil-initial-state (mode &optional default)
--- a/evil-vars.el
+++ b/evil-vars.el
@@ -605,6 +605,21 @@ in `evil-emacs-state-modes', `evil-insert-state-modes' or
   :type  'symbol
   :group 'evil)
 
+(defcustom evil-use-derived-mode-state nil
+  "Whether to consider parent modes when deciding on initial state.
+When non-nil, if the current major mode does not have an initial
+state set for it using `evil-set-initial-state', but its parent
+mode does, then use the initial state from the parent mode. For
+example,
+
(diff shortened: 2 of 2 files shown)


VanLaser:
In this pull https://github.com/emacs-evil/evil/pull/904 I also tried to solve, following @wasamasa 's advice, the situation when a certain major-mode plays decides to play with its parent list. Not sure if it applies here. Also, I got no reaction after modifying it (no criticism, no merge). Maybe you could find something useful in it.

justbur:
Thanks @VanLaser. I didn't see your pr

VanLaser:
@justbur no problem, if any form of the functionality gets added to evil, I'm happy :) And, although I put some effort there, I don't have too much time on my hands right now, so I'm not too keen to further update or modify the PR anyway.

wasamasa:
@VanLaser Sorry about that, I'm not terribly happy with my own solution to this and never got around testing your version.

@justbur This looks like a cleaner solution and should deal with circular cases, however there's a few more things I'd like to be addressed before merging:

- For some wretched reason it's popular to define mode aliases to keep compatibility with Emacsen not having `prog-mode` or `text-mode`.  This PR won't work with, say, web-mode.  The other PR resolves this issue with a helper function.
- Why is this opt-in? I'd rather expect it to be enabled by default so that if someone experiences issues with it, they can disable it if needed.
- Tests like in the other PR would codify edge cases, but please define your own modes for them instead of messing around with `cc-mode` or `prog-mode`.
- Circular definitions should be an error, really.  While the current behavior is fine, you'll never find out a case like with https://github.com/Malabarba/paradox/issues/86 this way.

justbur:
@wasamasa 

> For some wretched reason it's popular to define mode aliases to keep compatibility with Emacsen not having prog-mode or text-mode. This PR won't work with, say, web-mode. The other PR resolves this issue with a helper function.

I'll look into web-mode, thanks

> Why is this opt-in? I'd rather expect it to be enabled by default so that if someone experiences issues with it, they can disable it if needed.

I think it *should* be the default, but I hesitate to change the behavior of someone's config without them knowing. 

> Tests like in the other PR would codify edge cases, but please define your own modes for them instead of messing around with cc-mode or prog-mode.

Alright

> Circular definitions should be an error, really. While the current behavior is fine, you'll never find out a case like with Malabarba/paradox#86 this way.

My opinion is that this is not the correct place to detect such an error. We are just searching for a better default state if one exists. It's fine for that search to fail, and I think it would be weird to throw a loud error that has little to do with evil.

wasamasa:
> I think it *should* be the default, but I hesitate to change the behavior of someone's config without them knowing.

It shouldn't change anything considering that none of the most commonly derived modes (`text-mode`, `prog-mode` and `special-mode`) are declared with an initial state.

> My opinion is that this is not the correct place to detect such an error.

What would the correct place be then? I think of this like an assert, a basic assumption (it being possible to detect the parent mode without running into a loop) that must not be violated.  If it is, do better than spinning out of control and notify the programmer about it, be it with a warning (because we know what to do in that situation) or an error (if we don't know and panic).  Perhaps displaying a warning with `message` sounds better to you?

justbur:
> It shouldn't change anything considering that none of the most commonly derived modes (text-mode, prog-mode and special-mode) are declared with an initial state.

I don't think it changes anything with the evil defaults, but if someone had `(evil-set-initial-state 'special-mode 'motion)` in their config they might notice some changes (i.e., modes that used to start in the default normal state now start in motion). 

> What would the correct place be then? I think of this like an assert, a basic assumption (it being possible to detect the parent mode without running into a loop) that must not be violated. If it is, do better than spinning out of control and notify the programmer about it, be it with a warning (because we know what to do in that situation) or an error (if we don't know and panic). Perhaps displaying a warning with message sounds better to you?

A message is fine with me. However, is is definitely not a problem caused by evil, and you're kind of inviting bug reports against evil for this sort of issue. If you want to handle it here, I can put the message in.

wasamasa:
> I don't think it changes anything with the evil defaults, but if someone had `(evil-set-initial-state 'special-mode 'motion)` in their config they might notice some changes (i.e., modes that used to start in the default normal state now start in motion).

Hm, indeed.  I couldn't find any weird example like this with GitHub's code search, but that doesn't mean there isn't one.

> A message is fine with me. However, this is definitely not a problem caused by evil, and you're kind of inviting bug reports against evil for this sort of issue. If you want to handle it here, I can put the message in.

Keep in mind that this is both fairly rare (I've only found one occurrence of it so far and it involved incorrect usage of aliases) and a bug (why would you put a circular reference there in the first place?).  If Evil exposes bugs in other software, great, those bugs can be fixed in the respective modes.

PR #985, opened 2017-11-28
Rework #984

justbur:
@wasamasa Here's a different version of #984. Changes are

  1. I removed the customization variable. I think this is sufficiently correct functionality to make it the default even if it does modify someone's config slightly. 

  2. I now teach `evil-initial-state` to look at aliases for modes. This makes a little more sense to me and simplifies the code. 

  3. I detect the circular references and print a message as requested. 
Teach evil-initial-state to look at aliases for a mode when they exist and to
handle nil for modes

Note that I still need to add tests for this.

### DIFF
--- a/evil-core.el
+++ b/evil-core.el
@@ -282,42 +282,42 @@ See also `evil-initial-state'."
             (when (and (boundp mode) (symbol-value mode))
               (when (setq mode (evil-initial-state mode))
                 (throw 'done mode)))))
-        (evil-initial-state major-mode)
-        ;; Check parent modes. Similar to `derived-mode-p'
-        (catch 'state
-          (let ((mode major-mode)
-                checked-modes)
-            (while (and mode (symbolp mode))
-              (when (memq mode checked-modes)
-                (error "Circular reference detected in ancestors of %s\n%s"
-                       major-mode checked-modes)
-                (throw 'state nil))
-              (let ((state (evil-initial-state mode)))
-                (when state
-                  (throw 'state state)))
-              (push mode checked-modes)
-              (setq mode (get mode 'derived-mode-parent)))
-            nil))
+        (evil-initial-state major-mode nil t)
         default)))
 
-(defun evil-initial-state (mode &optional default)
+(defun evil-initial-state (mode &optional default follow-parent checked-modes)
   "Return the Evil state to use for MODE or its alias.
 Returns DEFAULT if no initial state is associated with MODE.
 The initial state for a mode can be set with
-`evil-set-initial-state'."
-  (when mode
+`evil-set-initial-state'.
+
+If FOLLOW-PARENT is non-nil, also check parent modes of MODE and
+its alias. CHECKED-MODES is used internally and should not be set
+initially."
+  (cond
+   ((and mode (symbolp mode) (memq mode checked-modes))
+    (error "Circular reference detected in ancestors of %s\n%s"
+           major-mode checked-modes))
+   ((and mode (symbolp mode))
     (let ((mode-alias (let ((func (symbol-function mode)))
                         (when (symbolp func)
                           func)))
           state modes)
-      (catch 'done
-        (dolist (entry (evil-state-property t :modes) default)
-          (setq state (car entry)
-                modes (symbol-value (cdr entry)))
-          (when (or (memq mode modes)
-                    (and mode-alias
-                         (memq mode-alias modes)))
-            (throw 'done state)))))))
+      (or
+       (catch 'done
+         (dolist (entry (evil-state-property t :modes) default)
+           (setq state (car entry)
+                 modes (symbol-value (cdr entry)))
+           (when (or (memq mode modes)
+                     (and mode-alias
+                          (memq mode-alias modes)))
+             (throw 'done state))))
+       (when follow-parent
+         (evil-initial-state (get mode 'derived-mode-parent)
+                             nil t (cons mode checked-modes)))
+       (when follow-parent
+         (evil-initial-state (get mode-alias 'derived-mode-parent)
+                             nil t (cons mode-alias checked-modes))))))))
 
 (defun evil-set-initial-state (mode state)
   "Set the initial state for MODE to STATE.
--- a/evil-core.el
+++ b/evil-core.el
@@ -289,8 +289,8 @@ See also `evil-initial-state'."
                 checked-modes)
             (while (and mode (symbolp mode))
               (when (memq mode checked-modes)
-                (message "Circular reference detected in ancestors of %s\n%s"
-                         major-mode checked-modes)
+                (error "Circular reference detected in ancestors of %s\n%s"
+                       major-mode checked-modes)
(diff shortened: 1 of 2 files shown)


evil-core.el
```diff
@@ -283,20 +283,40 @@ See also `evil-initial-state'."
...
+    (let ((mode-alias (let ((func (symbol-function mode)))
+                        (when (symbolp func)
+                          func)))
+          state modes)
+      (catch 'done
+        (dolist (entry (evil-state-property t :modes) default)
+          (setq state (car entry)
+                modes (symbol-value (cdr entry)))
+          (when (or (memq mode modes)
+                    (memq mode-alias modes))
```
wasamasa: I'd change this to check whether `mode-alias` isn't `nil` first. Relying on `nil` not being in `modes` feels kind of wrong.

wasamasa:
Interestingly enough, this doesn't appear to handle circular references.  Here's some sample code:

```elisp
(define-derived-mode foo-mode prog-mode "Foo")
(define-derived-mode bar-mode foo-mode "Bar")
(put 'foo-mode 'derived-mode-parent 'bar-mode)
(foo-mode)
```

If I evaluate this in an Emacs, it displays the message *and* hangs. Make sure to run it in a separate instance.

evil-core.el
justbur: fixed

justbur:
Circular references cause bigger problems in Emacs it turns out. `derived-mode-p` easily gets caught in an infinite loop for example.

wasamasa:
You see, this is why I originally suggested to throw an error to abort execution completely.

justbur:
I still don't like throwing an error. Take your example and enable `global-paren-face-mode`. If evil thwos an error in the middle of initialization it leaves evil in a weird state (not fully initialized). Emacs still hangs due to a line [like this in paren-face](https://github.com/tarsius/paren-face/blob/master/paren-face.el#L118). 

So the result is evil is not initialized (when it could be) and Emacs hangs anyway. If you have `debug-on-error` set then you might get a better indication of what's going on I guess.

wasamasa:
I'm sorry, but I don't quite understand how this example supports your argument.  The shown code calls `derived-mode-p` which will hang if there's a cyclical reference anywhere.  How is this related to Evil?  More importantly, how is a setting that affects the initial state of buffers relevant to load-up which only touches elisp files?  It's not like your standard elisp file does nothing else than triggering as many different major modes as possible until it hits one triggering that corner case. I expect the error to happen as you're editing, open a file with a less well written major mode and run into it.  Erroring out there makes more sense to me than making Emacs unusable.

justbur:
@wasamasa I'll defer to you, but let me explain my reasoning. I have at least two minor modes (not including evil) which hang with circular references, so here is my experience

If we throw an error here:
  1. Evil's initialization errors out.
  2. Emacs hangs initializing another minor mode
  3. If I quit I'm left with a buffer in which evil is uninitialized or maybe partially initialized, but doesn't work. 

If we issue a message here:
  1. Evil's initialization goes through fine with the default state being chosen for the buffer.
  2. Emacs hangs initializing another minor mode
  3. If I quit I'm left with a buffer in which evil is in a good state. 

In either case, the message about the circular reference is there, but in the latter case I have a more functional editor (at least from the experience of an evil user).

justbur:
@wasamasa I updated this to throw an error for the circular reference thing.

wasamasa:
Thanks, this leaves some tests which can be probably adapted from https://github.com/emacs-evil/evil/pull/904.

justbur:
@wasamasa added some tests

wasamasa:
I've found a case that doesn't work correctly, if I've configured Evil to only use `prog-mode` and `text-mode` for normal state and make Emacs state the default, this doesn't give me normal state when opening a `json-mode` buffer.  It seems as if the alias detection doesn't work correctly yet.  `json-mode` derives from `javascript-mode` which is an alias for `js-mode`, so I'd expect this PR's code to detect that and check `js-mode` and its parents until encountering `prog-mode` to reach normal state.

Here's my config:

```elisp
(setq evil-default-state 'emacs
      evil-emacs-state-modes nil
      evil-insert-state-modes nil
      evil-motion-state-modes nil
      evil-normal-state-modes '(text-mode prog-mode fundamental-mode
                                          css-mode conf-mode
                                          TeX-mode LaTeX-mode
                                          diff-mode))
```

justbur:
Yep, I wasn't checking parents of the aliases, which is what your example shows. I changed the logic to do this now.

evil-core.el
```diff
@@ -282,42 +282,40 @@ See also `evil-initial-state'."
...
+      (or
+       (catch 'done
+         (dolist (entry (evil-state-property t :modes) default)
+           (setq state (car entry)
+                 modes (symbol-value (cdr entry)))
+           (when (or (memq mode modes)
+                     (and mode-alias
+                          (memq mode-alias modes)))
+             (throw 'done state))))
+       (and follow-parent (setq tmp (get mode 'derived-mode-parent))
```
wasamasa: Using `setq` in an `and` isn't terribly readable, I'd rather use its value directly (and replace the `and` by `when`).

justbur: Thanks, I realized I could easily get rid of the temporary variable altogether, so I did that too

justbur:
@wasamasa seems like the travis segfaults are back.

wasamasa:
Damnit.  Code look fine now, but I'll have to do some more testing before this is good to go.

justbur:
I just rebased and the seg fault went away, so a good sign at least...

wasamasa:
That's no proof considering it went undetected for like dozens of test runs.

wasamasa:
Alright, I think this can be merged, that is, once you've fixed this (hopefully trivial) merge conflict.  I'd fix it myself if this were a branch on here...

justbur:
@wasamasa done. I squashed two of the commits, too.

wasamasa:
Thanks!

marktran:
This PR disables `evil` in `org-mode` for me. How do get it working with `org-mode` again?

cc @justbur @wasamasa

wasamasa:
@marktran First of all, the worst it could do is enabling Emacs state (as denoted by `<E>` in the mode line) which is *not* the same as disabling Evil.  Second, I cannot reproduce that withing a `make emacs` session.  It's therefore most probably something in your init file.  Please open a more specific bug report once you have a minimal repro.

marktran:
Yes, please excuse my terminology.

I see that this PR makes `org-mode` derive its initial state from `text-mode` and my config had the following:
```
(loop for (mode . state) in '((inferior-emacs-lisp-mode . emacs)
                              (comint-mode              . emacs)
                              (eshell-mode              . emacs)
                              (occur-mode               . emacs)
                              (paradox-menu-mode        . emacs)
                              (special-mode             . emacs)
                              (sql-interactive-mode     . emacs)
                              (text-mode                . emacs))
      do (evil-set-initial-state mode state))
```

All fixed now. Thanks.

PR #989, opened 2017-12-01
Add option to ignore closed buffers when doing evil-jump.

amosbird:


wasamasa:
This PR gives me lots of question marks:

- What exactly are closed buffers?
- Isn't jumping to a closed buffer an error?
- If it is, why is this an option?
- How does this handle buffers that have been renamed with something like `mv`?
- If the answer to the above is no, why not rewrite this to track buffers instead and check with `buffer-alive-p`?

amosbird:
> What exactly are closed buffers?

a dead buffer (a buffer which has been killed)

> Isn't jumping to a closed buffer an error?

it isn't

> How does this handle buffers that have been renamed with something like mv?

in Vim, if you jumpback to an inactive buffer (file backened), and the file is gone (renamed), it creates a new buffer with the original file name. Otherwise it reloads the file into the new buffer. This also applies to buffers with buftype "nofile", i.e. a buffer with the same name will be recreated. Vim doesn't support setting a filename (buffername) for hidden buffers. When changing the active buffer's name, all related jump marks are renamed to the new filename. This applies to :bdelete and :bunload, in contrast, :bwipe deletes all related marks including those in jumplists. However, in evil, :bwipeout doesn't clean the marks, so I'm planning to replace (evil-ex-define-cmd "bw[ipeout]" 'evil-delete-buffer) with (evil-ex-define-cmd "bw[ipeout]" 'evil-wipeout-buffer) which takes care of the jumplist.

Currently Evil sets jumpmarks for (no file backened) buffers only if their names match (defvar evil--jumps-buffer-targets "\\*\\(new\\|scratch\\)\\*"). So only this needs special treatment.

> If the answer to the above is no, why not rewrite this to track buffers instead and check with buffer-alive-p?

I think current Evil's behavior is quite similar to Vim, which tracks jumplist by filenames.

wasamasa:
OK, in that case it's fine to keep using the current representation, although you'll need to install advice for `rename-buffer` so that the buffer name change is applied to the jump list (see https://github.com/wasamasa/eyebrowse/commit/d36956c065b04a429748bdb4187e9c914a07d733#diff-915b027c5c497c835a9868e10753ddffR289 for an example).

evil-commands.el
```diff
@@ -2924,6 +2924,26 @@ for the last window in each frame."
...
             wins))))
 
+(evil-define-command evil-wipeout-buffer (buffer &optional bang)
+  "Deletes a buffer. Also clears related jump marks.
+All windows currently showing this buffer will be closed except
+for the last window in each frame."
+  (interactive "<b><!>")
+  (let* ((ring (make-ring evil-jumps-max-length))
+         (struct (evil--jumps-get-current))
+         (idx (evil-jumps-struct-idx struct))
```
wasamasa: What is this variable even used for?

evil-commands.el
```diff
@@ -2924,6 +2924,26 @@ for the last window in each frame."
                   (error nil)))
             wins))))
 
+(evil-define-command evil-wipeout-buffer (buffer &optional bang)
+  "Deletes a buffer. Also clears related jump marks.
+All windows currently showing this buffer will be closed except
+for the last window in each frame."
+  (interactive "<b><!>")
+  (let* ((ring (make-ring evil-jumps-max-length))
```
wasamasa: Don't use `let*` where `let` is sufficient.

evil-commands.el
```diff
@@ -2924,6 +2924,26 @@ for the last window in each frame."
...
+         (struct (evil--jumps-get-current))
```
wasamasa: This is not a very descriptive identifier.

evil-commands.el
```diff
@@ -2924,6 +2924,26 @@ for the last window in each frame."
...
+         (idx (evil-jumps-struct-idx struct))
+         (i 0))
+    (cl-loop for jump in (ring-elements (evil--jumps-get-window-jump-list))
+             do (let* ((file-name (cadr jump)))
+                  (if (or (string= file-name (buffer-file-name))
+                          (string-match-p evil--jumps-buffer-targets (buffer-name)))
+                      (if (<= i idx) (1- idx))
+                    ;; else
+                    (ring-insert ring jump)
```
wasamasa: This is the only side effect here, the rest doesn't end up doing anything.

evil-commands.el
```diff
@@ -2924,6 +2924,26 @@ for the last window in each frame."
...
 
+(evil-define-command evil-wipeout-buffer (buffer &optional bang)
+  "Deletes a buffer. Also clears related jump marks.
+All windows currently showing this buffer will be closed except
+for the last window in each frame."
+  (interactive "<b><!>")
+  (let* ((ring (make-ring evil-jumps-max-length))
+         (struct (evil--jumps-get-current))
+         (idx (evil-jumps-struct-idx struct))
+         (i 0))
```
wasamasa: Same with this one, you only subtract 1 from it and end up doing nothing with the result.

evil-commands.el
amosbird: there is a variable dependency

evil-commands.el
amosbird: it's the updated index of the new jump ring

Issue #992, opened 2017-12-08
[Request] Make evil-integration optional


jojojames:
- Enhancement request

Hi, I'd like to request making evil-integration.el optional.

There seems to be some good integrations in that file so it might be a better idea to make only the extra keybindings optional?

Thanks!

edkolev:
I can submit a PR to make loading of evil-integration.el optional.

@jojojames what do you meen extra keybindings? Are you suggesting we split evil-integration into two parts and make loading of one of them optional?

Here's a patch which introduces a new variable `evil-want-integration` (default `t`):

```diff
diff --git a/evil-vars.el b/evil-vars.el
index 899aab5..8870b47 100644
--- a/evil-vars.el
+++ b/evil-vars.el
@@ -1853,6 +1853,11 @@ Otherwise the previous command is assumed as substitute.")
           "1.2.13"))))
   "The current version of Evil")

+(defcustom evil-want-integration t
+  "Wheter to load evil-integration.el"
+  :type 'boolean
+  :group 'evil)
+
 (defun evil-version ()
   (interactive)
   (message "Evil version %s" evil-version))
diff --git a/evil.el b/evil.el
index 01b9bc1..5c9e54a 100644
--- a/evil.el
+++ b/evil.el
@@ -137,7 +137,9 @@
 (require 'evil-commands)
 (require 'evil-jumps)
 (require 'evil-maps)
-(require 'evil-integration)
+
+(when evil-want-integration
+  (require 'evil-integration))

 (run-hooks 'evil-after-load-hook)
```

jojojames:
There's quite a few useful functionality in evil-integration. The ones that are not so useful (potentially) are the keybinding ones or the ones that set the default modes and definitely one swith overriding maps.

These are quick examples of ones that might be better optional.

```
(evil-add-hjkl-bindings Buffer-menu-mode-map 'motion)

;; dictionary.el

(evil-add-hjkl-bindings dictionary-mode-map 'motion
  "?" 'dictionary-help        ; "h"
  "C-o" 'dictionary-previous) ; "l"

;;; Dired

(eval-after-load 'dired
  '(progn
     ;; use the standard Dired bindings as a base
     (defvar dired-mode-map)
     (evil-make-overriding-map dired-mode-map 'normal)
     (evil-add-hjkl-bindings dired-mode-map 'normal
       "J" 'dired-goto-file                   ; "j"
       "K" 'dired-do-kill-lines               ; "k"
[...]

PR #993, opened 2017-12-08
Optionally load evil-integration.el

edkolev:
This is a WIP PR as a result of #992 

I haven't self-tested thoroughly yet. Plus, there's an issue with the generation of `.depend` which is currently worked-around like so:

```elisp
(when evil-want-integration
;; TODO creating the .depend file breaks if the following require is
;; not on a line of its own
(require 'evil-integration)
)
```

Edit: the workaround is is now removed.

Any suggestions for improvements are welcome.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ compile: $(ELCFILES)
 	@echo Compute dependencies
 	@rm -f .depend
 	@for f in $(FILES); do \
-	    sed -n "s/(require '\(evil-.*\))/$${f}c: \1.elc/p" $$f >> .depend;\
+	    sed -n "s/ *(require '\(evil-[^)]*\).*)/$${f}c: \1.elc/p" $$f >> .depend;\
 	done
 
 -include .depend
--- a/evil-vars.el
+++ b/evil-vars.el
@@ -1853,6 +1853,12 @@ Otherwise the previous command is assumed as substitute.")
           "1.2.13"))))
   "The current version of Evil")
 
+(defcustom evil-want-integration t
+  "Whether to load evil-integration.el.
+This variable must be set before evil is loaded."
+  :type 'boolean
+  :group 'evil)
+
 (defun evil-version ()
   (interactive)
   (message "Evil version %s" evil-version))
--- a/evil.el
+++ b/evil.el
@@ -139,10 +139,7 @@
 (require 'evil-maps)
 
 (when evil-want-integration
-;; TODO creating the .depend file breaks if the following require is
-;; not on a line of its own
-(require 'evil-integration)
-)
+  (require 'evil-integration))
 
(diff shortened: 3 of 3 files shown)


noctuid:
I think that it would be much preferable to split the files into one with keybindings and one with everything else (or to further split everything else, but I think the main thing most people would want to disable is only the keybindings). See my comment in #992. I don't think everyone would agree on what is "basic" (for example, I'd include the advice and setting of evil command properties).

wasamasa:
Sigh. I don't want to split up anything at all.

VanLaser:
For one, yes, I'd like evil not to load these keybindings by default, so I'd welcome a split too. Right now I have to use (are these easy to find for an evil newcomer? I think no):

```elisp
  ;; don't let modes override any states (!)
  (setq evil-overriding-maps nil
        evil-intercept-maps nil
        evil-pending-intercept-maps nil
        evil-pending-overriding-maps nil)
  ;; subvert evil-operation.el overrides (dired, ibuffer etc.)
  (advice-add 'evil-make-overriding-map :override #'ignore)
  (advice-add 'evil-make-intercept-map  :override #'ignore)
  (advice-add 'evil-add-hjkl-bindings   :override #'ignore)
```

... so it would be much more "natural" to not have to retort to these workarounds.

TheBB:
I'm fine making evil-integration optional. Strongly agree with @wasamasa, let's not split it.

noctuid:
Providing a setting to not load `evil-integration` without splitting it doesn't really solve anything. Users still have to require it for the configuration they want and use workarounds to prevent/undo the keybindings/overriding maps parts. That or they have to not load it all (now with an official setting) and copy the other parts into their own config or a new package. Nothing really changes.

Why don't you want to split it?

wasamasa:
See the linked ticket.

edkolev:
OK, this PR is now modified - it only adds the option to load evil-integration.el optionally. The generation of the .depend file is also tweaked.

Makefile
```diff
@@ -20,7 +20,7 @@ compile: $(ELCFILES)
 	@echo Compute dependencies
 	@rm -f .depend
 	@for f in $(FILES); do \
-	    sed -n "s/(require '\(evil-.*\))/$${f}c: \1.elc/p" $$f >> .depend;\
+	    sed -n "s/[^(]*(require '\(evil-[^)]*\).*)/$${f}c: \1.elc/p" $$f >> .depend;\
```
wasamasa: This seems somewhat broad, can't you just allow leading space in front of `(require 'evil-...)`?

edkolev:
Sure, the commit is now amended - only spaces are allowed in front on the `(require ...`

wasamasa:
Thanks!

PR #994, opened 2017-12-11
Add evil-define-key-with-key-func

justbur:
This is just an idea I had for simplifying key bindings...

This is an anaphoric version of evil-define-key which allows one to transform
all of the key descriptions according to a "key-func". A useful example is

(evil-define-key-with-key-func (kbd key) STATE KEYMAP BINDINGS)

which expands to a standard evil-define-key call after applying kbd to every key
in BINDINGS. More info is in the docstring.

### DIFF
--- a/evil-core.el
+++ b/evil-core.el
@@ -1075,6 +1075,37 @@ its behavior more predictable."
     (dolist (map maps)
       (evil-set-keymap-prompt map (keymap-prompt map)))))
 
+(defmacro evil-define-key-with-key-func (key-func state keymap &rest bindings)
+  "Call `evil-define-key' using KEY-FUNC on each key in BINDINGS.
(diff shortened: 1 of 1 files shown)


wasamasa:
Sorry, but I don't really see the benefit for this.  I can get if someone prefers to get rid of an `evil-after-load` or multiple `define-key` invocations by using `evil-define-key`, but using a specialized version for not having to type `(kbd ...)`?

justbur:
It's not just for `kbd`. It can also be used for defining keys with common prefixes as in 

```elisp
(evil-define-key-with-key-func
 (kbd (concat "C-c a " key))
 "a" 'foo
 "b" 'bar
 ...)
```

That being said, I'm not sure how useful it either. It's more of a convenience option for those who like it.

PR #995, opened 2017-12-13
Teach imenu about evil macros

justbur:


wasamasa:
This is neat, but I'd rather go the way Smartparens did and put such stuff into Evil's `.dir-locals.el` as it's only helpful for us, not Evil users.

justbur:
Fair point, although it doesn't harm them and evil is intended to be extensible...

justbur:
How about this as a compromise? I started a new file called `evil-development.el` that will not be required with `evil.el`. That way evil developers can optionally require the file in their config. I don't think I want it to be a `.dir-locals.el` thing, because then it's not that helpful for working on external evil pacakges.

wasamasa:
Yeah, that works for me.  With this approach, will the file also be included in the built package or will developers not using the Git checkout have to copy the files over manually?

edkolev:
Melpa will include it. The current recipe excludes only evil-test-helpers.el https://github.com/melpa/melpa/blob/master/recipes/evil

wasamasa:
Cool, then I see no issues with merging.

edkolev:
After giving some thought to this, I think it might be a good idea allow evil users rather than just evil developers to take advantage of the imenu integration. I would say it's likely that many evil users actually have some form of `evil-define-*` in their in init.el.

Maybe evil could have an option to enable imenu integration. I've seen this approach in use-package.

wasamasa:
No need, anyone can put `(require 'evil-development)` in their init file, no matter whether they're hacking on Evil-related packages or not.

PR #997, opened 2017-12-14
Suggest repro steps should include `make emacs`

naclander:


wasamasa:
Thanks, but the "see above" isn't really helpful.  I'd rather use the above link again.

naclander:
@wasamasa sounds good, I've updated the patch.

PR #1002, opened 2017-12-20
Fix loading of advice for find-tag-noselect

justbur:
Defer advice until after etags has loaded to prevent a find-tag-noselect was
redefined message.


edkolev:
Thanks! This has been bugging me for a long time.

wasamasa:
Thanks!

Issue #945, opened 2017-10-17
Segfaults on CI


rski:
Moving it from here:
https://github.com/emacs-evil/evil/pull/758#issuecomment-336720241

The issue is more or less that segfaults are showing up intermittently on the CI with emacs25.

Locally I have an Emacs 26@2e1b3522b8562f94a98fed07aeddb6b5ba0dbc6c built with and without debugging enabled.

I have seen failures intermittently on the non debug build.  I ran the tests with:
```
$ for i in `seq 1 10`; do; make test; done
````
I thought I narrowed down the issue to  `evil-test-copy`, but I'm not sure anymore.                                                 

On the debug build I always get failures:
```
~/Code/emacs/src/emacs -nw -Q -L . -L lib -l evil-tests.el \
--eval "(evil-tests-initialize '() '())"
Fatal error 6: Aborted
lisp.h:1355: Emacs fatal error: assertion failed: 0 <= nchars
Aborted (core dumped)
```
I think it is the same test every time, I will add some logging to figure out which one it is and report back

PR #1003, opened 2017-12-21
Try evm emacs for travis again

justbur:
I recently updated some of my travis configs with the latest instructions for using evm. They seem to work well for me, so I was thinking doing something similar on evil might help with the segfaults. 

Note that with Emacs 26 and master I'm getting ssl errors with cask. Like these https://github.com/cask/cask/issues/373. But that seems like a separate issue with travis and cask. 

Ref #945 

cc: @edkolev

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,21 +2,19 @@ language: generic
 sudo: false
 
 env:
-  global:
-    - CURL="curl -fsSkL --retry 9 --retry-delay 9"
-  matrix:
-  - EMACS_VERSION=24.3
-  - EMACS_VERSION=24.4
-  - EMACS_VERSION=24.5
-  - EMACS_VERSION=25.1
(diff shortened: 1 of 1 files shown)


wasamasa:
I dunno, we've been there before. I've tried reproducing the segfaults in a VM, but failed completely, both with prebuilt and self-built binaries.  The next logical option to try is rewriting the scripts so that binaries are rebuilt every time.  While this is somewhat overkill and would make tests take about 20 minutes, it would eliminate all crashes related to environment mismatches (such as the built Emacs relying on different glibc internals than present on the current system).

justbur:
Well... I haven’t seen segfaults yet with this setup. And I see them all the time with the current one. 

I could keep this branch up to date and see if I get any. Travis is building prs

justbur:
Never mind. It doesn't seem to have helped much

PR #1005, opened 2017-12-24
Add missing avy commands in evil-integration

wbolster:
Evil's built-in integration with avy is missing some avy commands,
which causes unexpected behaviour: some avy commands do not properly
behave as evil motions.

Add the missing commands, and while at it, sort the names in
alphabetic order.


wasamasa:
Thanks!

PR #1007, opened 2017-12-25
Switch to VM-based CI

wasamasa:
I've experimented with two CI branches today, in the first one I built the Emacsen from source, the second one happened after I realized all of those segfault started happening after I opted out of VMs when implementing flyspell-related tests. So, I switched back to VMs and this seems to have made a great difference, no segfault so far.  I also took the liberty of reducing the number of builds, the rationale here being that most of them included only minor changes, so the latest one with many changes was chosen.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: generic
-sudo: false
+sudo: true
 
 env:
   global:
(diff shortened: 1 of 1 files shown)


wasamasa:
I briefly considered that, but CircleCI also uses containers whereas Travis lets you choose, so that would probably not help.

PR #1008, opened 2017-12-27
If `evil-ex` is called with initial input, call `evil-ex-update`

edkolev:
This PR tries to fix the following issue - calling `evil-ex` non-interactively with initial input `%s/.../...` doesn't show the substitution overlays.

To reproduce the issue I'm trying to fix:
- enter the following in the scratch buffer:
```elisp
;; hello this world
(evil-ex "%s/this/that/")
```
- execute the second line

Without this PR, the substitute overlays aren't displayed -- you don't see what will be replaced.

### DIFF
--- a/evil-ex.el
+++ b/evil-ex.el
@@ -181,7 +181,7 @@ is appended to the line."
         evil-ex-info-string
         result)
     (minibuffer-with-setup-hook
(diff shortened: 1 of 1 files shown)


wasamasa:
Thanks!

PR #1009, opened 2017-12-29
Reveal hidden search terms in evil-ex-substitute

### LINKED ISSUE #983: Have evil-ex-substitute open folds automatically
I couldn't find an issue for this already, but it would be great if `evil-ex-substitute` automatically opened folds after moving the point when used with confirmation (otherwise you can't see what is being replaced).

justbur:
Do this by simulating the effect of turning on reveal-mode when the substitution
requires confirmation.

Fixes #983

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -3346,6 +3346,29 @@ resp.  after executing the command."
                      evil-symbol-word-search))
   (evil-ex-start-word-search t 'backward count symbol))
 
+(defun evil-revert-reveal (open-spots)
+  "Unconditionally close overlays in OPEN-SPOTS in current window.
+Modified version of `reveal-close-old-overlays' from
+reveal.el. OPEN-SPOTS is a local version of `reveal-open-spots'."
+  (dolist (spot open-spots)
+    (let ((window (car spot))
+          (ol (cdr spot)))
+      (unless (eq window (selected-window))
+        (error "evil-revert-reveal: slot with wrong window"))
+      (let* ((inv (overlay-get ol 'reveal-invisible))
+             (open (or (overlay-get ol 'reveal-toggle-invisible)
+                       (get inv 'reveal-toggle-invisible)
+                       (overlay-get ol 'isearch-open-invisible-temporary))))
+        (if (and (overlay-start ol) ;Check it's still live.
+                 open)
+            (condition-case err
+                (funcall open ol t)
+              (error (message "!!Reveal-hide (funcall %s %s t): %s !!"
+                              open ol err)))
+          (overlay-put ol 'invisible inv))
+        ;; Remove the overlay from the list of open spots.
+        (overlay-put ol 'reveal-invisible nil)))))
+
 (evil-define-operator evil-ex-substitute
   (beg end pattern replacement flags)
   "The Ex substitute command.
@@ -3373,6 +3396,8 @@ resp.  after executing the command."
          (evil-ex-substitute-hl (evil-ex-make-hl 'evil-ex-substitute))
          (orig-point-marker (move-marker (make-marker) (point)))
          (end-marker (move-marker (make-marker) end))
+         (use-reveal (and confirm (require 'reveal nil t)))
+         reveal-open-spots
          zero-length-match
          match-contains-newline
          transient-mark-mode)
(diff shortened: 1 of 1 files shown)


wasamasa:
This is pretty nice! One thing I've noticed is that this leaves the last overlay open.  Is that expected behavior or should it be changed accordingly?

justbur:
I’m not sure. What’s your feeling?

wasamasa:
I find both options mildly annoying, like if I planned to edit the now folded text, I need to unfold it first, likewise if I'm not interested in the now unfolded text, I need to fold it again.  Judging from a quick test, Vim unfolds *everything* (which I doubt is necessary to emulate), then restores the previous folding state.  So, assuming you've replaced something in an already open fold, it wouldn't be closed after the replacement is done.

wolfv6:
@justbur, Thanks for doing all the work.
My vote is to restore the previous folded state like Vim, because unlikely to edit around the last replace.

justbur:
Where should the cursor end up if the fold is closed? Where does vim leave it?

wolfv6:
@justbur,
After "replace? y", the cursor is at beginning of line that contains the last replacement.
After "replace? n", the cursor is at beginning of the last word found.
I think either way is good.

justbur:
@wasamasa and @wolfv6 I think the latest commit is closer to what you’re after.

wolfv6:
This will be my first time testing an emacs pull request.
How do I get a copy of your new evil-commands.el file?
The one on https://github.com/justbur/evil is 5 months old.
Then I can replace my ~/evil/evil-commands.el  with your new one.
Is that the usual way to test evil PR files?

TheBB:
This PR is from @justbur's `fold-sub` branch, not his master branch.

wolfv6:
I tested your new evil-commands.el file.
The substitute command replaced, but the org heading did not unfold.
Maybe I didn't install the file correctly because I am new at this.

Here is what I did:
I backup my ~/evil/evil-commands.el file.
Save your new file to the same location, from https://github.com/justbur/evil/blob/fold-sub/evil-commands.el > Raw
Start Spacemacs and open an .org file.
Fold the headings.
Enter a substitute command:

    :%s/ASCII/ASCIII/gc

Pressing 'y' replaced as expected, but did not unfold the org heading.
Which is the same behavior as before.

justbur:
@wolfv6 org folds behave as expected for me. You're probably (unintentionally) loading another version of evil-commands.el. After loading Emacs, try opening the correct version of evil-commands.el and doing `M-x eval-buffer`. Then do your test.

wolfv6:
@justbur 
Thanks for the help.  eval-buffer worked as advertized.

The substitute opens and restore heading state nicely.
But I noticed that y w (yank word) does not yank.

The following is not in the original request, mission creep if you will.
u (undo) works, but the heading remains closed.
Vim opens the heading for undo, similar to substitute behavior.
This is less important, unless it is easy to fix, leave it as is.

justbur:
> But I noticed that y w (yank word) does not yank.

In what context? This seems like a separate issue.

> The following is not in the original request, mission creep if you will.
u (undo) works, but the heading remains closed.
Vim opens the heading for undo and restores heading state, similar to substitute behavior.
This is less important, unless it is easy to fix, leave it as is.

undo is tied to the undo-tree package and there doesn't seem to be a simple way to fix this. However, the following might work as a partial solution. 

```elisp
(defun my-reveal-point (oldfunc &rest args)
  (let (reveal-open-spots)
    (apply oldfunc args)
    (reveal-post-command)))

(advice-add #'undo-tree-undo :around #'my-reveal-point)
```

All this does is open up the fold if point lands inside a fold after undo.

wolfv6:
@justbur 

> In what context? This seems like a separate issue.

y w (yank word), then p (put), and text from the previous clipboard is inserted.
This only happens when I am running the new file from https://github.com/justbur/evil/blob/fold-sub/evil-commands.el
I have tried it on two different .org files.

justbur:
@wolfv6 please open a new issue with reproduction steps for that one.

wolfv6:
@justbur I opened an issue at https://github.com/emacs-evil/evil/issues/1012

wasamasa:
@justbur Thanks, this works as envisioned. One more thing though, the code introduces a new warning due to `reveal-post-command` being possibly undefined.  I think we can just put `(require 'reveal)` above the function and simplify the binding to `(use-reveal confirm)`.

wolfv6:
@justbur, Sorry I can 't help with the testing. 
I don't know how to use evil Emacs.  I only know how to use evil Spacemacs org-mode.

justbur:
@wasamasa fine with me.

@wolfv6 that’s ok but we can’t do anything without being able to reproduce it.

justbur:
@wasamasa I couldn't get the compiler warning to go away without putting the require at the top level. reveal is an old library and it's small, so I don't think it makes much difference.

PR #1010, opened 2017-12-29
Don't move point following ex-substitute search

justbur:
The current behavior isn't consistent with vim (at least in my tests). It's also
annoying to have point jump to weird locations following an aborted search.

For example, with
`:%s/a/b/c`

If you abort the search with `C-g` before making replacements, point jumps to the
beginning of the region which is the beginning of the buffer in this case. This
is odd, because nothing happened but you lose your place in the buffer.

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -3402,6 +3402,7 @@ resp.  after executing the command."
                     ;; beginning of this last line will be matched which we don't
                     ;; want, so we abort here.
                     (throw 'exit-search t))
+                  (setq evil-ex-substitute-last-point match-beg)
                   (if confirm
                       (let ((prompt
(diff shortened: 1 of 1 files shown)


justbur:
I might be missing some rationale for leaving this in. It was a hold over from the old code, but I've never liked this behavior.

wasamasa:
I don't see any particular reason for sticking to this behavior.  What worries me though is that it breaks five tests.  Have you investigated into those yet?

justbur:
No not yet. I saw those and will look into them.

________________________________

justbur:
@wasamasa I think I fixed the issues. I didn't adjust the tests, just the logic a bit. This most noticeably affects where point goes when you abort a search (should be nowhere).

wasamasa:
Thanks!

PR #1011, opened 2017-12-31
Use xref-find-definitions if possible

wasamasa:
This makes Evil compile without a single warning.  The `(info "(elisp) Compiler Errors")` info node contained the missing bits to make this work.  Note that you cannot function-quote things in an `(fboundp 'foo)` construct, otherwise the warning isn't supressed.  I suspect that's a bug though, given how little function-quoting is used in existing code and might figure out a patch to make the byte-compiler smarter about it.

I also noticed that `g C-]` is still using `find-tag`, so I fixed that, even if it's not 100% correct.


edkolev:
Nice, this is good progress.

I think the implementation with xref can be simplified a bit since xref-find-definitions already behaves very similarly to evil-jump-to-tag:

```
...
(cond
   ((fboundp 'xref-find-definitions)
    (let ((xref-prompt-for-identifier arg))
      (call-interactively #'xref-find-definitions)))

   (fboundp 'find-tag)
    ...)
```

I'm not sure if the above implementation won't introduce another compiler warning though.

wasamasa:
Good call, thanks!

edkolev:
I think this line can be safely removed as well
 
```
(declare-function xref-backend-identifier-at-point "xref.el")
```

wasamasa:
Right, with that simplification the "function" is no longer used.

PR #1013, opened 2018-01-06
Allow auxiliary maps to be intercept maps

noctuid:
An example use case of making an evil keymap an intercept keymap is to allow a user to reliably give keybindings precedence over any evil keybindings made with `evil-define-key` or `evil-define-minor-mode-key` (e.g. the user uses a non-conventional prefix key that a lot of packages bind by default). Currently, doing this with intercept maps is fairly convoluted. With this commit, it would possible to just do something like this:
```
(define-minor-mode my-intercept-mode ...)
....
(evil-define-key 'normal my-intercept-mode-map
  [intercept-state] 'normal
  "<high precedence key>" #'binding)
```

### DIFF
--- a/evil-core.el
+++ b/evil-core.el
@@ -476,11 +476,15 @@ higher precedence. See also `evil-make-intercept-map'."
       (define-key copy key (or state 'all))
       (define-key keymap key copy))))
 
-(defun evil-make-intercept-map (keymap &optional state)
+(defun evil-make-intercept-map (keymap &optional state aux)
   "Give KEYMAP precedence over all Evil keymaps in STATE.
-If STATE is nil, give it precedence over all states.
-See also `evil-make-overriding-map'."
-  (let ((key [intercept-state]))
+If STATE is nil, give it precedence over all states. If AUX is non-nil, make the
+auxiliary keymap corresponding to KEYMAP in STATE an intercept keymap instead of
+KEYMAP itself. See also `evil-make-overriding-map'."
+  (let ((key [intercept-state])
+        (keymap (if aux
+                    (evil-get-auxiliary-keymap keymap state t t)
+                  keymap)))
     (define-key keymap key (or state 'all))))
 
 (defmacro evil-define-keymap (keymap doc &rest body)
@@ -810,7 +814,9 @@ See also `evil-mode-for-keymap'."
   (let* ((state (or state evil-state))
          result)
     (dolist (map (current-active-maps))
-      (when (setq map (evil-intercept-keymap-p map state))
+      (when (setq map (or (evil-intercept-keymap-p map state)
+                          (evil-intercept-keymap-p
+                           (evil-get-auxiliary-keymap map state) state)))
         (push (cons (evil-mode-for-keymap map t) map) result)))
     (setq result (nreverse result))
     result))


justbur:
I’m not opposed to this commit but it seems like the interface could be improved further. What if you could specify something like `’intercept-normal` for the state which would translate into what you have?

noctuid:
I'm not sure that I understand what you mean. Do you mean that specifying `intercept-normal` as the state for `evil-define-key` would automatically add the `[intercept-state] 'normal` keybinding? That doesn't seem like a good idea to me because it could cause confusion. It's a permanent change (undoable but will affect later keybindings with just `normal`) to the keymap as opposed to only affecting the `evil-define-key` it is used with, which is what I'd expect if there were both `normal` and `intercept-normal`. Sorry if I've misunderstood. If it's just an issue of making it easier to create the intercept map, it might be better to add another optional argument (e.g. `AUX`) to `evil-make-intercept-map`, so you could just specify `(evil-make-intercept-map keymap 'normal t)`. I think that would be more clear and less complicated.

justbur:
@noctuid You make a good point about it being a change that affects future keybindings, but isn't the same true of your example?

What I don't like about the example is that you are using your knowledge of how intercept maps are implemented, where a user likely won't know this. 

I think it's a useful enough feature to make a simple to understand way for a user to make such bindings without knowing the implementation details. Does that make sense?

noctuid:
> You make a good point about it being a change that affects future keybindings, but isn't the same true of your example?

Yes, but it's more clear that this is the case (particularly because it requires knowledge about intercept keymaps). That said, I wasn't suggesting that users should be expected to do it like in the original example, only that it would probably be preferable to directly extend `evil-make-intercept-map` instead of changing `evil-define-key`. I've updated `evil-make-intercept-map` to allow making an auxiliary keymap an intercept map.

wasamasa:
I don't have any particularly strong opinion on this and would rather let you two talk it out because I don't even know what an intercept map is supposed to be.

justbur:
@noctuid Can you give an example of the problem you are trying to fix? I'm still a little confused by this. Right now, the ranking of evil keymaps is 

1. intercept
2. local state
3. minor-mode
4. auxiliary
5. overriding
6. global state

This change would all you to promote an auxiliary map to intercept, but why does that get around a bad interaction with other packages? Wouldn't the other packages using that auxiliary map also have their bindings promoted in this case? Or is the issue with the auxiliary bindings falling below the minor-mode ones?

noctuid:
> Wouldn't the other packages using that auxiliary map also have their bindings promoted in this case?

No other packages would be using the auxiliary map. It would be defined by the user or by a package explicitly for this purpose (so that no package's local/minor-mode/auxiliary keybindings can override the user's keybindings).

justbur:
> No other packages would be using the auxiliary map. It would be defined by the user or by a package explicitly for this purpose (so that no package's local/minor-mode/auxiliary keybindings can override the user's keybindings).

I understand the purpose you are going for now. You just want to replace something like this with what you have

```elisp
(define-minor-mode my-intercept-mode "" ...)
(evil-make-intercept-map my-intercept-mode-map 'normal)
(define-key my-intercept-mode-map "a" 'func)
```
I'm not crazy about having auxiliary maps behave as intercept maps, because it seems like it could be confusing. But as your example shows you still have to know what you're doing to achieve this, so the risk seems limited.

I guess what I was expecting (and maybe hoping for) is something like 

```elisp
(evil-define-intercept-key 'normal lisp-mode-shared-map
  "a" 'func)
```
where this means that this binding functions as an intercept binding when `lisp-mode-shared-map` is active. I was also thinking of it as something that would ideally be independent of any auxiliary bindings in `lisp-mode-shared-map` (so that the above binding has no influence on bindings made with `evil-define-key`). This would require a little more work under the covers, but might be a useful feature.

noctuid:
> I'm not crazy about having auxiliary maps behave as intercept maps, because it seems like it could be confusing.

I'm not sure why it would be confusing. `evil-define-key` is the most common way of creating evil keybindings for modes. If I provide a single minor mode/keymap that's advertised as being higher precedence for evil keybindings in a package, using auxiliary maps/`evil-define-key` is the expected method for binding keys. Users don't even need to know any details about intercept maps.

As for this
```
(define-minor-mode my-intercept-mode "" ...)
(evil-make-intercept-map my-intercept-mode-map 'normal)
(define-key my-intercept-mode-map "a" 'func)
```
Even ignoring the intercept map part, `a` is now `func` in insert state. While you could alternatively intentionally create intercept versions of every global evil keymap, I think just using auxiliary maps for a single minor mode's keymap is simpler.

> where this means that this binding functions as an intercept binding when lisp-mode-shared-map is active. I was also thinking of it as something that would ideally be independent of any auxiliary bindings in lisp-mode-shared-map (so that the above binding has no influence on bindings made with evil-define-key). This would require a little more work under the covers, but might be a useful feature.

I can't think of a use case, but that does potentially seem useful. I think it should be considered separately from this pull request though.

Do you have any other suggestions for this pull request? Do you think it would be okay to merge as it is?

justbur:
> Even ignoring the intercept map part, a is now func in insert state. 

Oh right. You've obviously thought through this more than I have.

> Do you have any other suggestions for this pull request? Do you think it would be okay to merge as it is?

I understand it better now. Yes, it's fine with me to merge this. 

cc @wasamasa

wasamasa:
LGTM as well.

Ambrevar:
Is anything blocking this merge?

wasamasa:
Nope.

PR #1014, opened 2018-01-09
Don't use make-temp-name

justbur:
The docstring says there is a race condition if you use `make-temp-name` then
later try to write the file which is exactly what was happening in
`evil-test-write`. I suspect this is the cause of the (random) segfaults in
travis.

Also disentangle the tests in `evil-test-write`.

@wasamasa I looked through old builds and most seemed to be failing around this test, so I think this is the issue with the seg faults. Not to mention that trying to illegally access a file (for whatever reason) along with a race condition fits with random seg faults.

### DIFF
--- a/evil-test-helpers.el
+++ b/evil-test-helpers.el
@@ -385,8 +385,7 @@ is executed at the end."
 
 (defun evil-temp-filename ()
   "Return an appropriate temporary filename."
-  (make-temp-name (expand-file-name "evil-test"
-                                    temporary-file-directory)))
+  (make-temp-file "evil-test"))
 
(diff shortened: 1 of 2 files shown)


justbur:
Including the second commit would be my preference, but I understand that this might be more controversial.

wasamasa:
I don't buy the conclusion from a race condition to a segfault (the traces I've looked at suggested allocator issues interleaved with glyph matrices), but this kind of code definitely doesn't need to be.  Heck, [I've ranted about it before](http://emacshorrors.com/posts/make-temp-name.html).

In case this actually resolves the segfaults I might roll back the dockerized testing.  It will be faster and who knows, Travis might get rid of VMs alltogether...

Issue #532, opened 2015-06-21
Evil uses the clipboard instead of the primary selection for mouse selection


TheBB:
Originally reported by: **hpdeifel (Bitbucket: [hpdeifel](https://bitbucket.org/hpdeifel), GitHub: [hpdeifel](https://github.com/hpdeifel))**

----------------------------------------

I'll copy and paste my email (Subject: visual-mode and X-selections)
from 2012:

    I think that the interaction between visual-mode and the X11 selections
    is not optimal. Since I don't know if the current behavior is
    intentional, I haven't yet opened a bug report.
    
    The selected text in visual-mode is currently copied via the function
    x-select-text. With the Emacs 24.1 default configuration, this copies to
    the X-selection CLIPBOARD (configurable through
    x-select-enable-clipboard and x-select-enable-primary.)
    
    In the rest of Emacs, this function is only used for the explicit
    copy/yank commands. Copying the active region is controlled by the
    variable select-active-regions and uses the PRIMARY X-selection by
    default.
    
    This behavior is consistent with most other X applications: The mouse
    selection is saved in PRIMARY and the explicit copy/paste commands use
    the CLIPBOARD.
    
    Thus, it would be nice if Evil would also use the PRIMARY selection by
    default for the selected text in visual-mode.
    
    I currently use the following redefinition of
    evil-visual-update-x-selection, that does what I need:
    
    #+BEGIN_SRC elisp
    (defun evil-visual-update-x-selection (&optional buffer)
      "Update the X selection with the current visual region."
      (with-current-buffer (or buffer (current-buffer))
        (when (and (evil-visual-state-p)
                   (fboundp 'x-select-text)
                   (or (not (boundp 'ns-initialized))
                       (with-no-warnings ns-initialized))
                   (not (eq evil-visual-selection 'block)))
          (x-set-selection 'PRIMARY (buffer-substring-no-properties
                                     evil-visual-beginning
                                     evil-visual-end))
          (setq x-last-selected-text-primary ))))
    #+END_SRC

@lyro suggested to use the following code instead:

```
(defun evil-visual-update-x-selection (&optional buffer)
  "Update the X selection with the current visual region."
  (with-current-buffer (or buffer (current-buffer))
    (when (and (evil-visual-state-p)
               (fboundp 'x-set-selection)   ; <-- small change, hope it works ;)
               (or (not (boundp 'ns-initialized))
                   (with-no-warnings ns-initialized))
               (not (eq evil-visual-selection 'block)))
[...]

PR #1015, opened 2018-01-11
Copy text to PRIMARY, not CLIPBOARD in visual mode

hpdeifel:
Previously, the selected text in visual mode was copied to the X
CLIPBOARD, while 'normal' Emacs copies the active region to the
PRIMARY selection (controlled by `select-active-regions`). The latter
is also what the major X toolkits do by default.

This changes evil-visual-update-x-selection to use the PRIMARY
selection and also...

- Uses the wrapper `evil-set-selection` instead of the obsolete
  `x-select-text`
- Tests for `display-selections-p` instead of boundness of
  `x-select-text` or `ns-initialized`

### DIFF
--- a/evil-states.el
+++ b/evil-states.el
@@ -350,17 +350,13 @@ otherwise exit Visual state."
 
 (defun evil-visual-update-x-selection (&optional buffer)
   "Update the X selection with the current visual region."
-  (let ((buf (or buffer (current-buffer))))
-    (when (buffer-live-p buf)
-      (with-current-buffer buf
-        (when (and (evil-visual-state-p)
-                   (fboundp 'x-select-text)
-                   (or (not (boundp 'ns-initialized))
-                       (with-no-warnings ns-initialized))
-                   (not (eq evil-visual-selection 'block)))
-          (x-select-text (buffer-substring-no-properties
-                          evil-visual-beginning
-                          evil-visual-end)))))))
+  (with-current-buffer (or buffer (current-buffer))
+    (when (and (evil-visual-state-p)
+               (display-selections-p)
+               (not (eq evil-visual-selection 'block)))
+      (evil-set-selection 'PRIMARY (buffer-substring-no-properties
+                                    evil-visual-beginning
+                                    evil-visual-end)))))
 
 (defun evil-visual-activate-hook (&optional command)
   "Enable Visual state if the region is activated."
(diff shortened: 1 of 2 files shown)


hpdeifel:
Tested with emacs-25 compiled with or without X and emacs-24 without X.

This would be a fix for bug #532

hpdeifel:
Two things that I'm not quite sure about:

1. Should we respect `select-active-region`? I think every visual selection can be considered as *active* region in emacs lingo and I'm not sure if evil has a concept of transient-mark (as in: A mouse selection)
2. Is it even correct to call `evil-visual-update-x-selection` from an idle-timer? What if the user selects a frame on another display (e.g another X server) before the timer expires?

olejorgenb:
1. My vote would be to treat every selection the same

evil-states.el
```diff
@@ -350,17 +350,13 @@ otherwise exit Visual state."
 
 (defun evil-visual-update-x-selection (&optional buffer)
   "Update the X selection with the current visual region."
-  (let ((buf (or buffer (current-buffer))))
-    (when (buffer-live-p buf)
```
wasamasa: Why did you remove the `buffer-live-p` check?

wasamasa:
1. Evil intercepts setting the mark (be it with `C-SPC` or the mouse) and enables visual char state, so it doesn't even have the concept of a transient mark.  There's currently a bit of code dealing with `select-active-regions` in Evil for it to not interfere with other commands, so I wouldn't honor it.  In case we're violating Vim compatibility or want to support a Vim option where it would be useful to honor, we can still ponder this issue.
2. I'd hope so! If the user manages such a feat in less than 0.1s and runs into an error, they'll hopefully report a bug.  In other words, I don't care too much about unhatched eggs, proven bugs however...

evil-states.el
hpdeifel: Because I based the code on the version from 2012 that didn't have this and completely ignored this part of the diff. Thanks for noticing, I fixed it in a new commit (and added a test).

wasamasa:
Tests keep running and the functionality works as expected.  The only annoyance left is that flycheck (curiously `make` doesn't) complains about `evil-set-selection` being possibly undefined, but I'll investigate into this.

hpdeifel:
Thanks!

PR #1016, opened 2018-01-17
Add notmuch-tree-mode to evil-emacs-state-modes

leezu:
I assume notmuch-tree-mode was introduced some while after https://github.com/emacs-evil/evil/commit/bb2999c12d67d88e254b43c03827c8e8252a391c (i.e. 7 years ago), which was when the other notmuch modes were added to evil-emacs-state-modes.


wasamasa:
Thanks!

PR #1018, opened 2018-02-09
Add C-u command binding to evil-scroll-up

mwuttke97:
This pull requests adds the mapping from c-u to evil-scroll-up.

Somehow, this mapping was missing, although the mapping for c-d was present.


wasamasa:
Sorry, but no.  There's already `evil-want-C-u-scroll` for those who don't need the universal argument (which is what `C-u` is bound to in Emacs).

mwuttke97:
Ok, fine.

PR #1019, opened 2018-02-10
evil-ex-global: parse COMMAND only once, not on each matching line

edkolev:
This PR tries to speed up the `:global` ex command. The speedup is roughly 3 times, although I haven't done extensive performance testing.

I've tested `:g/fake/d` on a file with 1k and a file with 10k lines, each line with text `fake`. The results, collected with `elp-instrument-function` on a laptop, are (in seconds):

```
before optimization:
    1k lines:  evil-ex-global       3.469728
    10k lines: evil-ex-global       34.234797

after optimization:
  1k lines:  evil-ex-global     1.6581169999
  10k lines: evil-ex-global     11.175745
```

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -3599,6 +3599,9 @@ This is the same as :%s//~/&"
   (evil-with-single-undo
     (let ((case-fold-search
            (eq (evil-ex-regex-case pattern 'smart) 'insensitive))
+          (command-form (evil-ex-parse command))
+          (transient-mark-mode transient-mark-mode)
+          (deactivate-mark deactivate-mark)
           match markers)
       (when (and pattern command)
         (setq isearch-string pattern)
@@ -3618,7 +3621,7 @@ This is the same as :%s//~/&"
         (unwind-protect
             (dolist (marker markers)
               (goto-char marker)
-              (evil-ex-eval command))
+              (eval command-form))
           ;; ensure that all markers are deleted afterwards,
           ;; even in the event of failure
           (dolist (marker markers)
--- a/evil-ex.el
+++ b/evil-ex.el
@@ -826,14 +826,8 @@ Returns the line number of the match."
 NUMBER defaults to 1."
   (funcall sign (or number 1)))
 
-(defun evil-ex-eval (string &optional start)
-  "Evaluate STRING as an Ex command.
-START is the start symbol, which defaults to `expression'."
-  ;; disable the mark before executing, otherwise the visual region
-  ;; may be used as operator range instead of the ex-range
-  (let ((form (evil-ex-parse string nil start))
-        transient-mark-mode deactivate-mark)
-    (eval form)))
+;; function `evil-ex-eval' has been superseded by `evil-ex-parse' plus `eval'
+(make-obsolete 'evil-ex-eval 'evil-ex-parse "1.2.14")
 
 (defun evil-ex-parse (string &optional syntax start)
   "Parse STRING as an Ex expression and return an evaluation tree.


wasamasa:
Thanks! The speed-up isn't terribly surprising because parsing is known to be relatively slow. There's a few things I wonder about:

- This is the only place `evil-ex-eval` is used at, no? We might remove it, although it's unclear whether anything else depends on it, so perhaps just deprecate it if it's otherwise useless
- The simplification to just `(eval command-form)` isn't quite correct because the original let-binds `transient-mark-mode` and `deactivate-mark`, so I'd do that in the proposed change as well, ideally around the loop

edkolev:
Hey, thanks for the feedback.

> This is the only place evil-ex-eval is used at, no? We might remove it, although it's unclear whether anything else depends on it, so perhaps just deprecate it if it's otherwise useless

Sure. Should I declare it superseded by `evil-ex-parse` with `(make-obsolete 'evil-ex-eval 'evil-ex-parse)`? I'm not sure I can declare it as superseded by `evil-ex-parse` plus `eval`.

> The simplification to just (eval command-form) isn't quite correct because the original let-binds transient-mark-mode and deactivate-mark, so I'd do that in the proposed change as well, ideally around the loop

I left these vars out of the PR because

1. an enormous amount of spam in the *Messages* buffer was created - these 2 lines were added to the buffer per each matched line:

```
Transient-Mark mode enabled
Transient-Mark mode disabled
```

2. when running plain ex commands, outside of `:global` (for example `:pwd`), these 2 variables are not modified https://github.com/emacs-evil/evil/blob/evil-ex-global-parse-command-once/evil-ex.el#L208

wasamasa:
Yes, I'd declare `evil-ex-eval` as obsoleted by `evil-ex-parse` (not sure about the `when` argument though, use the next stable version number?) and add the note about `eval` in a comment.

I suspect the purpose of let-binding those variables is so that if there is an active region it's hidden for the duration of the command and not deactivated by changes to the buffer (so that it's restored properly afterwards). Can't you just let-bind them around the loop so that the message appears just once?

edkolev:
Done.

However, there's still spam generated in the Messages buffer when I use `:g/fake/d` or `:g/fake/yank A`. But no spam for `:g/fake/pwd`. Probably `evil-motion-range` or some other function related to handling the range of commands is triggering the spam.

I believe this happens:
- `evil-ex-global` deactivates transient-mark-mode (around the loop for each matching line)
- the `d` (delete) command, being an evil-operator, gets its range handled by `evil-motion-range`, which in turn uses `evil-with-transient-mark-mode` macro, which enables the transient-mark-mode momentarily, then restores it.

Maybe `evil-ex-global` should remember the value of transient-mark-mode, run the loop per mathinc line, then restore the value of transient-mark-mode if changed.

wasamasa:
Ugh, that is indeed a problem. However, binding `transient-mark-mode` in a let will preserve the global binding being changed from the inner code (instead the local binding will be affected).

edkolev:
How about doing something like this around the loop:

```elisp
;; make sure transient-mark-mode/deactivate-mark are preserved even if some command modifies them
(let ((transient-mark-mode transient-mark-mode)
      (deactivate-mark deactivate-mark))
   ...
)
```

Do you think this is a good option? Or is this what you suggested above?

wasamasa:
That's what I had in mind, the same pattern can be found in other places in the code.

edkolev:
Done, I've pushed the change.

wasamasa:
Thanks!

jojojames:
Thanks @edkolev. This is awesome!

I wonder how g/d performs against 'delete-matching-lines' now. I'll have to test it out myself.

edkolev:
@jojojames, you're welcome!

PR #1022, opened 2018-02-22
Treat going to marks jumps

edkolev:
This PR fixes the following scenario:

- `make emacs`
- enter in a fresh buffer
```
111
222
333
```
- go to first line
- set mark with `mm`
- using `j`, go to last line
- jump to the `m` mark with
```
`m
```
- you're on the first line now
- hit `C-o` - point should jump back to 3rd line, but it doesn't

As always, please let me know if the PR can be improved.

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -732,6 +732,7 @@ Columns are counted from zero."
   :keep-visual t
   :repeat nil
   :type exclusive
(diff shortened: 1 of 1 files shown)


wasamasa:
For a PR of this complexity, I doubt you can do better.

Issue #1026, opened 2018-03-05
evil-insert-newline- functions should respect use-hard-newline


wyleyr:
# Issue type
- Bug report

# Environment

Emacs version: 25.1.1
Operating System: Debian GNU/Linux (9.3)
Evil version: 1.2.13 (evil-20180222 from MELPA)
Evil installation type: MELPA
Graphical/Terminal: Graphical, under X
Tested in a `make emacs` session (see [CONTRIBUTING.md](https://github.com/emacs-evil/evil/blob/master/CONTRIBUTING.md#bug-reports)): No

# Reproduction steps
- Start Emacs and open a new buffer in Normal state (`C-x b test RET`)
- Turn on hard newlines (`M-x use-hard-newlines y`)
- Insert a newline with Enter key and return to Normal state (`i RET ESC`)
- Observe that this newline has the "hard" property set (`M-: (text-properties-at (1- (point)))`)
- Insert newline below with evil-insert-newline-below and return to Normal state (`o ESC`)
- Observe that this newline does not have the "hard" property set (`M-: (text-properties-at (1- (point)))`)

# Expected behavior
evil-insert-newline-{above/below} functions should respect use-hard-newline: pressing "o" or "O" within Normal state should insert a hard newline character if use-hard-newlines is t.

(I *think* they should, anyway. I am an Evil user but not a vi/vim user. Maybe vi/vim has a concept of "hard newline" that differs from the behavior of "o"/"O", and has a different keybinding in Evil?)

# Actual behavior
Currently, evil-insert-newline-{above/below} just call `(insert "\n")` to insert newlines.  This does not cause the inserted newlines to be marked "hard" when use-hard-newlines is t.

# Further notes
A quick fix seems to be to replace `(insert "\n")` with simply `(newline)` (though the latter performs some motions in interactive use that might not be desirable).

wasamasa:
I've never used that feature and haven't seen it in Vim, but I agree that Evil would be better off respecting that setting.

wyleyr:
There are a couple of other functions (evil-visual-paste, evil-read, evil-yank-line-handler, evil-yank-block-handler) that also just do `(insert "\n")`.  Should they also take use-hard-newlines into account?

wasamasa:
Closed by #1027.

PR #1027, opened 2018-03-05
Respect use-hard-newlines in evil-insert-newline-{above/below}

wyleyr:
Call `newline`, which respects use-hard-newlines.

This is a (naive initial attempt at a) fix for #1026.

### DIFF
--- a/evil-common.el
+++ b/evil-common.el
@@ -1886,7 +1886,7 @@ closer if MOVE is non-nil."
 with regard to indentation."
   (evil-narrow-to-field
     (evil-move-beginning-of-line)
-    (insert "\n")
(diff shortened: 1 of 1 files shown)


wasamasa:
`newline` isn't good for non-interactive use as it does too many things. I've checked the Emacs sources and found code that does `(if use-hard-newlines hard-newline "\n")`.

wyleyr:
Great, thanks for looking that up!  Here's a revised version that uses that code instead.

wasamasa:
Thanks!

tam5:
`newline` preserves text properties better, why can't we use that again?

PR #1029, opened 2018-03-06
Use string-match-p when possible

rgrinberg:
In some situations, we don't use the match data. So we might as well not capture
it.


wasamasa:
Thanks!

PR #1036, opened 2018-03-28
Add optional `:tag-face' keyword to `evil-define-state'.

fritzgrabo:
The given tag-face is used to add text properties to the state's mode line indicator.
This also creates an `evil-STATE-state-tag-face` variable for easy customization.

The keyword and the variable accept the same values as the `face` text property (see https://www.gnu.org/software/emacs/manual/html_node/elisp/Special-Properties.html)

evil tag face

### DIFF
--- a/evil-core.el
+++ b/evil-core.el
@@ -378,10 +378,12 @@ then this function does nothing."
 
 (defun evil-generate-mode-line-tag (&optional state)
   "Generate the evil mode-line tag for STATE."
-  (let ((tag (evil-state-property state :tag t)))
+  (let ((tag (evil-state-property state :tag t))
+        (tag-face (evil-state-property state :tag-face t)))
     ;; prepare mode-line: add tooltip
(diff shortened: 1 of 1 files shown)


wasamasa:
This looks good to me, but wouldn't it be sufficient to just propertize the tag face?

fritzgrabo:
Lol, you're right -- I didn't quite think of that.

FWIW, if anyone's interested: these settings yield the same results as the screenshot above:

```lisp
  (setq evil-normal-state-tag (propertize " <N> " 'face 'highlight))
  (setq evil-emacs-state-tag (propertize " <E> " 'face '(:background "red")))
```

If anything, my PR would provide a slightly easier way of configuring just the face of a predefined state w/o touching anything else? That could be a plus for novice evil/emacs users, but I'm not convinced that justifies the additional code.

wasamasa:
Yeah, same here.

Issue #1037, opened 2018-04-04
evil-initialize-local-keymaps is a memory hog and exacerbated by mmm-mode


aaronjensen:
# Issue type

- Perf bug report/Enhancement request

# Environment

Emacs version: GNU Emacs 27.0.50 (build 1, x86_64-apple-darwin17.4.0, NS appkit-1561.20 Version 10.13.3 (Build 17D102)) of 2018-03-30
Operating System: macOS High Sierra
Evil version: Evil version evil-git-f5d47df3
Evil installation type: Melpa
Graphical/Terminal: Graphical
Tested in a `make emacs` session (see [CONTRIBUTING.md](https://github.com/emacs-evil/evil/blob/master/CONTRIBUTING.md#bug-reports)): No

# Reproduction steps
- Install markdown-mode and mmm-mode
- Start Emacs
- `M-x eval-buffer` with (first fix your elpa path):

    ```elisp
    (let ((default-directory  "~/.emacs.d/elpa/27.0"))
      (normal-top-level-add-subdirs-to-load-path))
    (require 'package) ;; You might already have this line
    (let* ((no-ssl (and (memq system-type '(windows-nt ms-dos))
                        (not (gnutls-available-p))))
           (url (concat (if no-ssl "http" "https") "://melpa.org/packages/")))
      (add-to-list 'package-archives (cons "melpa" url) t))
    (require 'markdown-mode)
    (require 'mmm-mode)
    (markdown-mode)
    (mmm-mode-on)
    (setq garbage-collection-messages t)
    ```

- `M-x profiler-start` select mem
- Go into insert mode and hold a letter

# Expected behavior

Few garbage collections, little memory used

# Actual behavior

Garbage collections every few characters. Megs of memory used very quickly

# Further notes

Here is a sample profile:

```
+ command-execute                                          14,493,471  48%
- redisplay_internal (C function)                           9,124,634  30%
 - jit-lock-function                                        8,956,466  30%
  - jit-lock-fontify-now                                    8,956,466  30%
   - jit-lock--run-functions                                8,951,186  30%
    - run-hook-wrapped                                      8,951,186  30%
     - #<compiled 0x46d6c171>                               8,951,186  30%
      - font-lock-fontify-region                            8,951,186  30%
       - mmm-fontify-region                                 8,951,186  30%
        - mapc                                              4,922,690  16%
         - #<compiled 0x40e78771>                           4,922,690  16%
[...]

PR #1039, opened 2018-04-08
Use local-variable-p instead of assq

aaronjensen:
assq uses more memory for some reason. See #1037

### DIFF
--- a/evil-common.el
+++ b/evil-common.el
@@ -2180,7 +2180,7 @@ Variables pertaining to Transient Mark mode are listed in
     (when (and (boundp var)
                (not (assq var evil-transient-vals)))
       (push (list var (symbol-value var)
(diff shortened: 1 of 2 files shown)


aaronjensen:
Yeah, I don't know. There was an `(and ... t)` in there that I stripped out too. I'm guessing that was just accidentally left in there and there wasn't a good reason for it. Thanks for the quick merge.

TheBB:
That probably wasn't an accident, `(and x t)` is a way of normalizing booleans to `t` or `nil`, i.e. mapping all non-nils to `t`.

Seems like `local-variable-p` returns only `t` or `nil` though.

aaronjensen:
Ah, good to know, thanks.

Issue #1038, opened 2018-04-07
Backspace characters make pasting from a register fail. 


fiveNinePlusR:
# Issue type
- Bug report

# Environment

Emacs version: GNU Emacs 25.1.1 (x86_64-apple-darwin15.6.0, Carbon Version 157 AppKit 1404.47) of 2017-01-05
Operating System: macos high sierra
Evil version: evil-git-0fa3658cd
Evil installation type: spacemacs
Graphical/Terminal: gui
Tested in a `make emacs` session (see [CONTRIBUTING.md](https://github.com/emacs-evil/evil/blob/master/CONTRIBUTING.md#bug-reports)): No

# Reproduction steps
- Start Emacs
press `qritestt<backspace>q`

then press `"rpq`

# Expected behavior
should paste a string

# Actual behavior
it errors because there was a backspace character in the string. `evil-vector-to-string` and naive string creation are the culprit.

# Further notes
none

fiveNinePlusR:
seems they print out the termcap representation of the keys if they are unprintable except for some cases. 

`<80>bs` is the code for backspace ran a few other tests and the function keys are `k1-k9`, ` k;`, `F1-F19`

a full list is here http://vimdoc.sourceforge.net/htmldoc/term.html#builtin-terms

I would say that's an implementation detail of vim that probably doesn't make sense to fully emulate. whatever `edit-last-kbd-macro` does to turn characters into printable representations probably makes sense here too.

wasamasa:
Indeed. We can't even properly emulate it because Vim displays control characters (a newline/carriage return) here while we display two normal characters (`^` followed by `J` or `M`). It seems that the keyboard macro editor uses `edmacro-format-keys` for this purpose, we can use it in case something is a vector. For the case of something being a string, not sure, I'd keep the existing code for that.

PR #1040, opened 2018-04-09
Bugfix 1038 handle vector macros

wasamasa:
This includes a somewhat related change I'm not 100% sure about. When deprecating versions, shall we use the upcoming stable version? It can't be the current stable version because it already happened and doesn't include the deprecation. The next one will, however we'd either have to declare the next version number in advance or if we change our mind (like, when making a minor version instead of patch version bump) change all deprecation version numbers.

The actual fix uses the same function as the keyboard macro editor for representing a vector. I'm not 100% happy with the representation, but it's better than having it fail for the user.  I've left the string mogrification alone for now which currently replaces `\n` with `^J`, we could achieve a similar effect to Vim by replacing `\n` with `\r`, but I'm not sure whether we can rely that it will be printed as control character.

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -1878,7 +1878,7 @@ The return value is the yanked text."
                 (insert-for-yank text))
             ;; no yank-handler, default
             (when (vectorp text)
-              (setq text (evil-vector-to-string text)))
+              (setq text (edmacro-format-keys text)))
             (set-text-properties 0 (length text) nil text)
             (push-mark opoint t)
             (dotimes (i (or count 1))
@@ -1924,7 +1924,7 @@ The return value is the yanked text."
                 (insert-for-yank text))
             ;; no yank-handler, default
             (when (vectorp text)
-              (setq text (evil-vector-to-string text)))
+              (setq text (edmacro-format-keys text)))
             (set-text-properties 0 (length text) nil text)
             (unless (eolp) (forward-char))
             (push-mark (point) t)
@@ -3083,10 +3083,13 @@ If ARG is nil this function calls `recompile', otherwise it calls
     :entries
     (cl-loop for (key . val) in (evil-register-list)
              collect `(nil [,(char-to-string key)
-                            ,(or (and val
-                                      (stringp val)
-                                      (replace-regexp-in-string "\n" "^J" val))
-                                 "")]))))
+                            ,(cond
+                              ((stringp val)
+                               (replace-regexp-in-string "\n" "^J" val))
+                              ((vectorp val)
+                               (edmacro-format-keys val))
+                              ((not val)
+                               ""))]))))
 
 (evil-define-command evil-show-marks (mrks)
   "Shows all marks.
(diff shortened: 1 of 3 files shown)


edkolev:
@wasamasa I think `1.2.14` is fine for a version number. If we go from `1.2.13` to `1.3.0` and we forget to replace  
 s/`1.2.14`/`1.3.0`/ there won't be much harm - it would be definitely an inconvenience for a user to find out something was deprecated by a version which was never released, but I don't think the pain would be that big.

Regarding the bugfix - I think we could introduce an insignificant bug while we fix a significant one. However, I can't make the call what's insignificant and what isn't.

wasamasa:
Right, if people want to save/restore registers, they better do it via elisp anyways, just like all other Emacsers do. I'll change that test then to ensure *something* is pasted without any error.

noctuid:
I'm not sure if I'm understanding the issue correctly, but if this would prevent pasting (`"qp`), editing the macro, and then copying back into the same register (`"qy$`) (which seems to be the case based on the removed tests), then that's not a minor bug, that's arguably a major bug. This is a pretty common use case in vim as well (e.g. tip 72: Edit the Contents of a Macro from Practical Vim).

wasamasa:
Yes, instead you'd have to use an actual macro editor.

noctuid:
I think it would be better keep the original vim behavior if reasonably possible (not sure if it is though). For example, I can yank `iwordab` into  a register (`(kbd "i worda DEL b C-b C-d ESC")`), and it executes exactly as expected (inputs "word"). This isn't possible for all keys (e.g. non-characters like the directional keys). I don't think that there is a generic function for converting a symbol (e.g. `backspace`) to the corresponding character, but `evil-vector-to-string` could be extended to do this. Maybe this could be an option, but it doesn't seem like an ideal solution since it can't handle all keys.

On the other hand, I guess it's more simple and consistent to print all vectors the same. I think it would be preferable to use `key-description` over `edmacro-format-keys` though for readability unless there was some specific other reason `edmacro-format-keys` was chosen. Potentially, a prefix argument could be added to `evil-yank` that would modify the yanked text with `kbd`, so that you could still edit macros in any buffer, just using `kbd` format instead.

wasamasa:
The problem with the string representation is that it was chosen to be parsable by `read`. In Vim's case, something far more elaborate is done by using a terminal-specific scheme as mentioned on https://github.com/emacs-evil/evil/issues/1038#issuecomment-379591155. I challenge you to actually edit such escaped sequences without having to look at the manual. Having the ability to edit something human-readable is pretty neat to have, but seems to be a hard sell.

I prefer the `key-description` notation, but it worries me that the manual notes how `read-kbd-macro`/`kbd` is "a rough inverse for ‘key-description’". What does this even mean? `edmacro-format-keys` and `edmacro-parse-keys` are inverses of each other as demonstrated by the test I wrote.

Regarding modifying `evil-yank` behavior, prefix arguments are rather unusual in the Vim world and instead numeric arguments are implemented in terms of them. It would also bother me that you'd need to retry pasting a problematic macro with a prefix argument, that's not terribly ergonomic.

PR #1044, opened 2018-04-21
Evil window selection should trigger ElDoc.

hotpxl:
In `evil-define-motion`, especially [this line](https://github.com/emacs-evil/evil/blob/c53f8cac4897a360bc19fa4b49b556ab39c4cfee/evil-macros.el#L146), it will bind the command being defined to `eldoc-add-command` so that when this command is triggered, ElDoc will be triggered to redisplay documentation as well.

But the window selection commands (such as `evil-window-up`) are not "motion" and will not trigger ElDoc. So it's slightly annoying: after I switch to another window, and the cursor is now on top of something interesting, it will not show ElDoc documentation unless I move my cursor a bit.

This PR makes sure that window selection will trigger ElDoc message as well.

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -4159,6 +4159,10 @@ and redisplays the current buffer there."
             (select-window newwin)))))
     (balance-windows)))
 
+(with-eval-after-load 'eldoc
+  (when (fboundp 'eldoc-add-command-completions)
+    (eldoc-add-command-completions "evil-window-")))
+
 ;;; Mouse handling
 
 ;; Large parts of this code are taken from mouse.el which is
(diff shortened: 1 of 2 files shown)


wasamasa:
Hm. I can see this being annoying and why you'd want this behavior.  Judging from eldoc's sources, such a thing is done for window commands already, so what you've added is fine, too.  However, I'm not 100% happy with putting it after the definition of window commands, perhaps `evil-integration.el` is better for this purpose.

@TheBB what do you think?

TheBB:
I agree, this is fine but it's more suitable for evil-integration.

hotpxl:
Alright. I moved it to the end of the integration file.

TheBB:
Thanks, I squashed the commits for you.

PR #1048, opened 2018-05-10
Update README.md

peacetara:
Fix doc link, the old one didn't work for me.


wasamasa:
Both work for me.

wasamasa:
I don't think it's worth it catering for browsers not auto-detecting this. If a person really wants to, they can navigate to the file themselves.

PR #1055, opened 2018-05-24
Prevent errant scroll on mouse click

aaronjensen:
See https://debbugs.gnu.org/cgi/bugreport.cgi?bug=31546

This hasn't been accepted into emacs yet. It's probably worth waiting until the
final version of the patch is applied.


TheBB:
Thanks!

PR #1062, opened 2018-06-10
support `n` option for :sort command (fix #1061)

### LINKED ISSUE #1061: [n] option for `:sort' method is missing.
# Issue type

- Enhancement request

# Environment

Emacs version: 26.1
Operating System: Ubuntu 16.04
Evil version: 1.2.13
Evil installation type: MELPA
Graphical/Terminal:  under X

# Reproduction steps
- Start Emacs
- type below
101 a
1 b
33 c
10005 d
200 e
- gg V G :sort

# Expected behavior
1 b
33 c
101 a
200 e
10005 d

# Actual behavior
1 b
10005 d
101 a
200 e
33 c

# Further notes
It will be great if [n] option for `:sort' method would be also implemented.

tttuuu888:
This is implementation of 'n' option for :sort command.
Review and pointing any bad part of code are always welcome.

Thanks for this great package!

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -34,6 +34,7 @@
 (require 'flyspell)
 (require 'cl-lib)
 (require 'reveal)
+(require 'sort)
 
 ;;; Motions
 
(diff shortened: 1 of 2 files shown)


edkolev:
This is great, the lack of `n` support has been bugging me for a long time!

I'm on mobile now, so I can't go into the details of the code now, but do you think emacs' own sort-numeric-lines can be reused here?

TheBB:
You have misspelled number as nubmer in several places.

Hopefully I can provide some better comments than that later.

As @edkolev says, great initiative!

tttuuu888:
@edkolev Thanks for your opinion. Maybe you meant sort-numeric-fields. sort-numeric-fields didn't work if the region include empty lines. That's why I added evil--sort-numeric-fields function although there is only minor change from sort-numeric-fields.

@TheBB  Sorry I missed the point. Thanks for the review.

dzop:
What about just removing all the blank lines and then using `sort-numeric-fields` like what @edkolev has suggested. Vim just puts all of the blank lines at the beginning of the region. I was able to get it working with this

```emacs-lisp
(defun evil-sort-numeric (beg end)
  (interactive "r")
  (let ((n 0))
    (save-excursion
      (goto-char beg)
      (goto-char (line-beginning-position))
      (while (< (point) end)
        (if (not (looking-at "[ \t]*$")) (forward-line)
          (setq n (1+ n))
          (setq end (- end (- (1+ (line-end-position))
                              (line-beginning-position))))
          (delete-region (line-beginning-position) (1+ (line-end-position))))))
    (sort-numeric-fields 1 beg end)
    (goto-char beg)
    (save-excursion
      (insert (make-string n ?\n)))))
```

tttuuu888:
@dzop Vim put blank lines on top with ':sort n' command and put it on bottom with ':sort! n'. I just wanted this work as similar as vim does.(Although still there is a little difference from vim. Vim groups decimal and hexadecimal numbers respectively.)

tttuuu888:
@dzop I am sorry I did get you wrong. Yes your approach is how vim treat blank lines and your code is quite less verbose than mine. I think using your evil-sort-numeric instead of my evil--sort-numeric-fields looks better.

dzop:
Actually I looked more at the way that Vim sorts numbered lines as well, the following is what I gathered for `sort n` (this is based on Vim 8.0):

- Hex numbers are treated as the number 0

Consider

```
0x03
1
0
0x2
-1
```

This get sorted into

```
-1
0x3
0
0x2
1
```
- Octal numbers are treated as decimal numbers

In the following, 023 is 19 in decimal and 0111 is 73 in decimal: 

```
20
74
023
0111
```

which gets sorted to

```
20
023
74
0111
```

- Non-numeric lines are placed before numeric lines

- Prefixed numbers are handled and so are suffixed numbers. This is related with the issue of hex numbers, the `0x` prefix of a hex number is considered as a 0 with a suffix when sorting.

Consider

```
a3
a2
a1
2x
```

this gets sorted into

```
a1
a2
2x
a3
```

This last point seems very useful to have. It looks like we can get the support for suffixes for free from `string-to-number`, but will have to do some work to support prefixes. But what should be done about hex and octal numbers? Should they be separated out into their own groups?

I've got a tentative solution that separates out decimal, octal, and hex numbers into their own groupings and supports numbers with suffixes at dzop/evil@1cd586606d8eb2c1d4dc8498b8eff6b0aeaaa21c.

It looks like to support prefixed numbers we can possibly remove the prefixes and save them in a text property to be restored after sorting  or make a custom sorting function like what @tttuuu888 was proposing.

tttuuu888:
@dzop I found Vim is grouping hexadecimal numbers so I thought Vim sort hex numbers separately but it was not! I didn't catch that hex numbers are just treated as 0.

Now your last commit is grouping hex, oct, int numbers respectively and even sort it properly. Working Vim's way, and even better. I think we have no reason this commit is not merged into main-line. I will close this pull request. Please open your pull request. I appreciate your job.

tttuuu888:
@dzop Could you please open your pull request of the commit? So that we can  close  #1061.
I am already using your code locally though :)

Issue #1051, opened 2018-05-16
Macros don't behave as expected when macro register was written with a yank command vs macro recording command


adlerd:
# Issue type

- Bug report

# Environment

Emacs version: GNU Emacs 25.3.1 (x86_64-pc-linux-gnu) of 2018-03-29
Operating System: Linux gentoo unstable (May 2018)
Evil version: 1.2.13
Evil installation type: MELPA (via portage)
Graphical/Terminal: terminal (via ssh, tmux, etc.)
Tested in a `make emacs` session (see [CONTRIBUTING.md](https://github.com/emacs-evil/evil/blob/master/CONTRIBUTING.md#bug-reports)): No

# Reproduction steps
- Start Emacs
- :new
- insert "ifoo^[", without quotes, with ^[ being the actual escape character (via C-v C-[)
- yank the line into a register, without the newline; i.e., 0"ay$
- execute this register as a macro (@a)

# Expected behavior
"foo" should be inserted after point and we should return to Normal mode

# Actual behavior
"foo" is inserted but we remain in Insert mode

# Further notes
- The same repro steps in Vim 8.0 work as expected; probably in a lot of other vi's too
- I think evil-execute-macro should sanitize its input w/r/t escape just like evil-ex-normal does before passing to execute-kbd-macro; but I'm really not sure what the other consequences of this would be.
- Here's some alternative repro steps which make this look more like a register bug than a macro bug, in case that seems more problematic, since the problem is that paste+yank isn't idempotent:
  - record a macro into register a using the key sequence qaifoo<ESC>q
  - on a fresh line, paste register a then yank the line back into the same register ("ap0"ay$)
  - execute the macro
  - results: same as above. From a vim user's perspective, the register contents shouldn't have changed; from evil's perspective, they are subtly different.

wasamasa:
The core issue here is that Evil reuses Emacs macros.  Those can be represented as strings or vectors.  Vectors cannot be pasted as is, so you'd need to translate them to strings, but then you'd also need to parse them which is anything but trivial.  Vim goes for strings only and terminal-specific escapes, I don't really want to reimplement that, just to be able to yank keyboard macros into registers and re-execute them.

Anyway, my opinion on this is that when in Rome, do as the Romans do.  For keyboard macros this means to use the dedicated editor and use elisp code to restore them programmatically.  More thoughts on this can be found in https://github.com/emacs-evil/evil/issues/1038 and https://github.com/emacs-evil/evil/pull/1040.

PR #1064, opened 2018-06-13
Evil macro is recorded as kmacro in same time properly.

tttuuu888:
This commit has to do with #1051 and my comment there.

Currently once we record evil macro, we cannot edit it properly. Meanwhile evil macro is record as kmacro in same time but only the last evil macro remains in kmacro view ring.

This commit make evil macro properly recorded as kmacro in same time so that we can change previous evil macro at least using kmacro view ring.(`kmacro-cycle-ring-previous` and `kmacro-edit-macro` and finally use it with `kmacro-call-macro`)

There is only minor change and I found no harm so far.  Passed ert test too. However any other approach and opinion are welcome. 

Thanks for this great package.

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -2046,7 +2046,7 @@ will be opened instead."
    ((and evil-this-macro defining-kbd-macro)
     (setq evil-macro-buffer nil)
     (condition-case nil
-        (end-kbd-macro)
+        (kmacro-end-macro nil)
       (error nil))
     (when last-kbd-macro
       (when (member last-kbd-macro '("" []))
@@ -2065,7 +2065,7 @@ will be opened instead."
     (when defining-kbd-macro (end-kbd-macro))
     (setq evil-this-macro register)
     (evil-set-register evil-this-macro nil)
-    (start-kbd-macro nil)
+    (kmacro-start-macro nil)
     (setq evil-macro-buffer (current-buffer)))
    (t (error "Invalid register"))))
 


tttuuu888:
For those who might be interested, helm users can see whole evil macro list with `helm-execute-kmacro` properly after this commit. And it is also able to modify the macro on the list since recent helm update. (https://github.com/emacs-helm/helm/pull/2057)
However the modification is not synchronized with evil macro. The modified macro can only be executed by kmacro execution.

bigodel:
> However the modification is not synchronized with evil macro. The modified macro can only be executed by kmacro execution.

i have been using this modified version for a couple of weeks now, with Emacs 28 and evil 1.14.0, and i see no problems with it! i actually prefer this functionality over the previous one, but i can see how it could break some people's workflow, so i propose it be a use option. something like `evil-want-kmacro`

tttuuu888:
@bigodel Thanks for the nice comment. 

However, I don't think this update will break the workflow of existing users. Because this update, 1. adding the macro recorded as `evil-record-macro` to the `kmacro ring`, 2. making the added kmacro editable. Of these, 1 is just that a new macro is added to the `kmacro ring`, and 2 is that the evil macro does not change even if the kmacro is modified. Therefore, I think that this update will only add an option to properly modify the evil macro to the user, and will not affect the existing user's workflow. 

Any additional comments would be appreciated. thank you.

tttuuu888:
@bigodel Since this pull request was branched from a very old code, so I opened a new pull request while updating it. Additional comments are welcome at #1554.
thank you.

tomdl89:
> > However the modification is not synchronized with evil macro. The modified macro can only be executed by kmacro execution.
> 
> i have been using this modified version for a couple of weeks now, with Emacs 28 and evil 1.14.0, and i see no problems with it! i actually prefer this functionality over the previous one, but i can see how it could break some people's workflow, so i propose it be a use option. something like `evil-want-kmacro`

@bigodel how can you see "how it could break some people's workflow"? I think @tttuuu888 's reply made sense, but maybe I'm missing something?

PR #1069, opened 2018-06-28
Fix evil-set-cursor when spec is a function

justbur:
Docstring says that the spec can be a zero-argument function, but
evil-set-cursor incorrectly handles such a function like a list preventing the
function from being executed.


edkolev:
There's a commented-out debug message, committed by accident probably.

justbur:
Fixed and force pushed. Thanks

TheBB:
Thanks!

Issue #843, opened 2017-06-01
Searching for substring that an org-mode link contains freezes Emacs


rieje:
# Issue type
Bug report

# Environment
Emacs version: 25.2-1
Operating System: Arch Linux
Evil version: 20170531.1149
Evil installation type: MELPA
Graphical/Terminal: X

# Reproduction steps
- Delete everything in emac.d for a fresh install. Create init.el with the following contents:
     
    (require 'package)
    (add-to-list 'package-archives '("melpa" . "http://melpa.org/packages/"))
    (package-initialize)
- Start emacs, do the following:
    
    M-x package-refresh-contents
    M-x package-install RET evil
    M-x evil-mode
    C-x C-f <org-file containing a shortcut-link> 
    /<substring of a link that is shortcuted> followed by a letter 

For example, the org-mode has the shorcut link: `[[https://www.reddit.com/r/vim/comments/626no2/vim_without_nerd_tree_or_ctrlp/][file navigation and search]]`. The URL contains the substring `ctrl`, so you search via `/ctrl` and Emacs will hang.

# Expected behavior
Should search for the substring without issues.

# Actual behavior
Emacs freezes until I cancel it via `C-g`.

# Further notes

This issue only seems to affects links are shortcut-hyperlinked (i.e. when org-mode converts the format [[url][name-of-link]] to a clickable name_of_link. Also, if the full url is anywhere else in the document (i.e. not shortcut-linked by org-mode), then you won't have problems searching for the substring.

wasamasa:
Thanks, I can reproduce this one.  After toggling link display with `M-x org-toggle-link-display` the issue disappears, so this suggests that either something is wrong in Evil's wrapper around isearch or Org's hacks to isearch (the sources contain stuff for dealing with invisibility, but it's not obvious whether this applies to folded content only or links as well).

Alexander-Shukaev:
See last workaround for clue: syl20bnr/spacemacs#3623

Miciah:
I made #1070 to fix this issue.  Feedback welcome!

Notes: To reproduce this issue, I had to make sure `evil-search-wrap` was set to `t`, and I had to make sure Evil was using isearch: `(evil-select-search-module 'evil-search-module 'isearch)`.

PR #1070, opened 2018-07-12
Fix bound of wrapped search in evil-search-function

Miciah:
`evil-search-function` performs an initial search starting from point.  If that search fails and search wrapping is enabled, `evil-search-function` then performs a second search, this time from the start of the buffer.  However, `evil-search-function` was using the same bound for both searches.  This commit changes `evil-search-function` to use the first search's starting position as the search bound for the second search.

This commit fixes issue #843.

* `evil-search.el` (`evil-search-function`): Use the initial search's start as the bound for the second search when searching wraps.

### DIFF
--- a/evil-search.el
+++ b/evil-search.el
@@ -191,10 +191,10 @@ of the buffer."
            result)
        (setq result (funcall search-fun string bound
                              ,(if wrap t 'noerror) count))
-       (when (and ,wrap (null result))
+       (when (and ,wrap (null result) (> start 1))
          (goto-char ,(if forward '(point-min) '(point-max)))
          (unwind-protect
-             (setq result (funcall search-fun string bound noerror count))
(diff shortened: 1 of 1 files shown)


evil-search.el
```diff
@@ -191,10 +191,10 @@ of the buffer."
            result)
        (setq result (funcall search-fun string bound
                              ,(if wrap t 'noerror) count))
-       (when (and ,wrap (null result))
+       (when (and ,wrap (null result) (> start 1))
```
dzop: This should actually be `(> start (point-min))` since the buffer may be narrowed.

Also this only considers searching in the forward direction, we should also consider searching in the backward direction, i.e. when `forward` is `nil`.

TheBB:
Thanks for this PR. I've implemented a more robust solution in #1221 crediting you.

PR #1071, opened 2018-07-16
evil-show-registers: Show macros stored as vectors

Miciah:
When displaying register contents, `evil-show-registers` substitutes the empty string for non-string values.  If a register contains a macro, then the value is a key sequence, and a key sequence may be represented by either a string or a vector.  The latter would be substituted by the empty string.

This commit checks for vector values and uses `key-description` to display them.

* `evil-commands.el` (`evil-show-registers`): If a register contains a vector, display it by calling `key-description`.

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -3083,10 +3083,11 @@ If ARG is nil this function calls `recompile', otherwise it calls
     :entries
     (cl-loop for (key . val) in (evil-register-list)
              collect `(nil [,(char-to-string key)
-                            ,(or (and val
-                                      (stringp val)
-                                      (replace-regexp-in-string "\n" "^J" val))
-                                 "")]))))
+                            ,(cond ((stringp val)
+                                    (replace-regexp-in-string "\n" "^J" val))
+				   ((vectorp val)
+				    (key-description val))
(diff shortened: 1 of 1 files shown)


duianto:
Confirmed, this fixes the empty macro value in the `*evil-registers*` buffer.

Reproduction steps copied from the Spacemacs issue:
[Can't see macro contents in evil register with `evil-show-registers` #10116](https://github.com/syl20bnr/spacemacs/issues/10116)

### Reproduction guide 🐞
- Start Emacs
- Go to `*scratch*` buffer
- Type `q l i foobar <ESC> q`
- Type `@l`
- Type `"lp`
- Type `SPC r e`, or run `evil-show-registers`, or type `:registers` in normal mode

Current behaviour:
```
l
```
After applying this fix:
```
l          i f o o b a r <escape>
```

### System info
evil-20180723.1243
Emacs 26.1
Windows 10 Version 1803

timjstewart:
Hi, I'm still seeing this issue.  Will this be merged soon?  Thanks!

TheBB:
Thanks! Merged manually by cherry-picking.

PR #1072, opened 2018-07-18
Refine vdiff folding

edkolev:
This PR improves vdiff folding integration - vdiff is moved higher in`evil-fold-list`, making it take precedence over other folding modes. This makes sense because a user might have "long-lived" modes such as outline-mode/origami-mode always enabled and occasionally use vdiff which is a "short-lived" mode - there's no point in keeping vdiff enabled all the time. As a result, when the user is vdiff-ing a buffer which also happens to have outline-mode on, `zo` will open folds using vdiff instead of outline.

This PR also adds vdiff-3way-mode to `evil-fold-list`.

Refs https://github.com/emacs-evil/evil-collection/pull/179


edkolev:
I've been using this for a few days now, without any issues. Merging.

PR #1081, opened 2018-08-30
Support recursive minibuffers in Ex

### LINKED ISSUE #1079: Issue with Ex when recursive minibuffers are enabled
# Issue type

- Bug report

# Environment

Emacs version: GNU Emacs 26.1 (build 1, x86_64-apple-darwin17.5.0, Carbon Version 158 AppKit 1561.4) of 2018-06-17
Operating System: macOS Sierra
Evil version: Evil version evil-git-3d9f76f
Evil installation type: El-Get
Graphical/Terminal: GUI

# Reproduction steps
- Start Emacs
- Set `enable-recursive-minibuffers` to `t`
- Type `:<M-x><C-g><C-g><M-x>a`

# Expected behavior
No error should occur when typing in the minibuffer after the second `M-x` command.

# Actual behavior
A `wrong type argument: strinp, nil` error is raised after pressing `<M-x>a`

# Further notes
I have tracked the problem to `evil-ex-current-buffer` being `nil` when `evil-ex-update` is called. In `evil-ex-update`, `evil-ex-current-buffer` is used as the buffer argument to `with-current-buffer` and when the buffer argument is `nil` in `with-current-buffer` a `wrong type argument: stringp, nil` error is raised.

In `evil-ex-setup`, `evil-ex-teardown` is added to the global value of `minibuffer-exit-hook` so on the first `C-g`, `evil-ex-teardown` is called in ` *Minibuf-2*`, created from the first `M-x` command, and is never called in ` *Minibuf-1*`, created from the Ex command. Again, this is because `evil-ex-teardown` is added to the global value of `minibuffer-exit-hook` and removes itself from that hook when called after ` *Minibuf-2*` exits. But then `evil-ex-update` is never removed from the local value of `after-change-functions` in ` *Minibuf-1*`.

The second `M-x` command will put you back in ` *Minibuf-1*` where `evil-ex-update` is still present in the local value of `after-change-functions`. This means that `evil-ex-update` will be called in a minibuffer not setup properly for Ex and this results in `evil-ex-current-buffer` being `nil` when `evil-ex-update` is called.

dzop:
Fixes #1079, see that issue for a detailed explanation of the problem. Instead of adding `evil-ex-teardown` to the global value of `minibuffer-exit-hook`, it is added to the local value of that hook in the Ex minibuffer so that it is called in the right minibuffer when recursive minibuffers are enabled.

### DIFF
--- a/evil-ex.el
+++ b/evil-ex.el
@@ -228,7 +228,7 @@ This function registers several hooks that are used for the
 interactive actions during ex state."
   (add-hook 'post-command-hook #'evil-ex-abort)
   (add-hook 'after-change-functions #'evil-ex-update nil t)
-  (add-hook 'minibuffer-exit-hook #'evil-ex-teardown)
+  (add-hook 'minibuffer-exit-hook #'evil-ex-teardown nil t)
   (when evil-ex-previous-command
     (add-hook 'pre-command-hook #'evil-ex-remove-default))
   (remove-hook 'minibuffer-setup-hook #'evil-ex-setup)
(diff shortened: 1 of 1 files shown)


TheBB:
Thanks, good call!

PR #1087, opened 2018-09-11
Split keybindings out from evil-integration

jojojames:
@edkolev @ambrevar @noctuid @justbur @wasamasa @syl20bnr @TheBB 

Following from: https://github.com/emacs-evil/evil-collection/issues/60

Most everyone seems to be on board to pull the keybindings out so that evil-collection can more easily leverage evil's evil-integration.el.

I took a quick look and pulled out the keybindings from evil-integration and put them in a separate file.

Any comments appreciated!

### DIFF
--- a/evil-keybindings.el
+++ b/evil-keybindings.el
@@ -1,0 +1,121 @@
+;;; evil-keybindings.el --- Add some Evil keybindings to other modules
+
+;; Author: Vegard Øye <vegard_oye at hotmail.com>
+;; Maintainer: Vegard Øye <vegard_oye at hotmail.com>
+
+;; Version: 1.2.13
+
+;;
+;; This file is NOT part of GNU Emacs.
+
+;;; License:
+
+;; This file is part of Evil.
+;;
+;; Evil is free software: you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+;;
+;; Evil is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+;;
+;; You should have received a copy of the GNU General Public License
+;; along with Evil.  If not, see <http://www.gnu.org/licenses/>.
+
+;;; Commentary:
+;; This provides a set of keybindings for other emacs modes. This also includes
+;; setting up the initial evil state of those other modes.
+
+(require 'evil-maps)
+(require 'evil-core)
+(require 'evil-macros)
+(require 'evil-types)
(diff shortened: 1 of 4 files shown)


evil-keybindings.el
```diff
@@ -0,0 +1,111 @@
...
+
+(require 'evil-maps)
+(require 'evil-core)
+(require 'evil-macros)
+(require 'evil-types)
+(require 'evil-repeat)
+
+;; etags-select
+;; FIXME: probably etags-select should be recomended in docs
+(eval-after-load 'etags-select
```
Ambrevar: Why not using `with-eval-after-load`?

TheBB: At least for now, Evil still supports Emacs 24.1.

evil-keybindings.el
```diff
@@ -0,0 +1,111 @@
+;;; evil-integration.el --- Integrate Evil with other modules
```
TheBB: Filename.

evil-keybindings.el
jojojames: Thanks for the explanation @TheBB .

I did a straight copy from evil-integration.el to reduce the blast radius of the change @Ambrevar .

evil-keybindings.el
jojojames: Thanks, nice catch.

noctuid:
Since initial state changes and `evil-make-overriding-map` calls are also being moved (which I agree with) in addition to keybindings, you should probably explicitly document that as well.

jojojames:
Good idea, where to document this?

noctuid:
I'd mention what is expected to be included in the file in both `evil-integration` and `evil-keybindings` (in a commentary section), so that it's hopefully clear for anyone who wants to add something to those files (assuming they read the commentary). I'd also probably mention it in the docstring for `evil-want-keybindings` (though with a more high level description that doesn't mention the actual evil function/variables; for example, say that functionality in `evil-keybindings` may remove (mainly) insertion keybindings by changing the initial state and may have some mode's keybindings override evil's defaults).

noctuid:
I also think that might be better to rename the file (and maybe the variable; e.g. add `-integration` or something) because if I didn't already know the difference, I might be confused about `evil-keybindings.el` vs. `evil-maps.el`.

jojojames:
I'll add something for the commentary section but my documentation skills aren't too good. :)

> I also think that might be better to rename the file (and maybe the variable; e.g. add -integration or something) because if I didn't already know the difference, I might be confused about evil-keybindings.el vs. evil-maps.el.

I like the shortness of the current file/variable. evil-keybinding-integration and evil-want-keybinding-integration is a mouthful! Hopefully the commentary would be revealing enough? Not a strong opinion though.

jojojames:
Added some commentary, open to more polish, of course.

TheBB:
Thanks! I made some minor changes to the comments and merged manually.

jojojames:
Awesome, thanks @TheBB

Issue #805, opened 2017-04-05
Evil Search and Clear Across multiple Windows


naclander:
This is the same issue as #803 , but I will try to reword it to make it clearer.

In vim, searching for a keyword causes that keyword to highlight both in the current active buffer, as well as other open buffers that have other files opened.

In Emacs/Evil, one can achieve this same functionality with `evil-search`, except for the last part. Specifically, having all matches show up highlighted across all opened Windows ( where each window has a buffer with a different file opened ).

Evil should behave the same way as vim, and highlight all matches across all windows automatically. This could be extended by giving users the ability to choose whether to highlight just the current window, all windows in the current frame, or across all buffers Emacs currently has opened.

Naturally, `evil-ex-nohighlight` should clear all matches across all windows. Or again, the user should have the ability to toggle this.

I hope this description is a little more clear in regards to the feature request I'm describing.

PR #1091, opened 2018-09-21
Search highlighting across multiple buffers

ChoppinBlockParty:
Provide a pilot version as PoC.
No cross-buffer highlighting for substitution in this version.
On the way improve overall performance of highlighting
by removing `run-at-time` and introducing `window-end`
with `update` equal `t`.

Search pattern in Vim is highlighted in every visible buffer, this does not sound like a big deal, however it is very useful in many small situations. This feature is also addressed in issue #805.

Please, try it out, feedback is highly appreciated. I have tested it quickly on few buffers with some large content as well (like ~20k) worked well, I think the previous version with timer was lagging much more. After some feedback and a few days of real life testing, I will polish this one to be ready for merge.


ChoppinBlockParty:
Hello Guys, I have been using this one for more than a year it is working well. Can we consider merging it?

TheBB:
Sorry. :-)

I'll have a look this week hopefully.

PR #1094, opened 2018-09-27
Fix `evil-ex` called with a current line prefix

dzop:
Previously when a current line prefix was specified (i.e.
giving a numeric argument of 1 before calling `evil-ex`),
Ex attempted to `concat` together a string wrapped in a
list with `evil-ex-initial-input`, but if `concat` is
passed a list it expects it to be a list of characters not
strings.


TheBB:
Thanks!

PR #1096, opened 2018-10-01
Fix bug with evil-swap-marker for global markers

yilinwei:
This fixes a bug with the `evil-swap-marker` implementation currently in evil.

**Reproduction**

- Open a file
- Create a local marker using `m a`
- Create a global marker using `m A` elsewhere
- Kill the buffer
- Fails to navigate to the buffer using `' A`

### DIFF
--- a/evil-common.el
+++ b/evil-common.el
@@ -1927,11 +1927,11 @@ otherwise, it stays behind."
           (setq alist (default-value 'evil-markers-alist)
                 marker (make-marker))
           (evil-add-to-alist 'alist char marker)
-          (setq-default evil-markers-alist alist))
+          (setq-default evil-markers-alist alist)
+          (add-hook 'kill-buffer-hook #'evil-swap-out-markers nil t))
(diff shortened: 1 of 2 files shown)


evil-common.el
```diff
@@ -1927,11 +1927,11 @@ otherwise, it stays behind."
           (setq alist (default-value 'evil-markers-alist)
                 marker (make-marker))
           (evil-add-to-alist 'alist char marker)
-          (setq-default evil-markers-alist alist))
+          (setq-default evil-markers-alist alist)
+          (add-hook 'kill-buffer-hook #'evil-swap-out-markers nil t))
          (t
           (setq marker (make-marker))
           (evil-add-to-alist 'evil-markers-alist char marker))))
-      (add-hook 'kill-buffer-hook #'evil-swap-out-markers nil t)
```
yilinwei: We move this hook since the only time we need it is when we have set a global mark

PR #1111, opened 2018-12-05
Allow C-g to abort evil-read-key

justbur:
This prevents evil-replace from inserting a literal C-g if C-g is typed after
entering replace state.


edkolev:
Looks good to me!

I think you can go ahead and merge.

justbur:
Thanks I will

PR #1121, opened 2019-01-25
Allow colon `:` for symbols in `binding` of `evil-ex-grammar`

Alexander-Shukaev:


Alexander-Shukaev:
Just a friendly ping to merge this change.

TheBB:
Sorry for the delay. Is there an example ex command where this is needed?

Alexander-Shukaev:
I defined a number of custom `ex` commands for myself, which begin with `:`, so that I can quickly type `::command` as `evil-ex` is itself `:`.

TheBB:
But if you define them yourself then you can choose different names. The ex parser is somewhat fragile and I'm reluctant to change it just for that.

I tried adding a test for this and as you can see it fails to parse properly. If you experience differently I'd like to see the command(s) you've added.

TheBB:
I guess it's the colon in the first branch of the regexp that's greedily matching just the first colon of `:command`.

Alexander-Shukaev:
Apologies, I think you misunderstood that I meant custom `ex` commands (i.e. defined with `evil-ex-define-cmd`), not the regular Emacs `interactive` functions.  In this case, the test should be against `(evil-ex-parse-command ":command")`.

TheBB:
It still doesn't work as expected.

    ELISP> (defun :command () (interactive))
    :command
    ELISP> (evil-ex-parse-command ":command")
    (":" . "command")

    ELISP> (defun command () (interactive))
    command
    ELISP> (evil-ex-parse-command "command")
    ("command" . "")

    ELISP> (defun comm:and () (interactive))
    comm:and
    ELISP> (evil-ex-parse-command "comm:and")
    ("comm:and" . "")

Do you see other behaviour?

Alexander-Shukaev:
Once again, you have to use `evil-ex-define-cmd` rather than plain `interactive` function.

TheBB:
Sorry. I misunderstood in which direction I was supposed to have misunderstood. Then my original comment stands: you can substitute the defuns with evil-ex-define-cmd in my previous comment and the behaviour is identical.

Alexander-Shukaev:
I get

```
ELISP> (evil-ex-parse-command ":command")
(":command" . "")
```

TheBB:
Okay. I rebased this branch on top of master, which also has #1203, so maybe that interferes (if you haven't updated the PR branch since). If not then I'm not sure why the difference.

Alexander-Shukaev:
Right, since I was rather customizing the variable, I had the following

```
 (binding "[[:alpha:]-:]+\\|[~&*@<>=:]+\\|[[:alpha:]-]+\\|!")
```

which works but is not what we want to release of course.  Sorry.

PR #1127, opened 2019-02-12
C-w support yank word when search-module is 'evil-search

jixiuf:


edkolev:
Could you describe what behaviour this is trying to achieve? `C-w` deletes the previous work in the search prompt in vim.

jixiuf:
same to `C-w` in isearch,   search words under point ,and you can press `C-w` multiple times

dzop:
Yes, actually `C-w` should have the opposite behavior. It should delete the previous word in the Ex minibuffer since that is how vim does it. That is more in line with how `evil` is meant to behave anyways since its main goal is to emulate vi/vim.

But there are variables like `evil-want-C-u-scroll` which take into account discrepancies between fundamental keys used in both Emacs and Vim but have different meaning. Perhaps this could be another one of those cases and we could have an
`evil-want-C-w-delete-in-search` variable or just piggyback off of `evil-want-C-w-delete`? I'm thinking the latter makes more sense since if someone is already used to the Emacs way of `C-w` meaning to kill text, they wouldn't want that behavior to be different in Ex/search. If this is done then it would probably make sense for `evil-ex-completion-map` to also obey `evil-want-C-w-delete`.

leungbk:
Now that `evil-want-C-w-delete` is used to opt into yanking, it would be great if this could get reviewed and merged. This is the only thing stopping me from using `evil-search` as my main search module.

evil-search.el
```diff
@@ -1289,6 +1290,30 @@ a :substitute command with arguments."
...
+
+(defun evil-search-yank-word (&optional arg)
+  "Pull next word from buffer into search string."
+  (interactive)
+  (let ((fwd-fn #'forward-word)
+        word)
+    (with-current-buffer evil-ex-current-buffer
+      ;; Start to initial point if C-w have never been hit.
+      (unless evil-ex-search-yank-point
+        (setq evil-ex-search-yank-point evil-ex-search-start-point))
```
leungbk, line 1304: I tried this branch today, and I don't think `evil-ex-search-start-point` is what we should use as a starting value for `evil-ex-search-yank-point`.

With a buffer containing

```
foo barbaz
foo bar123
```

If `point` is at the beginning of the buffer and I start to search forward and type `bar C-w`, I would expect the `C-w` to add `baz` to the search string, since this is what `isearch`'s `C-w` does. But right now we add `foo`, since `foo` is what is in front of `evil-ex-search-start-point`.

i'm thinking we should define another global variable, `evil-ex-search-current-match-beg` (distinct from `evil-ex-search-match-beg`, which seems to correspond to the last search that was actually exited), to be set by `evil-ex-search-goto-offset` that we can use here.

jixiuf: please review again

leungbk: Thanks for updating. I spotted several other things.

With a file containing

```
abc def ghi jkl
```

Point is at the beginning of the buffer.

- If I use M-x `isearch-forward` and type `def C-w C-w`, the search string is `def ghi jkl`. However, with this patch, if I use `evil-search-forward` with `evil-search-module` set to `evil-search`, the first `C-w` does not change the search string, and the second one consumes the `ghi`, so that the final search string is `def ghi`.

- If I use M-x `isearch-forward` and type `def C-w <backspace> C-w`, the search string is `def ghi`. With this patch, if I do the same search using `evil-search-forward` and the `evil-search` module, the search string is `def gh jkl`. `<backspace>` should reset the `evil-ex-search-yank-point` to something appropriate.

evil-vars.el
```diff
@@ -518,7 +518,7 @@ ubiquity of prefix arguments."
                (define-key evil-insert-state-map (kbd "C-u") 'evil-delete-back-to-indentation))))))
 
 (defcustom evil-want-C-w-delete t
-  "Whether `C-w' deletes a word in Insert state."
+  "Whether `C-w' deletes a word in Insert/Ex/search state."
```
leungbk, line 521: Elsewhere, the patch has

``` elisp
(define-key evil-ex-completion-map "\C-w" nil)
```

So for now the documentation shouldn't mention support for Ex state.

jixiuf: fixed.

evil-search.el
jixiuf: should be fixed

evil-vars.el
jixiuf: But it means if `evil-want-C-w-delete` is not nil, C-w in `evil-ex-completion-map` bind to `backward-kill-word`
if not C-w would call `kill-region` or whatever command  binding  in `global-map`.

```
(if evil-want-C-w-delete
    (define-key evil-ex-completion-map "\C-w" 'backward-kill-word)
    (define-key evil-ex-completion-map "\C-w" nil))
```

evil-search.el
leungbk: Looks good, thanks!

evil-vars.el
leungbk: Oh sorry, you're correct. I had the yanking functionality in mind when I commented on this, but you had it right the first time.

leungbk:
I encountered the following issue:

```
alphaBetCatDog
```

If `evil-ex-search-case` is `smart` and I use `evil-search-forward` with `evil-search` and type `bet C-w`, the search string changes to `betCat` and fails to match.

With `isearch-forward`, typing `bet C-w` changes the search string to `betcat`. I think this is the preferred behavior when `evil-ex-search-case` is set to `smart`.

jixiuf:
> `evil-ex-search-case`

fixed

leungbk:
> > `evil-ex-search-case`
> 
> fixed

With `isearch-forward` and smart casing, if an uppercase char is typed into the search string, the `C-w` invocations preserve case. For example, in a buffer with `alphaBetCatDog`, typing `M-x isearch-forward` and `Bet C-w` changes the search string to `BetCat`.

Currently, using the patched `evil-search` module and `evil-ex-search-case` set to `smart`, typing `Bet C-w` changes the search string to `Betcat`, which fails to match.

jixiuf:
> > > `evil-ex-search-case`
> > 
> > 
> > fixed
> 
> With `isearch-forward` and smart casing, if an uppercase char is typed into the search string, the `C-w` invocations preserve case. For example, in a buffer with `alphaBetCatDog`, typing `M-x isearch-forward` and `Bet C-w` changes the search string to `BetCat`.
> 
> Currently, using the patched `evil-search` module and `evil-ex-search-case` set to `smart`, typing `Bet C-w` changes the search string to `Betcat`, which fails to match.

fixed.

angrybacon:
Is this still being experimented with or is there another way to get the desired behavior?

tomdl89:
@angrybacon I'll take a look soon and merge if all looks good

tomdl89:
As soon as github actions recovers (https://www.githubstatus.com/) I will merge #1694 and close this.

tomdl89:
Closed by #1694

PR #1130, opened 2019-02-18
Better integration of window manipulation commands

dzop:
Commands like `evil-window-rotate-upwards` and `evil-window-move-very-top` only
consider the buffer in each window when manipulating the window tree, but if a
window has specific window properties that need to be maintained (like those in
`window-persistent-parameters`) that information is lost.

Refactor the window manipulation commands to use `window-state-get` and
`window-state-put` to retrieved and restore the state of the window tree so
that window properties in each window are maintained.

Also integrate the window manipulation commands with side windows, i.e. windows
with a non-nil `window-side` window parameter, by only manipulating the part of
the window tree that are not side windows.

* Add the macro `evil-save-side-windows` which removes all side windows on
  entering the macro environment and restores the side windows leaving.

* Add the function `evil-move-window` which does the work of
  `evil-window-move-far-left`, `evil-window-move-very-top`,
  `evil-window-move-far-right`, and `evil-window-move-very-bottom`. These
  functions were all identical, the only difference being which window was used
  to restore the `selected-window`.

* Use `window-state-(get|put)` to get the window tree and to restore it. This
  makes `evil-get-buffer-tree` and `evil-restore-window-tree` unnecessary and
  they are removed. Note that `window-state-(get|put)` also handles window
  parameters.

* In addition to ignoring side windows when moving a window, ignore them in
  `evil-window-rotate-downwards` and `evil-window-rotate-upwards`.

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -3834,6 +3834,18 @@ The 'bang' argument means to sort in reverse order."
 
 ;;; Window navigation
 
+(defmacro evil-save-side-windows (&rest body)
+  "Toggle side windows, evaluate BODY, restore side windows."
+  (declare (indent defun) (debug (&rest form)))
+  (let ((sides (make-symbol "sidesvar")))
+    `(let ((,sides (and (functionp 'window-toggle-side-windows)
+                        (window-with-parameter 'window-side))))
+       (when ,sides
+         (window-toggle-side-windows))
+       ,@body
+       (when ,sides
+         (window-toggle-side-windows)))))
+
 (defun evil-resize-window (new-size &optional horizontal)
   "Set the current window's width or height to NEW-SIZE.
 If HORIZONTAL is non-nil the width of the window is changed,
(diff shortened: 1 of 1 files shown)


evil-commands.el
```diff
@@ -3742,6 +3742,18 @@ The 'bang' argument means to sort in reverse order."
...
+  "Toggle side windows, evaluate BODY, restore side windows."
+  (declare (indent defun) (debug (&rest form)))
+  (let ((sides (make-symbol "sidesvar")))
+    `(let ((,sides (and (functionp 'window-toggle-side-windows)
+                        (window-with-parameter 'window-side))))
+       (when ,sides
+         (window-toggle-side-windows))
+       ,@body
+       (when ,sides
+         (window-toggle-side-windows)))))
```
ehahn: Shouldn't this macro use a `condition-case` block?

dzop: Yup your right there does need to be some error handling.

TheBB:
Thanks! I rebased this on master and the tests seem to pass fine.

I'm assuming this is needed for your jumplist refactoring work? Would you be willing to PR that at some point?

Will probably merge this soon.

PR #1134, opened 2019-02-27
Support to toggling in vdiff-3way-mode

edkolev:
Refs #1132


TheBB:
Appears to be in master already, somehow.

TheBB:
No it wasn't, but now it is.

Issue #1138, opened 2019-03-04
evil-jump-forward does not seem to work


nipunarora:
I use evil mode in spacemacs in the terminal mode and I've observed that `evil-jump-forward` does not seem to work.
There are multiple issues here:

1. `:jumps` seems to show all the positions, but not where you are similar to what vi is supposed to sho w with a `>` to denote current position
2. `c-o` seems to work fine with going backwards
3. neither `c-i` which indents, and `evil-jump-forward` seems to work as suggested

dzop:
Also see my comments about even more issues with the jumplist implementation in https://github.com/emacs-evil/evil/issues/732#issuecomment-454289474.

nipunarora:
@dzop is there an alternative? how can I get `evil-jump-forward` kind of feature?

dzop:
I have a branch containing my re-implementation of jumplists that fixes the problems that I mentioned and also shows the correct behavior of `:jumps` [here](https://github.com/dzop/evil/tree/jumplist) if you are comfortable with pulling in those changes in your own version of `evil`.

usq:
Any news here? I do have the same problem

dzop:
I plan on submitting an issue to discuss a possible transition to my branch for `evil-jumps`.  I'm not sure if there have been any fixes to the current implementation though.

kiennq:
Hello, is there update on this?

tshu-w:
same issue here

tomtomjhj:
There's another implementation of jumplist https://github.com/gilbertw1/better-jumper.
It integrates with evil and works fairly well. Also, it's bundled in doom-emacs. https://github.com/hlissner/doom-emacs/blob/19fa5e2559df406c3116400e7755b86e46ca217b/core/core-editor.el#L427

deep-calm:
"If you want to jump forwards with C-i in Spacemacs, like it does in Vim, then you have to set the dotspacemacs-distinguish-gui-tab variable to t, in .spacemacs. It's nil by default, and that makes it behave like TAB.". Direct excerpt from here https://github.com/syl20bnr/spacemacs/issues/9451#issuecomment-323514998.

wolfand11:
@deep-calm "set the dotspacemacs-distinguish-gui-tab variable to t" it's not work for me.

PR #1150, opened 2019-05-19
Treat all commands that change the current buffer as jump commands

ehahn:
In Vim, all commands that change the current file are jump commands. Evil previously implemented this using advices, but this did not catch all applicable commands: find-file, for example, was ignored.

By checking if a command changed the buffer in the post-command-hook, we can reliably catch all of buffer-changing commands.

This pull request also enables lexical binding in evil-jumps.el and includes some other minor refactoring.

### DIFF
--- a/evil-jumps.el
+++ b/evil-jumps.el
@@ -311,10 +311,25 @@ change the current buffer."
                (previous-pos (evil-jumps-struct-previous-pos struct)))
           (when previous-pos
             (setf (evil-jumps-struct-previous-pos struct) nil)
-            (if (and (not jumping-backward)
-                     (let ((previous-buffer (marker-buffer previous-pos)))
-                       (and previous-buffer
-                            (not (eq previous-buffer (window-buffer window))))))
+            (if (and
+                 ;; `evil-jump-backward' (and other backward jumping
+                 ;; commands) needs to be handled specially. When
+                 ;; jumping backward multiple times, calling
+                 ;; `evil-set-jump' is always wrong: If you jump back
+                 ;; twice and we call `evil-set-jump' after the second
+                 ;; time, we clear the forward jump list and
+                 ;; `evil--jump-forward' won't work.
+
+                 ;; The first time you jump backward, setting a jump
+                 ;; point is sometimes correct. But we don't do it
+                 ;; here because this function is called after
+                 ;; `evil--jump-backward' has updated our position in
+                 ;; the jump list so, again, `evil-set-jump' would
+                 ;; break `evil--jump-forward'.
+                 (not jumping-backward)
+                 (let ((previous-buffer (marker-buffer previous-pos)))
+                   (and previous-buffer
+                        (not (eq previous-buffer (window-buffer window))))))
                 (evil-set-jump previous-pos)
               (set-marker previous-pos nil))))))))
 
--- a/evil-jumps.el
+++ b/evil-jumps.el
@@ -66,6 +66,10 @@
 
 (defvar evil--jumps-jumping nil)
 
+(defvar evil--jumps-jumping-backward nil
+  "Set by `evil--jump-backward', used and cleared in the
+`post-command-hook' by `evil--jump-handle-buffer-crossing'")
+
 (eval-when-compile (defvar evil--jumps-debug nil))
 
 (defvar evil--jumps-buffer-targets "\\*\\(new\\|scratch\\)\\*"
@@ -79,7 +83,8 @@
 
 (cl-defstruct evil-jumps-struct
   ring
-  (idx -1))
+  (idx -1)
+  previous-pos)
 
 ;; Is inlining this really worth it?
 (defsubst evil--jumps-message (format &rest args)
@@ -239,6 +244,7 @@ POS defaults to point."
       (evil--jumps-push))))
 
 (defun evil--jump-backward (count)
+  (setq evil--jumps-jumping-backward t)
   (let ((count (or count 1)))
     (evil-motion-loop (nil count)
       (let* ((struct (evil--jumps-get-current))
@@ -275,6 +281,7 @@ POS defaults to point."
           (let* ((source-jump-struct (evil--jumps-get-current existing-window))
                  (source-list (evil--jumps-get-jumps source-jump-struct)))
             (when (= (ring-length (evil--jumps-get-jumps target-jump-struct)) 0)
+              (setf (evil-jumps-struct-previous-pos target-jump-struct) (evil-jumps-struct-previous-pos source-jump-struct))
               (setf (evil-jumps-struct-idx target-jump-struct) (evil-jumps-struct-idx source-jump-struct))
               (setf (evil-jumps-struct-ring target-jump-struct) (ring-copy source-list)))))))
     ;; delete obsolete windows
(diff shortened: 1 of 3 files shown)


ehahn:
How did I get Emacs to segfault?

TheBB:
Don't worry about the CI, it's fragile.

Alexander-Shukaev:
Looking forward to this being merged! 💯

dzop:
It looks like the reason why `find-file` was ignored is because it uses a combination of `display-buffer` and `select-window` as opposed to `switch-to-buffer`, see `pop-to-buffer`. Can we get away with just advising `select-window` to set a jump point? I think `switch-to-buffer` and `select-window` are the main primitives that cause `point` to jump to a different buffer so they should be the only advise needed.

ehahn:
On 10 Jun 2019 10:27 PM, Nathaniel Nicandro <notifications@github.com> wrote:It looks like the reason why find-file was ignored is because it uses a combination of display-buffer and select-window as opposed to switch-to-buffer, see pop-to-buffer. Can we get away with just advising select-window to set a jump point? I think switch-to-buffer and select-window are the main primitives that cause point to jump to a different buffer so they should be the only advise needed.

—You are receiving this because you authored the thread.Reply to this email directly, view it on GitHub, or mute the thread.
I don't think so. Functions like `other-window` also call `select-window` but shouldn't set a jump point.

dzop:
Couldn't you do some clever checking with around advice to determine if the selected window is a different buffer than the previously selected window? This would avoid having to add a new slot in the `evil-jumps-struct`.

It would also avoid another possible issue with your solution. Namely, that now only commands can set jump points. By removing the advice you get rid of the possibility of calling `switch-to-buffer` (or `split-window`) from some function and having that also set a jump point. I guess the question is do we want `evil-jumps` to be integrated more with primitive functions that alter the state of Emacs or with the command loop? It does appear to be a better solution to stick with the command loop since functions like `switch-to-buffer` can be called multiple times during the execution of a command and we wouldn't necessarily want to set a jump point for each of those calls.

Back to the `select-window` advice, we can do something like

```emacs-lisp
(defadvice select-window (around evil-jumps activate)
  (let ((previous-pos (point-marker))
        (previous-buffer (current-buffer)))
    ad-do-it
    (unless (eq previous-buffer (current-buffer))
      (evil-set-jump previous-pos))
    (set-marker previous-pos nil)))
```

I'm not entirely sure what the `evil--jumps-jumping-backward` variable is trying to solve, so not sure if it would also be needed in the above code.

ehahn:
We definitely should stick to the command loop. Using advises, we run the risk of missing a primitive. Even if we do advise all relevant primitives, there is no guarantee that future versions of Emacs won't add another.

More importantly, we just don't know if we *should* set a jump point when a primitive is called. As you wrote, a function may call a primitive multiple times. It could also call a primitive but undo the change later. I think `(save-window-excursion (select-window <whatever>))` is a no-op, for example.

If we want to set a function that is not a command to set a jump point we can still advise it separately. The only reason for that I can think of is when a command asynchronously changes the buffer. Is there any command, internal or in a package, that does that?

I don't think adding a new slot to `evil-jumps-struct` is a huge issue. Since `evil-jumps-slot` is never serialized, this only causes problems when upgrading evil without restarting Emacs. We could avoid that by explicitly checking if `evil-jumps-struct` has the new slot and upgrading it when necessary. In that case I think we should also migrate the variable to a proper window parameter while we're at it.

`evil--jumps-jumping-backward` is necessary because `evil--jump-backward` shouldn't add a new jump point so any solution that automatically detects buffer-switching commands needs something like it. I added a comment that explains this in more detail. By the way, I'm not sure if `evil--jump-backward` always sets a jump point when it should. I definitely trigger a bug when testing this but that's probably  #1138.

dzop:
I was mentioning avoiding adding a new slot if we were going the route of staying with advice since it wouldn't be needed with the `select-window` advice. But the command loop solution (which I think is the right approach) would need to keep track of where `point` is at before the command somehow. However I  think it would be better if we had a global variable to keep track of `point` before a command since there is only one command being executed at any given time. Having a slot in the `evil-jump-struct` implies that its some kind of per-window state.

> If we want to set a function that is not a command to set a jump point we can still advise it separately. The only reason for that I can think of is when a command asynchronously changes the buffer. Is there any command, internal or in a package, that does that?

Yes `point` can jump to some unexpected place during evaluation of a command, e.g. if a command calls `sleep-for` or `accept-process-output` which would allow timers or process filter functions to run. So there is a possibility to store `point` in `pre-command-hook` and have it move before setting a jump point. 

> In that case I think we should also migrate the variable to a proper window parameter while we're at it.

Yes we should definitely do this at some point. I have an entire re-implementation of `evil-jumps.el` that I've been using for over a year and half which implements the jumplist as a window parameter and fixes a lot of issues with the current implementation as I mentioned in #1138 and related.

I just need to update the tests and add some more for my branch to be pull request worthy. But I'm hesitant to work on it and submit a pull request. Seeing as Evil development has slowed to a crawl I doubt there would be any eyes willing to review it since it is a monolithic change.

> evil--jumps-jumping-backward is necessary because evil--jump-backward shouldn't add a new jump point so any solution that automatically detects buffer-switching commands needs something like it.

Do you not also need to do something for the forward direction? Jumping forward can also change the buffer. With your current code wouldn't a new jump point be set when it shouldn't be?

ehahn:
> However I think it would be better if we had a global variable to keep track of point before a command since there is only one command being executed at any given time. Having a slot in the `evil-jump-struct` implies that its some kind of per-window state.

It is a per-window state. Originally, I used two global variables to store point (with the current buffer) and the selected window before the command. The latter was necessary to avoid setting a jump point when switching between windows. If the buffer changed but the selected window didn't, I set a jump point. This approach failed whenever a command involved the minibuffer. Example:

1. Initially, window `A` shows buffer `X` and is selected.
2. `C-x b`: now the minibuffer window is selected.
3. Select buffer `Y` and confirm. Now `A` shows `Y` . We should set the point saved before step 2 as a jump point but 1. the selected window changed so we don't know that and 2. `previous-pos` is now in the minibuffer anyway.

> Yes `point` can jump to some unexpected place during evaluation of a command, e.g. if a command calls `sleep-for` or `accept-process-output` which would allow timers or process filter functions to run. So there is a possibility to store `point` in `pre-command-hook` and have it move before setting a jump point.

I see. This could be an issue when during the evaluation a command, `point` changes in the initial buffer and a different buffer is selected. There are two possible scenarios. 

1. A background task unrelated to the command (e.g. `comint` waiting for and handling a response from the process) moved `point`. The command just allowed whatever background tasks happen to be active to run.
2. The command intentionally both moved `point` and changed the buffer. It could even do that by letting a timer or process filter coupled to it run.

There is no way to tell them apart so hopefully we want the same behavior in both cases.

In scenario 1, we can ignore the `point` move and pretend it only happened after the current buffer changed. That means we use the old `point` stored in `pre-command-hook`. The user generally cannot tell the difference anyway.

They may be able to tell the difference if the display is updated while the command is being evaluated and they actually see `point` move. Then we have to choose between the `point` stored in `pre-command-hook` and the last `point` the user saw. I think the former is fine.

We can ignore scenario 2 IMO. We cannot achieve expected behavior here anyway because moving `point` and changing the buffer is unexpected behavior to begin with.



> I have an entire re-implementation of `evil-jumps.el` that I've been using for over a year and half which implements the jumplist as a window parameter and fixes a lot of issues with the current implementation as I mentioned in #1138 and related.
> 
> I just need to update the tests and add some more for my branch to be pull request worthy. But I'm hesitant to work on it and submit a pull request. Seeing as Evil development has slowed to a crawl I doubt there would be any eyes willing to review it since it is a monolithic change.

I don't think the size of the change is the problem here. Smaller changes don't get merged either and it doesn't take a detailed review to see that your branch is a big improvement to the current implementation.

> > evil--jumps-jumping-backward is necessary because evil--jump-backward shouldn't add a new jump point so any solution that automatically detects buffer-switching commands needs something like it.
> 
> Do you not also need to do something for the forward direction? Jumping forward can also change the buffer. With your current code wouldn't a new jump point be set when it shouldn't be?

Probably. I don't understand the jumping code well enough to be sure. Also I don't know how to test this without understanding the code since jumping forward is buggy anyway. Using `:jumps` shouldn't set jump points either AFAIK. I think the best course of action is to rebase my changes on top of your branch.

evil-jumps.el
```diff
@@ -272,46 +281,76 @@ POS defaults to point."
...
+
+                 ;; The first time you jump backward, setting a jump
+                 ;; point is sometimes correct. But we don't do it
+                 ;; here because this function is called after
+                 ;; `evil--jump-backward' has updated our position in
+                 ;; the jump list so, again, `evil-set-jump' would
+                 ;; break `evil--jump-forward'.
+                 (not jumping-backward)
+                 (let ((previous-buffer (marker-buffer previous-pos)))
+                   (and previous-buffer
```
dzop: I think jump points can also be revived from the corresponding files right, i.e. jump points are also `(PATH POS)` pairs in addition to markers? So we should also save the `buffer-file-name` of the `previous-pos` and, if its non-nil, use `find-buffer-visiting` to determine if a buffer showing the file exists. If it does, then use it. Otherwise, when there is no buffer, save the file path along with the previous position to the jump list.

I'm unsure if the current code can also set a `(PATH POS)` pair as a jump point and use it in that way. I think the `(PATH POS)` pair is only used when loading a jumplist from the file.

kiennq:
Is there any chance this can be merged?
Having `evil-mode` cannot jump reliably with `find-file` is really annoying.

TheBB:
Thanks for the bump. Rebased, made a few fixes and merged. Sorry for the wait.

tomtomjhj:
@mpereira I've been using @nnicandro's [reimplementation of jumps](https://github.com/emacs-evil/evil/issues/1138#issuecomment-469895242) mentioned above for quite a while and I'm pretty satisfied with it.

kiennq:
Friendly ping

PR #1154, opened 2019-05-21
Fix behavior of evil-scroll-up and evil-scroll-down

### LINKED ISSUE #713: in eww, evil-scroll-down gives (wrong-type-argument integer-or-marker-p nil)
Originally reported by: **Kevin Brubeck Unhammer (Bitbucket: [unhammer](https://bitbucket.org/unhammer), GitHub: [unhammer](https://github.com/unhammer))**

Using Emacs  "25.1.50.2" in X with evil  "1.2.12" (20160827.1510 from melpa), if I do M-s M-w to search for something with eww, the evil-scroll-down function on C-d gives 

```
#!emacs-lisp

Debugger entered--Lisp error: (wrong-type-argument integer-or-marker-p nil)
  evil-scroll-down(nil)
  funcall-interactively(evil-scroll-down nil)
  call-interactively(evil-scroll-down nil nil)
  command-execute(evil-scroll-down)
```
but still scrolls down. In other major modes, it works fine.

- Bitbucket: https://bitbucket.org/lyro/evil/issue/713

### LINKED ISSUE #965: Evil scroll issue when set the header line
- Issue Type: scroll command
- Bug report
#### Description :octocat:
Evil scroll up and down not by the same count when the header line is set.

Here is the code for header-line set.
```emacs lisp
(which-function-mode)

(defun zt/set-header-line ()
  (interactive)
  (which-function-mode t)
  (setq header-line-format
        '((which-func-mode ("" which-func-format " ")))))
(add-hook 'prog-mode-hook 'zt/set-header-line)
(add-hook 'text-mode-hook 'zt/set-header-line)
```
#### Reproduction guide :beetle:
- Start Emacs
- For a Emacs window with 40 lines, open a file and use `C-u`, it will scroll up 18 lines. Use `C-d` it will scroll down 20 lines.

*Expected behaviour:* :heart: :smile:
It should scoll the same amount lines for up and down.

#### System Info :computer:
- OS: darwin
- Emacs: 25.3.3
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 36c15b7)
- Graphic display: t
- Distribution: spacemacs-base
- Editing style: hybrid
- Completion: ivy
- Layers:
```elisp
((spacemacs-completion :packages
                       (not default-helm-config ido-vertical-mode))
 spacemacs-layouts
 (spacemacs-editing :packages
                    (not lorem-ipsum clean-aindent-mode smartparens))
 (spacemacs-editing-visual :packages
                           (not volatile-highlights rainbow-delimiters highlight-indentation lorem-ipsum))
 (spacemacs-evil :packages
                 (not evil-args evil-mc evil-ediff evil-exchange evil-unimpaired evil-indent-plus vi-tilde-fringe evil-lisp-state))
 spacemacs-misc
 (spacemacs-navigation :packages
                       (not flx-ido smooth-scrolling))
 spacemacs-org spacemacs-visual
 (pandoc :variables org-pandoc-options
         '((standalone . t)))
 (dash :packages
       (not counsel-dash helm-dash))
 better-defaults colors prodigy search-engine graphviz
 (ivy :packages
      (not helm-make counsel-projectile))
 (syntax-checking :variables syntax-checking-enable-by-default nil syntax-checking-enable-tooltips nil)
 (spell-checking :packages
                 (not auto-dictionary flyspell-correct-helm)
                 :variables spell-checking-enable-by-default nil)
 (auto-completion :packages
                  (not ac-ispell auto-complete helm-c-yasnippet company-quickhelp)
                  :variables auto-completion-enable-sort-by-usage t auto-completion-enable-snippets-in-popup t)
 (vinegar :variables vinegar-reuse-dired-buffer t)
 (spacemacs-layouts :variables layouts-enable-autosave nil layouts-autosave-delay 300)
 (git :packages
      (not magit-gitflow orgit smeargle)
      :variables git-magit-status-fullscreen t magit-push-always-verify nil magit-save-repository-buffers 'dontask magit-revert-buffers 'silent magit-refs-show-commit-count 'all magit-revision-show-gravatars nil)
 (ibuffer :variables ibuffer-group-buffers-by 'projects)
 (osx :variables osx-dictionary-dictionary-choice "Simplified Chinese - English" osx-command-as 'super)
 restclient
 (shell :variables shell-default-shell 'ansi-term)
 deft
 (chrome :packages
         (not edit-server gmail-message-mode))
 imenu-list
 (treemacs :variables treemacs-use-filewatch-mode t :variables treemacs-use-follow-mode t)
 react
 (markdown :packages
           (not gh-md))
 (org :packages
      (not org-projectile org-download org-present)
      :variables org-want-todo-bindings t)
 gpu yaml php
 (python :variables python-test-runner
         '(nose pytest))
 html
 (javascript :packages
             (not skewer-mode livid-mode))
 (typescript :variables typescript-fmt-on-save nil typescript-fmt-tool 'typescript-formatter)
 emacs-lisp
 (c-c++ :packages
        (not realgud clang-format disaster)
        :variables c-c++-default-mode-for-headers 'c++-mode)
 ztlevi
 (chinese :packages youdao-dictionary :variables chinese-enable-youdao-dict t))
```
- System configuration features: NOTIFY ACL LIBXML2 ZLIB TOOLKIT_SCROLL_BARS

### LINKED ISSUE #1063: Error when scrolling down from the first line of a buffer with a header line
# Issue type

- Bug report

# Environment

Emacs version: GNU Emacs 26.1 (build 1, x86_64-apple-darwin17.5.0, Carbon Version 158 AppKit 1561.4)
Operating System: macOS Sierra
Evil version: evil-git-5508b95
Evil installation type: el-get
Graphical/Terminal: graphical

# Reproduction steps
- Start Emacs
- Evaluate `(setq header-line-format "foo")`
- Press `C-w C-s 10 C-w _ i <RET> <ESC> 10 . gg C-d`

# Expected behavior
Scrolling down with `C-d` should not emit an error.

# Actual behavior
A wrong type argument error is emitted.

# Further notes
The problem stems from the use of the `posn-*` functions in `evil-scroll-down`. When there is a header line for the current window and `point` is at the first line of the window, the y position returned by `posn-x-y` will be 0. The later call to `posn-at-x-y` will return a header line `posn`, i.e. a `posn` with a `posn-area` equal to `header-line`, but a header line `posn` has a nil value for `posn-point` and `evil-scroll-down` assumes `posn-point` to be non-nil.

I was able to work around the error by replacing

```emacs-lisp
(goto-char (posn-point p))
```
in `evil-scroll-down` with

```emacs-lisp
(if (eq (posn-area p) 'header-line)
    (move-to-window-line 0)
  (goto-char (posn-point p)))
```

bet4it:
When header line is enabled, `evil-scroll-up` and `evil-scroll-down` won't remain in the same position when scroll.
```
From my investigation, the root cause was as follows.

On Emacs 24 with non-nil header-line-format, y offset returned by 
`posn-at-point' is relative to the text area not including the header line.  
However, `posn-at-x-y' takes y offset relative to the buffer area including the 
header line.

Before Emacs 24, both of `posn-at-point' and `posn-at-x-y' are relative to the 
buffer area including the header line.
```

Original comment by `yukishi...@google.com` on 1 Nov 2013 at 3:06
- Changed state: **Started**

_Originally posted by @GoogleCodeExporter in https://github.com/google/mozc/issues/196#issuecomment-95066744_

Fix #713. Fix #965. Fix #1063. Close #1080.

### DIFF
--- a/evil-common.el
+++ b/evil-common.el
@@ -867,10 +867,6 @@ Inhibits echo area messages, mode line updates and cursor changes."
   `(let ((evil-no-display t))
      ,@body))
 
-(defun evil-num-visible-lines ()
-  "Returns the number of currently visible lines."
-  (- (window-height) 1))
-
 (defvar evil-cached-header-line-height nil
   "Cached height of the header line.")
 
--- a/evil-common.el
+++ b/evil-common.el
@@ -881,18 +881,15 @@ If there is no header line, return nil."
   "Return the x and y coordinates in POSITION.
 
 This function returns y offset from the top of the buffer area including
-the header line.  This definition could be changed in future.
-Note: On Emacs 22 and 23, y offset, returned by `posn-at-point' and taken
-by `posn-at-x-y', is relative to the top of the buffer area including
 the header line.
-However, on Emacs 24, y offset returned by `posn-at-point' is relative to
-the text area excluding the header line, while y offset taken by
-`posn-at-x-y' is relative to the buffer area including the header line.
+on Emacs 24 and later versions, y offset returned by `posn-at-point' is
+relative to the text area excluding the header line, while y offset taken
+by `posn-at-x-y' is relative to the buffer area including the header line.
 This asymmetry is by design according to GNU Emacs team.
-This function fixes the asymmetry between them on Emacs 24 and later versions.
-Borrowed from mozc.el."
+This function fixes the asymmetry between them.
+Learned from mozc.el."
   (let ((xy (posn-x-y position)))
-    (when (and (> emacs-major-version 24) header-line-format)
+    (when header-line-format
       (setcdr xy (+ (cdr xy)
                     (or evil-cached-header-line-height
                         (setq evil-cached-header-line-height (evil-header-line-height))
--- a/evil-common.el
+++ b/evil-common.el
@@ -871,6 +871,38 @@ Inhibits echo area messages, mode line updates and cursor changes."
   "Returns the number of currently visible lines."
   (- (window-height) 1))
 
+(defvar evil-cached-header-line-height nil
+  "Cached height of the header line.")
+
+(defun evil-header-line-height ()
+  "Return the height of the header line.
+If there is no header line, return nil."
+  (let ((posn (posn-at-x-y 0 0)))
+    (when (eq (posn-area posn) 'header-line)
+      (cdr (posn-object-width-height posn)))))
+
+(defun evil-posn-x-y (position)
+  "Return the x and y coordinates in POSITION.
+
+This function returns y offset from the top of the buffer area including
+the header line.  This definition could be changed in future.
+Note: On Emacs 22 and 23, y offset, returned by `posn-at-point' and taken
+by `posn-at-x-y', is relative to the top of the buffer area including
+the header line.
+However, on Emacs 24, y offset returned by `posn-at-point' is relative to
+the text area excluding the header line, while y offset taken by
+`posn-at-x-y' is relative to the buffer area including the header line.
+This asymmetry is by design according to GNU Emacs team.
+This function fixes the asymmetry between them on Emacs 24 and later versions.
+Borrowed from mozc.el."
+  (let ((xy (posn-x-y position)))
+    (when (and (> emacs-major-version 24) header-line-format)
+      (setcdr xy (+ (cdr xy)
+                    (or evil-cached-header-line-height
+                        (setq evil-cached-header-line-height (evil-header-line-height))
+                        0))))
+    xy))
+
 (defun evil-count-lines (beg end)
   "Return absolute line-number-difference betweeen `beg` and `end`.
 This should give the same results no matter where on the line `beg`
--- a/evil-common.el
+++ b/evil-common.el
@@ -879,14 +879,15 @@ If there is no header line, return nil."
 
 (defun evil-posn-x-y (position)
   "Return the x and y coordinates in POSITION.
-
 This function returns y offset from the top of the buffer area including
 the header line.
-on Emacs 24 and later versions, y offset returned by `posn-at-point' is
-relative to the text area excluding the header line, while y offset taken
-by `posn-at-x-y' is relative to the buffer area including the header line.
-This asymmetry is by design according to GNU Emacs team.
-This function fixes the asymmetry between them.
+
+On Emacs 24 and later versions, the y-offset returned by
+`posn-at-point' is relative to the text area excluding the header
+line, while y offset taken by `posn-at-x-y' is relative to the buffer
+area including the header line.  This asymmetry is by design according
+to GNU Emacs team.  This function fixes the asymmetry between them.
+
 Learned from mozc.el."
   (let ((xy (posn-x-y position)))
     (when header-line-format
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -956,7 +956,7 @@ If the scroll count is zero the command scrolls half the screen."
       (signal 'beginning-of-buffer nil))
     (when (zerop count)
       (setq count (/ (1- (window-height)) 2)))
-    (let ((xy (posn-x-y (posn-at-point))))
+    (let ((xy (evil-posn-x-y (posn-at-point))))
       (condition-case nil
           (progn
             (scroll-down count)
@@ -984,7 +984,7 @@ If the scroll count is zero the command scrolls half the screen."
     ;; In that case we do not scroll but merely move point.
     (if (<= (point-max) (window-end))
         (with-no-warnings (next-line count nil))
-      (let ((xy (posn-x-y (posn-at-point))))
+      (let ((xy (evil-posn-x-y (posn-at-point))))
         (condition-case nil
             (progn
               (scroll-up count)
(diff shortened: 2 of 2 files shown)


evil-common.el
```diff
@@ -864,9 +864,37 @@ Inhibits echo area messages, mode line updates and cursor changes."
...
+  "Cached height of the header line.")
+
+(defun evil-header-line-height ()
+  "Return the height of the header line.
+If there is no header line, return nil."
+  (let ((posn (posn-at-x-y 0 0)))
+    (when (eq (posn-area posn) 'header-line)
+      (cdr (posn-object-width-height posn)))))
+
+(defun evil-posn-x-y (position)
```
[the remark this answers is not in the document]
bet4it: Can you give an example about this?

duianto:
Confirmed. This solves the issue of scrolling down `C-d` from the first line in a 
`magit-revision: .emacs.d` buffer.

Currently it shows:
>let*: Wrong type argument: integer-or-marker-p, nil

System info:
Evil version 1.2.14
Magit 20190708.2154, Git 2.21.0.windows.1, Emacs 26.2, windows-nt
GNU Emacs 26.2 (build 1, x86_64-w64-mingw32) of 2019-04-13

walseb:
It also fixes it for me, but if Emacs is substituting UTF chars with UTF chars from another font than the one you are using this sort of problem also appears, although this is unrelated to this patch. I'm using Emacs 26.1, 26.2 and 27.50 on nixos with latest evil-mode

evil-common.el
```diff
@@ -864,9 +864,37 @@ Inhibits echo area messages, mode line updates and cursor changes."
...
+by `posn-at-x-y', is relative to the top of the buffer area including
+the header line.
+However, on Emacs 24, y offset returned by `posn-at-point' is relative to
+the text area excluding the header line, while y offset taken by
+`posn-at-x-y' is relative to the buffer area including the header line.
+This asymmetry is by design according to GNU Emacs team.
+This function fixes the asymmetry between them on Emacs 24 and later versions.
+Borrowed from mozc.el."
+  (let ((xy (posn-x-y position)))
+    (when (and (> emacs-major-version 24) header-line-format)
```
dzop: Why do do we need this `emacs-major-version` check if the minimum supported version of Emacs is 24.1?

https://github.com/emacs-evil/evil/blob/297b8f323754d967beeaed28080311f5afbe25c8/evil-pkg.el#L5

dzop:
So it seems the difference between your solution and my attempt (#1080) is that you take into account the header line by offsetting the y coordinate of `posn-x-y` with the header line height. This way the call to `posn-at-x-y` always returns a position that is accessible by `point`. This makes more sense.

evil-common.el
bet4it: You are right. I learned it from
https://github.com/emacs-evil/evil/blob/297b8f323754d967beeaed28080311f5afbe25c8/evil-commands.el#L3749-L3751
but support for Emacs < 24 was dropped in 2cad47110dd4d1e47fe5691c11566e4fcd2f658d.

I have removed these code.

evil-common.el
```diff
@@ -864,9 +864,37 @@ Inhibits echo area messages, mode line updates and cursor changes."
...
+  (let ((posn (posn-at-x-y 0 0)))
+    (when (eq (posn-area posn) 'header-line)
+      (cdr (posn-object-width-height posn)))))
+
+(defun evil-posn-x-y (position)
+  "Return the x and y coordinates in POSITION.
+
+This function returns y offset from the top of the buffer area including
+the header line.  This definition could be changed in future.
+Note: On Emacs 22 and 23, y offset, returned by `posn-at-point' and taken
```
dzop: Looks like we don't need this note since Emacs 24 is the minimum supported version.

bet4it: Removed

TheBB:
Thanks!

evil-common.el
```diff
@@ -867,9 +867,35 @@ Inhibits echo area messages, mode line updates and cursor changes."
   `(let ((evil-no-display t))
      ,@body))
 
-(defun evil-num-visible-lines ()
-  "Returns the number of currently visible lines."
-  (- (window-height) 1))
+(defvar evil-cached-header-line-height nil
+  "Cached height of the header line.")
+
+(defun evil-header-line-height ()
```
sorawee: @bet4it It looks like this function doesn't give a reliable result, leading to the bug that this PR is supposed to fix.

To elaborate, whenever I enable [centaur-tabs](https://github.com/ema2159/centaur-tabs), which uses the header line to show tabs, I still encounter the bug on `evil-scroll-up` and `evil-scroll-down`. The header line by `centaur-tabs`  looks really tall compared to the plain header line, and yet `evil-header-line-height` always returns the same value.

Screen Shot 2019-12-31 at 09 21 05

Figure 1: the header line by `centaur-tabs`, along with the value of `evil-header-line-height`, `window-height`., and `window-body-height`.

Screen Shot 2019-12-31 at 09 20 39

Figure 2: the regular header line, along with the value of `evil-header-line-height`, `window-height`., and `window-body-height`.

sorawee: If I use `window-header-line-height` instead, the bug will be properly fixed. Will submit a PR.

PR #1161, opened 2019-06-17
evil-goto-definition only search in current buffer

redguardtoo:
`gd` actually searches local definition only, quoted from vim manual,

```
  Goto local Declaration.  When the cursor is on a local
 variable, this command will jump to its declaration.
First Vim searches for the start of the current
function, just like "[[".  If it is not found the
search stops in line 1.  If it is found, Vim goes back
until a blank line is found.  From this position Vim
searches for the keyword under the cursor, like with
"*", but lines that look like a comment are ignored
(see 'comments' option).
```

I also provides `evil-goto-definition-fucntion` which user could customize the goto definition behavior.

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -234,19 +234,6 @@ move COUNT - 1 screen lines downward first."
                        -1
                        (/ (with-no-warnings (window-body-width)) 2)))))
 
-(evil-define-motion evil-beginning-of-line-or-digit-argument ()
-  "Move the cursor to the beginning of the current line.
-This function passes its command to `digit-argument' (usually a 0)
-if it is not the first event."
-  :type exclusive
-  (cond
-   (current-prefix-arg
-    (setq this-command #'digit-argument)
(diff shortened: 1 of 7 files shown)


redguardtoo:
Don't know why some test is failing. I run the unit test with `24.4`, `25.2`, `26.1` locally. All passed.

TheBB:
I'm happy to introduce a variable with an optional function that we can call as a first choice, but I don't like this implementation. It looks entirely specialized for javascript, and I don't want to invite hundreds of PRs for other languages with complicated implementations that I need to check. Seems cleaner to leave it *nil* as a default value (and, correspondingly, guard against *nil* in the calling code). Then js-whatever-mode can bind it on activation if necessary, as can other modes.

redguardtoo:
@TheBB , 

 `evil-goto-definition-function` could be `nil` or `evil-default-goto-definition-function` could support more languages.

I don't have preference, I could delete `evil-default-goto-definition-function`.

PR #1166, opened 2019-07-01
Fix :global behavior with trailing slashes

### LINKED ISSUE #1165: :global doesn't correctly handle trailing slashes
# Issue type

- Bug report

# Environment

Emacs version: GNU Emacs 26.2 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.8) of 2019-04-12

Operating System: Manjaro
Evil version: 1.2.14
Evil installation type: `straight.el`
Graphical/Terminal: X11

# Reproduction steps

- Start Emacs with `make emacs`.
- Execute `:g/buffer/s/i/TEST/` on the scratch buffer with the default scratch text. Note the trailing slash.

# Expected behavior

The buffer text should change from:

```
;; This buffer is for text that is not saved, and for Lisp evaluation.
;; To create a file, visit it with C-x C-f and enter text in its buffer.
```

to the following, which was generated by calling `:1,2s/i/TEST/`:

```
;; ThTESTs buffer is for text that is not saved, and for Lisp evaluation.
;; To create a fTESTle, visit it with C-x C-f and enter text in its buffer.
```

# Actual behavior

The text becomes:

```
;; This  is for text that is not saved, and for Lisp evaluation.
;; To create a file, visit it with C-x C-f and enter text in its .
```

# Further notes

Calling `(evil-delimited-arguments "/buffer/s/i/TEST/" 2)` (which is what `evil-ex-parse-global` does) gives me `'("buffer" "s")`, so it seems that the `/i/TEST/` part gets dropped due to the `evil-delimited-arguments` call in `evil-ex-parse-global`.

mamapanda:
`evil-delimited-arguments` might drop characters from the chained command string if the input ends with a trailing slash (so in something like `:g/pattern/s/pat/repl/`, the `/pat/repl/` part gets dropped). Since the chained command follows after the slash following the pattern, we can just take a substring instead of relying on `evil-delimited-arguments` for it.

Fixes #1165.

### DIFF
--- a/evil-search.el
+++ b/evil-search.el
@@ -1210,9 +1210,10 @@ This handler highlights the pattern of the current substitution."
 
 (defun evil-ex-parse-global (string)
   "Parse STRING as a global argument."
-  (let* ((args (evil-delimited-arguments string 2))
-         (pattern (nth 0 args))
-         (command (nth 1 args)))
+  (let* ((pattern (nth 0 (evil-delimited-arguments string 2)))
+         (command (and pattern
+                       (>= (- (length string) (length pattern)) 2)
+                       (substring string (+ (length pattern) 2)))))
     ;; use last pattern if none given
     (when (zerop (length pattern))
       (setq pattern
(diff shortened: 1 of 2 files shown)


mamapanda:
I got no clue why Travis Emacs is failing to load libgnutls.so.30.

TheBB:
Thanks! Merged manually by cherry picking.

PR #1171, opened 2019-07-21
Teach more commands about evil-respect-visual-line-mode

justbur:
The following commands now act on visual lines when
`evil-respect-visual-line-mode` is non-nil:

`evil-insert-line`
`evil-append-line`
`evil-find-char`
`evil-find-char-backward`
`evil-find-char-to`
`evil-find-char-to-backward`


TheBB:
This is nice but apparently the find-char test fails.

justbur:
@TheBB Thanks. I fixed the issue.

TheBB:
Thanks!

PR #1186, opened 2019-08-13
Delete window on q if created to hold evil-list-view-mode buffer

kqr:
Windows created specifically to hold evil-list-view-mode buffers seem a lot like
other temporary windows in Emacs, e.g. the ones created by describe-function.

These tend to define q to mean quit-window, rather than kill-this-buffer; The
quit-window command has the nice property that if a window was created to hold
the buffer in question, that window will be deleted – but if an existing window
was repurposed to hold the buffer, the effect is the same as bury-buffer.

Note that this is also in line with the default behaviour of Vim, where pressing
q in the marks list restores the previous window layout.

### DIFF
--- a/evil-common.el
+++ b/evil-common.el
@@ -3845,7 +3845,7 @@ should be left-aligned for left justification."
            (entry (elt tabulated-list-entries (1- line))))
       (funcall evil-list-view-select-action (nth 1 entry)))))
 
-(define-key evil-list-view-mode-map (kbd "q") #'kill-this-buffer)
+(define-key evil-list-view-mode-map (kbd "q") #'quit-window)
 (define-key evil-list-view-mode-map [follow-link] nil) ;; allows mouse-1 to be activated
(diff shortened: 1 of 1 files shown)


kqr:
I have tried understanding why the build fails, but I don't quite see it. I'd appreciate any pointer.

TheBB:
Don't worry about CI, let it complain.

kqr:
Is there any chance of getting feedback on this?

dzop:
+1

I have something similar in my own local copy of `evil`. It makes using the registers much easier to deal with, e.g. when I forget what text is stored in which register I often do `M-x evil-show-registers` (bound to `M-'`), take a look, and then `q`.

I would actually suggest to use something like `(quit-window 'kill)` since, as you said, all the `evil-list-view-mode` buffers are transient and their content out of date after doing any kind of text editing. Burying them just leaves a buffer with outdated information lying around.

TheBB:
Thanks! I'll take @dzop's advice and kill the buffer too.

PR #1187, opened 2019-08-19
Fix evil-get-marker switching buffers

mamapanda:
At the moment, if the last jump marks are in a different buffer, then calling ``(evil-get-marker ?`)`` or `(evil-get-marker ?')` will switch the current buffer. This PR adds a `save-window-excursion` call to fix that.

You can reproduce the behavior by doing the following:
- Run `make emacs`.
- Jump to the top of the scratch buffer with `gg`.
- Switch to the messages buffer with `C-x b *Messages* [RET]`.
- Evaluate `(evil-get-marker ?')` with `M-:`. The buffer should change.

### DIFF
--- a/evil-common.el
+++ b/evil-common.el
@@ -1952,8 +1952,9 @@ or a marker object pointing nowhere."
         (when (and (symbolp marker) (boundp marker))
           (setq marker (symbol-value marker)))
         (when (functionp marker)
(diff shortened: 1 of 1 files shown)


TheBB:
Thanks for this. Seems to solve the problem and tests pass. Merged manually.

PR #1196, opened 2019-10-31
Remove bundled ert.el

TheBB:
This was needed to test on Emacs <24, for which support now has been dropped. The bundled ert.el was too outdated for testing on Emacs >26.


TheBB:
Travis has problems with 26 and above, but things pass locally on all versions so I'll deal with that some other time.

PR #1197, opened 2019-11-02
Add keybindings and ex commands for tabs

endoffile78:


TheBB:
Thanks! I merged this manually as it needed some fixes:

- I fixed `:tabp` to call previous instead of next.
- I also removed `:tabe` as, bound to `tab-bar-new-tab` it doesn't really do what it does in vim. An implementation with `find-file-other-tab` is more suitable.
- Added a guard so that these functions aren't bound on older Emacsen.

PR #1201, opened 2019-11-21
Fix evil motion 0 binding command

zilvinasu:
With the recent code changes, motion `0` was broken, because the command name used is not the one that is defined.


TheBB:
Thanks! I took the liberty of changing the name the other way around, since the corresponding non-visual-line command is called `evil-digit-argument-or-evil-beginning-of-line`.

zilvinasu:
Np, thanks for quick response!

PR #1205, opened 2019-11-27
Fix for void variable user-error - run-hook-with-args-until-success: Symbol’s value as variable is void: user-error `

jayka:
Getting error in Evil package  MELPA (build# 20191125.1525) or Checkout from Github `master` branch (`56eb22ac71f10a5475945086ebb9ed45af0a6638`) 

Steps to reproduce:
- While in `evil-normal-state`, goto a symbol definition by typing `gd`  ( bound to `evil-goto-definition` by default)
- Getting error `run-hook-with-args-until-success: Symbol’s value as variable is void: user-error
`
Fix:
This patch addresses void variable - `user-error`


TheBB:
Superseded by #1207

PR #1212, opened 2019-12-01
Filter bad register names

sanel:
Some modes, like fzf.el, will put :fzf-windows as register name,
causing Evil :register command to fail. This command, but also Vim,
expects registers to be named with number or character.

This is not much fault of fzf.el mode (or similar modes), because
Emacs keeps both window configuration and marks in a single
"register-alist" and mode developers are free to use whetever name for
restoring window configurations.

This commit fixes it by filtering out all register names not conforming to
proper Vim-like naming schema.

### DIFF
--- a/evil-common.el
+++ b/evil-common.el
@@ -2143,12 +2143,15 @@ register instead of replacing its content."
     (set-register register text))))
 
 (defun evil-register-list ()
-  "Returns an alist of all registers"
+  "Returns an alist of all registers, but only those named
+with number or character. Registers with symbol or string in names are ignored
+to keep Vim compatibility with register jumps."
   (sort (append (mapcar #'(lambda (reg)
                             (cons reg (evil-get-register reg t)))
                         '(?\" ?* ?+ ?% ?# ?/ ?: ?. ?-
                               ?1 ?2 ?3 ?4 ?5 ?6 ?7 ?8 ?9))
-                register-alist nil)
+                (cl-remove-if-not (lambda (x) (number-or-marker-p x)) register-alist)
+                nil)
         #'(lambda (reg1 reg2) (< (car reg1) (car reg2)))))
 
 (defsubst evil-kbd-macro-suppress-motion-error ()
--- a/evil-common.el
+++ b/evil-common.el
@@ -28,6 +28,7 @@
 (require 'evil-digraphs)
 (require 'rect)
 (require 'thingatpt)
+(require 'cl-lib)
 
 ;;; Code:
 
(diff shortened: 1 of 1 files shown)


evil-common.el
```diff
@@ -2143,12 +2143,15 @@ register instead of replacing its content."
...
-  "Returns an alist of all registers"
+  "Returns an alist of all registers, but only those named
+with number or character. Registers with symbol or string in names are ignored
+to keep Vim compatibility with register jumps."
   (sort (append (mapcar #'(lambda (reg)
                             (cons reg (evil-get-register reg t)))
                         '(?\" ?* ?+ ?% ?# ?/ ?: ?. ?-
                               ?1 ?2 ?3 ?4 ?5 ?6 ?7 ?8 ?9))
-                register-alist nil)
+                (cl-remove-if-not #'(lambda (reg) (number-or-marker-p reg)) register-alist)
```
TheBB: What's the difference between `(lambda (reg) (number-or-marker-p reg))` and `number-or-marker-p`? Why not just use the latter?

TheBB: Also, evil supports Emacs 24.1, which does not have cl-lib bundled. :-/

TheBB: Maybe you can do `(evil-filter-list (lambda (reg) (not (number-or-marker-p (car reg)))) (copy-list register-alist))` or something similar.

sanel: Thanks for catching this! I was working in few iterations, so `(lambda (reg) ...)` left. Will fix that.

Regarding `cl-lib` issue, isn't already used in Evil, e.g. `evil-jumps.el`? `cl-remove-if-not` is already used there; are you planning to remove these calls?

TheBB: Ah, turns out I'm actually stupid. Evil already depends on `cl-lib`. You'll just need to add a `(require 'cl-lib)` on top of `evil-common.el` and it'll be fine.

sanel: Done :)

TheBB:
Thanks! Squashed and merged manually.

- I changed the filtering function, since the list is an alist and you want to filter on the keys, not the cons cells.
- I also removed the note in the commit message that fzf.el is not at fault. The documentation of `register-alist` says that they keys must be numbers, so if some package adds non-numbers there then as far as I am concerned it is their problem. However I guess this filtering doesn't hurt.

TheBB:
Reported to fzf at https://github.com/bling/fzf.el/issues/54

sanel:
Thanks!

Issue #1074, opened 2018-07-21
Remove undo-tree dependency


mrbig033:
# Issue

Undo tree is an undo package that cannot keep undos.

# Environment

Emacs version: Emacs 26.1.50
Operating System: Manjaro (current)
Evil version: Evil version 1.2.13
Evil installation type: MELP
Graphical/Terminal: Graphical
Tested in a `make emacs` session (see [CONTRIBUTING.md](https://github.com/emacs-evil/evil/blob/master/CONTRIBUTING.md#bug-reports)):  Yes
Undo configurations:
```
(use-package undo-tree
  :ensure t
  :init
  (setq undo-limit 78643200)
  (setq undo-outer-limit 104857600)
  (setq undo-strong-limit 157286400)
  (setq undo-tree-mode-lighter " UN")
  (setq undo-tree-auto-save-history t)
  (setq undo-tree-enable-undo-in-region nil)
  (setq undo-tree-history-directory-alist '(("." . "~/emacs.d/undo")))
 (add-hook 'undo-tree-visualizer-mode-hook (lambda ()
                                              (undo-tree-visualizer-selection-mode)
                                              (setq display-line-numbers nil)))
  :config
  (global-undo-tree-mode 1))
```
# Reproduction steps
- Start Emacs
- Make 30 changes on a files an save each step
- Restart Emacs
- Try undoing and redoing those changes

# Expected behavior

You should be able to undo and/or redo every single change

# Actual behavior

When it works, I'm able to undo 5 changes at the most. Half the time I can't undo anything at all. Sometimes I get the message `unrecognized entry in undo list undo-tree-canary` while I'm performing the undo.

# Further notes

It is not just me:

https://old.reddit.com/r/emacs/comments/85t95p/undo_tree_unrecognized_entry_in_undo_list/
https://debbugs.gnu.org/cgi/bugreport.cgi?bug=16523
https://debbugs.gnu.org/cgi/bugreport.cgi?bug=16377
https://lists.ourproject.org/pipermail/implementations-list/2014-November/002091.html
https://github.com/syl20bnr/spacemacs/issues/298
https://github.com/syl20bnr/spacemacs/issues/9903

I also must add that I had this same problem before with entirely different configurations, other versions of Emacs and operating systems.

npostavs:
> Make 30 changes on a files an save each step

Is there some specific set of changes needed for this? I tried with the below and got no errors (this doesn't load evil, but as far as I understand the bug is not related to evil-mode).

[...]

Issue #522, opened 2015-05-26
When saving the buffer, change list got emptied.


TheBB:
Originally reported by: **ReneFroger (Bitbucket: [ReneFroger](https://bitbucket.org/ReneFroger), GitHub: [ReneFroger](https://github.com/ReneFroger))**

----------------------------------------

I use Emacs 24.4 in GUI-mode

From the Vim / Evil documentation: 

         g;          Go to [count] older position in change list.

         g,          Go to [count] newer cursor position in change list.

When I save the buffer inside Vim, and then press `g;`, it will go to the latest changed postion. 

In Evil, after saving a buffer, I press `g;`, which calls the function `(goto-last-change ARG)`. But I got the error message that there is no further change info. 

----------------------------------------
- Bitbucket: https://bitbucket.org/lyro/evil/issue/522

ReneFroger:
Is this bug still open? It was surprisingly to see my comments back here from somewhere, thanks for that!

wasamasa:
Why, yes, it is.

TheBB:
Undo-tree is no longer default after #1360, so closing.

Issue #418, opened 2014-09-14
undo-tree-undo doesn't work properly with a region selected


TheBB:
Originally reported by: **Lars Andersen (Bitbucket: [expez](https://bitbucket.org/expez), GitHub: [expez](https://github.com/expez))**

----------------------------------------

I recommend this binding while exploring: `(define-key evil-visual-state-map "u" 'undo-tree-undo)`

From `make emacs`:

In the scratch buffer add:
```
foo bar baz
foo bar baz
```

Change the first `bar` to `bdr`, with `r`.
hit `V` and then `try undo-tree-undo`.  It won't change the text.

Furthermore, calling `undo-tree-undo` in situations where nothing happens will add no-op edit points to the undo-tree list and cancel the selected region.

`undo-tree-undo` works some of the time with a region selected, but I haven't been able to determine exactly when it fails.

----------------------------------------
- Bitbucket: https://bitbucket.org/lyro/evil/issue/418

TheBB:
Closing as outdated. Undo-tree is now optional after #1360.

PR #1229, opened 2019-12-22
Add option to switch to undo-fu

TheBB:
Just a draft branch that I'll be playing around with for a while. Opening this as a heads up.

Hopefully this can end up closing #1074, #418 and #522.

### DIFF
--- a/evil-integration.el
+++ b/evil-integration.el
@@ -139,55 +139,55 @@
             (when (overlayp ov) (delete-overlay ov))))))))
 
 ;;; Undo tree
-(when (and (require 'undo-tree nil t)
-           (fboundp 'global-undo-tree-mode))
-  (global-undo-tree-mode 1))
-
-(eval-after-load 'undo-tree
-  '(with-no-warnings
-     (defun evil-turn-on-undo-tree-mode ()
-       "Enable `undo-tree-mode' if evil is enabled.
+(when (eq 'undo-tree evil-undo-provider)
+  (when (and (require 'undo-tree nil t)
+             (fboundp 'global-undo-tree-mode))
+    (global-undo-tree-mode 1))
+  (eval-after-load 'undo-tree
+    '(with-no-warnings
+       (defun evil-turn-on-undo-tree-mode ()
+         "Enable `undo-tree-mode' if evil is enabled.
 This function enables `undo-tree-mode' when Evil is activated in
 some buffer, but only if `global-undo-tree-mode' is also
 activated."
-       (when (and (boundp 'global-undo-tree-mode)
-                  global-undo-tree-mode)
-         (turn-on-undo-tree-mode)))
+         (when (and (boundp 'global-undo-tree-mode)
+                    global-undo-tree-mode)
+           (turn-on-undo-tree-mode)))
 
-     (add-hook 'evil-local-mode-hook #'evil-turn-on-undo-tree-mode)
+       (add-hook 'evil-local-mode-hook #'evil-turn-on-undo-tree-mode)
 
-     (defadvice undo-tree-visualize (after evil activate)
-       "Initialize Evil in the visualization buffer."
-       (when evil-local-mode
-         (evil-initialize-state)))
+       (defadvice undo-tree-visualize (after evil activate)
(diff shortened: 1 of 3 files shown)


edkolev:
Nice initiative!

A couple of notes:
- let me know if you need need me to dogfood this feature - I'd be happy to test
- supporting both undo-tree and undo-fu might be a stretch (in terms of implementation cost, and maintenance cost). I suggest supporting just one one them

TheBB:
Yeah I'd love for others to test.

Wasn't sure yet whether I wanted to rip out undo-tree completely or what. If testing goes well we can take that into account. One problem is that undo-fu has very strict Emacs version requirement. I'll check with the author if that's just because it's untested on older versions.

TheBB:
Also *goto-chg* depends on undo-tree but I think it's a soft dependency.

TheBB:
There's been a new release to undo-tree, which might mitigate some of the problems people have been seeing. I'll leave this PR in limbo for a while.

jixiuf:
I would got  on emacs28 
```
Symbol's function definition is void: undo-tree-save-history-from-hook
```
I think  this hooks should be added after function is defined 
```
(add-hook 'write-file-functions #'undo-tree-save-history-from-hook)
(add-hook 'kill-buffer-hook #'undo-tree-save-history-from-hook)
(add-hook 'find-file-hook #'undo-tree-load-history-from-hook)
```

can not reach the author of undo-tree

aaronjensen:
@jixiuf he responds to email, which you can find on his website: http://www.dr-qubit.org/undo-tree.html

That said, I think he is unlikely to make any changes for Emacs28 unless they are totally innocuous for current emacs versions.

The error you're seeing is surprising... they're changing the semantics of `#'symbol` or `add-hook` is doing validation now...

TheBB:
I already reported that to him. I got a fixed version by email but nothing
on git or ELPA yet. If someone else wants to bump it be my guest!

TheBB:
Undo-tree 0.7.3 is now out and it should have fixed the above mentioned issue.

ideasman42:
Any update on this?

Even in the case undo-tree is has been fixed, it's a fairly large and complex package, it installs hooks even when it's disabled.
It would be nice if there was a way to disable undo-tree in evil mode.

```
(add-hook 'post-gc-hook #'undo-tree-post-gc nil)
;; --- snip --- 
(add-hook 'write-file-functions #'undo-tree-save-history-from-hook)
(add-hook 'kill-buffer-hook #'undo-tree-save-history-from-hook)
(add-hook 'find-file-hook #'undo-tree-load-history-from-hook)
 ```
----

Undo-fu by comparison:

- doesn't install any hooks/advice/timers.
- it's 230-SLOC compared to 2740-SLOC.
- no GC / undo-data manipulation.
- no extra mode/state needed It just exposes undo and redo which operate on Emacs undo information.

**Comment by TheBB** [comment]
No updates yet. I wrote this to try undo-fu but at that time there were still bugs. Since then undo-tree received a promising update, so I decided I could focus on other things for the moment.

I'll get around to it eventually, but of course if someone else wants to implement it I'd be happy to review and merge.

I'm pretty sure I actually do want the provider variable though. :smile:

**Comment by TheBB** [comment]
Note: we also need to double check that goto-chg works fine when undo-tree isn't present. It's supposed to, but since evil for the moment requires undo-tree I'm not sure how much those code paths have been tested.

**Comment by ideasman42** [comment]
Testing `goto-last-change` and `goto-last-change-reverse`, they both seem to be working properly.

**Comment by condy0919** [comment]
Any updates?

**Comment by TheBB** [comment]
If there's significant interest in an option for choosing undo provider please feel free to clean up the PR and resubmit. I don't have much time for evil at the moment. I've received no reports of bugs in the newer undo-tree versions so I don't consider it a priority.

**Comment by wasamasa** [comment]
@jixiuf @aaronjensen Do you have a minimal reproduction example for that change in hook functions? I've tried the following on Emacs built from master and failed to reproduce it with the following in a file:

```elisp
(defun my-test () (message "Test"))
(add-hook 'post-command-hook 'my-test)
```

TheBB:
Closed in favor of #1360

PR #1235, opened 2019-12-31
Honor isearch filter predicates

### LINKED ISSUE #1233: Searching for a word with only one occurrence with "*" no longer highlights that word
After upgrading to the latest version of eveil, I noticed when a word has only one occurrence in a buffer, searching it with "*" would not highlight it. For example see the screenshot below when I try to search for the word "this":

**Before**

Screen Shot 2019-12-30 at 8 43 41 PM

**After**

The word "this" is no longer highlighted.

---------------------------------

After using git-bisect it seems PR https://github.com/emacs-evil/evil/pull/1221 is the one that introduced this behavior. Is this a bug or expected?

TheBB:
This PR reverts #1221 and introduces another fix for #843 without the regression of #1233.

The source of #843 is that isearch calls our search function, which may find a match that is filtered out, usually because the text is invisible. In that case isearch calls the search function again, until it either returns nil (no match found) or finds a match that isn't filtered.

If we wrap the search around the buffer endpoints and all matches are filtered out, this will cause an endless loop experienced by the user as Emacs freezing.

#1221 fixed this by modifying the bounds of the wrapped search, preventing it from finding the same match endlessly, but this introduces its own problems: some matches that should be found are now no longer found. This PR reverts that fix.

The original problem is fixed by manually checking `isearch-filter-predicate` against all our matches, forcibly returning nil if we can't find a satisfactory match rather than returning the match and letting isearch check it. This allows isearch to abort as it should.

Fixes #1233

### DIFF
--- a/evil-search.el
+++ b/evil-search.el
@@ -191,16 +191,10 @@ of the buffer."
            result)
        (setq result (funcall search-fun string bound
                              ,(if wrap t 'noerror) count))
-       ;; Wrap the search only if a result was not found, and a bound not set
-       (when (and ,wrap (null result) (null bound))
+       (when (and ,wrap (null result))
          (goto-char ,(if forward '(point-min) '(point-max)))
          (unwind-protect
-             ;; The wrapped search is bounded by the original starting point
-             (setq result (funcall search-fun string
-                                   ,(if forward
-                                        '(max (point-min) (1- start))
-                                      '(min (point-max) (1+ start)))
-                                   noerror count))
+             (setq result (funcall search-fun string bound noerror count))
            (unless result
              (goto-char start))))
        result)))
--- a/evil-search.el
+++ b/evil-search.el
@@ -257,13 +257,6 @@ one more than the current position."
          (goto-char orig)
          (user-error "\"%s\": %s not found"
                      string (if regexp-p "pattern" "string"))))
-      ;; handle opening and closing of invisible area
-      (cond
-       ((boundp 'isearch-filter-predicates)
-        (dolist (pred isearch-filter-predicates)
-          (funcall pred (match-beginning 0) (match-end 0))))
-       ((boundp 'isearch-filter-predicate)
-        (funcall isearch-filter-predicate (match-beginning 0) (match-end 0))))
       ;; always position point at the beginning of the match
       (goto-char (match-beginning 0))
       ;; determine message for echo area
--- a/evil-search.el
+++ b/evil-search.el
@@ -173,12 +173,30 @@ Disable anyway if FORCE is t."
   (remove-hook 'evil-operator-state-exit-hook #'evil-flash-hook t))
 (put 'evil-flash-hook 'permanent-local-hook t)
 
-(defun evil-search-function (&optional forward regexp-p wrap)
+(defun evil-search-with-predicate (search-fun pred string bound noerror count)
+  "Execute a search with a predicate function.
+SEARCH-FUN is a search function (e.g. `re-search-forward') and
+PREDICATE is a two-argument function satisfying the interface of
+`isearch-filter-predicate', or `nil'.  STRING, BOUND, NOERROR and
+COUNT are passed unchanged to SEARCH-FUN.  The first match
+satisfying the predicate (or `nil') is returned."
+  (catch 'done
+    (while t
+      (let ((result (funcall search-fun string bound noerror count)))
+        (cond
+         ((not result) (throw 'done nil))
+         ((not pred) (throw 'done result))
+         ((funcall pred (match-beginning 0) (match-end 0)) (throw 'done result)))))))
+
+(defun evil-search-function (&optional forward regexp-p wrap predicate)
   "Return a search function.
 If FORWARD is nil, search backward, otherwise forward.
 If REGEXP-P is non-nil, the input is a regular expression.
 If WRAP is non-nil, the search wraps around the top or bottom
-of the buffer."
+of the buffer.
+If PREDICATE is non-nil, it must be a function accepting two
+arguments: the bounds of a match, returning non-nil if that match is
+acceptable."
   `(lambda (string &optional bound noerror count)
      (let ((start (point))
            (search-fun ',(if regexp-p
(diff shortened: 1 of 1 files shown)


Alexander-Shukaev:
I only have one comment/question.  There already some related code at

https://github.com/emacs-evil/evil/blob/59675bff8d99b192ab1b96e36089c861aab40919/evil-search.el#L262

Not sure how your new change interoperates or interferes with both this whole `cond` in general and `isearch-filter-predicates` in particular.

TheBB:
I was unsure about that too, but it seems like calling `isearch-filter-predicate` has the double effect of opening folded sections of the buffer, e.g. under a hidden org-mode title. The purpose of that `cond` is to trigger that side-effect. It looks like with this change we can remove that.

I haven't found any reference to `isearch-filter-predicates` (plural) in Emacs 27 or 24.1 so I don't know what that is about. Blame points to https://github.com/emacs-evil/evil/commit/300a.

TheBB:
Actually it looks like the introduction of `isearch-filter-predicates` was later reverted.

- https://debbugs.gnu.org/cgi/bugreport.cgi?bug=14714
- https://git.savannah.gnu.org/cgit/emacs.git/commit/?id=dc6c0edad789a8aab129ae12b8f064d0e912cca7
- https://git.savannah.gnu.org/cgit/emacs.git/commit/?id=66fc57e3cdc9ffc2bed7a457a39a210fcf773e98

I don't think this ever made it into a release so we can probably ignore `isearch-filter-predicates`

Alexander-Shukaev:
I did the following test.  Make some dummy changes in some `git` repository in a way that I have several same words on different lines to search for.  Then open `magit-status` buffer and make sure that the `unstaged` section is closed, where all those new words reside in a diff.  If then I repeatedly do `isearch-forward` with `search-invisible` being `t` and input that new word, I see the point correctly jump into the section (without unfolding it) to the correct line.  However, neither `evil-search-next` nor `evil-search-forward` jump to the correct lines, while `evil-search-next` seems to even skip some matches inside folded section.  This may not be related to the current issue/fix though.

TheBB:
I'll have to double check that. I've just been using org-mode to help me understand.

TheBB:
In Evil master, search will find matches in hidden sections but they won't be expanded. In this PR, Evil won't find those matches at all. Basic Emacs isearch (without Evil) has the same behaviour as this PR does: it does not find matches in hidden sections.

Alexander-Shukaev:
That's only because `search-invisible` is `open` by default.  If you change it to `t`, the difference should be there.

TheBB:
If I set `search-invisible` to `t`, in both master and in this PR the evil/isearch behaviour seems to be the same: the match under a folded section is found but not expanded. Am I missing something?

TheBB:
Wait, I was searching for commit messages, not diff contents. Does this only happen with diff contents? (Sorry for always writing two comments in a row... haha.)

Alexander-Shukaev:
Fine, let's take something more deterministic:

```
* Hello

hello world!
hello world!
hello world!
------------

* Bye

Bye world!
Bye world!
Bye world!
----------

```

Change to `outline-mode` and `evil-toggle-fold` on both chapters to close them.  First, `isearch` test:

1.  Go to beginning of buffer;
2.  `isearch-forward` with `world` pattern 7 times, nothing is unfolded but the positions are:
    -   `3:11`, <kbd>return</kbd>, `3:11`;
    -   `4:11`, <kbd>return</kbd>, `4:11`;
    -   `5:11`, <kbd>return</kbd>, `5:11`;
    -   `10:9`, <kbd>return</kbd>, `10:9`;
    -   `11:9`, <kbd>return</kbd>, `11:9`;
    -   `12:9`, <kbd>return</kbd>, `12:9`;
    -   no more matches.

Second, `evil-search-forward` test:

1.  Go to beginning of buffer;
2.  `evil-search-forward` with `world` pattern 1 time:
    -   `3:11`, <kbd>return</kbd>, `1:7` (issue 1).

Second, `evil-search-next` test:

1.  Go to beginning of buffer;
2.  `evil-search-next` with `world` pattern 5 times:
    -   `1:7` (issue 2).
    -   `7:0` (issue 2).
    -   `8:5` (issue 2).
    -   `14:0` (issue 2).
    -   wrap around bottom;
    -   `7:0` rather than `1:7` (issue 3).

TheBB:
Thanks for that.

So the problem with this, based on what I can tell, is that isearch puts the cursor at the end of the match, while we want it to be at the beginning. There's some code for this in evil see, [here](https://github.com/emacs-evil/evil/blob/master/evil-search.el#L94-L110).

Note that the bounds of the matches in your example are

    16 - 21
    29 - 34
    32 - 47
    74 - 79
    85 - 90
    96 - 101

If you prepare the buffer as described and close the folds, and execute `(goto-char 16)` you'll find yourself instead at position 8, which is line 1 column 7, corresponding to one of the incorrect positions you have in your comment there.

Apparently `goto-char` will not let you put point inside a hidden portion of the buffer, but somehow isearch manages to do it. 

I tired also with `re-search-forward` and `search-forward` and the cursor ends up at the end of the match but inside the hidden portion. However a second or two later some timer seems to have triggered and the cursor moves outside again.

Tested with (for convenience)

```elisp
(add-to-list 'mode-line-format " %l:%c")
(add-to-list 'mode-line-format '(:eval (format "%S" (point))))
```

However I see the same behaviour in master so I believe this is a different issue and doesn't affect this PR.

TheBB:
Okay, I think I got it, after digging through the isearch source code.

In Emacs <25, you can let-bind `inhibit-point-motion-hooks` to `t` around `goto-char`.

In Emacs 25 or later you have to use `cursor-sensor-inhibit` in this way:

    (make-local-variable 'cursor-sensor-inhibit)
    (push 'evil cursor-sensor-inhibit)
    (goto-char ...)
    (delq 'evil cursor-sensor-inhibit)

See docstring for that variable for this slightly awkward usage. Using this I was successfully able to get point to precisely the location I wanted it.

TheBB:
Well I tried a quick fix but there must be something more to this story because it didn't work as expected.

Suggest just merging this and opening an issue with the above findings.

Alexander-Shukaev:
Wow, great findings!  Sure, I'm fine with the merge, and let's now forget your analysis for awkward point behavior in folds to tackle in future.

TheBB:
Thanks for your help testing this.

PR #1238, opened 2020-01-01
Fix behavior of evil-scroll-up and evil-scroll-down again

sorawee:
The PR #1154 is supposed to fix the issue about evil-scroll-up
and evil-scroll-down. It works nicely in most scenarios.
However, when the header line is really tall,
evil-header-line-height will give an incorrect result,
making the bug that is supposed to be fixed by the mentioned PR
reappear again.

This PR switches evil-header-line-height to Emacs' built-in
window-header-line-height which corrects the mistake.

### DIFF
--- a/evil-common.el
+++ b/evil-common.el
@@ -868,6 +868,19 @@ Inhibits echo area messages, mode line updates and cursor changes."
   `(let ((evil-no-display t))
      ,@body))
 
+(defvar evil-cached-header-line-height nil
+  "Cached height of the header line.")
+
+(defun evil-header-line-height ()
+  "Return the height of the header line.
+If there is no header line, return 0."
+  (if (fboundp 'window-header-line-height)
+      (window-header-line-height)
+      (let ((posn (posn-at-x-y 0 0)))
+        (or (when (eq (posn-area posn) 'header-line)
+              (cdr (posn-object-width-height posn)))
+            0))))
+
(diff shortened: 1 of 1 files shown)


TheBB:
It doesn't look like this function is available in all the Emacs 24 versions that Evil supports.

Could you make the logic conditional on `(fboundp 'window-header-line-height)` perhaps?

sorawee:
And if it's not bound, what should I do? Fall back to the incorrect one?

I'd also wait for @bet4it in case s/he has any better idea about the fix.

TheBB:
> And if it's not bound, what should I do? Fall back to the incorrect one?

That's the idea. There aren't many users on Emacs 24 and there are probably even fewer using a header line that will trigger this bug. I just want to avoid them getting a void-function error.

There are no hard rules for this, but I figure when Emacs 27 and Ubuntu 20.04 are released we can start phasing out Emacs 24 support, so this shim shouldn't have to stay around too long. Then Evil can support the last three major Emacs versions and the last two Ubuntu LTSes, whichever is smallest.

TheBB:
I pushed a third commit that changes the order of the logic. Now the caching only applies to the deprecated fallback implementation and not the Emacs function, which is implemented in C and I can only assume that means it's reasonably quick. If that's not the case we can adjust later.

PR #1239, opened 2020-01-01
Skip command window tests when run in batch mode

edkolev:
This PR does the following:
- tests related to the command window are skipped when emacs is run in batch mode

The cmd window tests always fail with something like "error resizing window". With this change these tests will be skipped in batch mode.

As to why I run evil's tests in batch mode - long story short, I'm switching [evil-goggles](https://github.com/edkolev/evil-goggles) to [makel](https://gitea.petton.fr/DamienCassou/makel). In addition to evil-goggles's own tests, I run the whole evil-tests.el suite with the evil-goggles mode on. Since makel runs the tests in batch mode, the evil tests suite always fails for me.

### DIFF
--- a/evil-tests.el
+++ b/evil-tests.el
@@ -7602,6 +7602,7 @@ maybe we need one line more with some text\n")
 
 (ert-deftest evil-test-command-window-ex ()
   "Test command line window for ex commands"
+  (skip-unless (not noninteractive))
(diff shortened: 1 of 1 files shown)


TheBB:
:+1:

This isn't the first time I wished there were a `(skip-when ...)`.

Issue #1147, opened 2019-05-13
Call evil-command-window-ex with keybinding C-f in evil-ex


socksy:
# Issue type
- Enhancement request

# Reproduction steps
- Start Emacs
- Press `:`
- press C-f

# Expected behavior
`(evil-command-window-ex)` is run (same as `q:` in normal mode)

# Actual behavior
Says "end-of-buffer"

# Further notes
As complained about here: https://news.ycombinator.com/item?id=19899436

I would open a PR, but my elisp skills are basically non-existent.

xabbu42:
C-f in the minibuffer is not exactly the same as q:, as it also adds the current line as the last command. This is very useful if you are writing a complex command and need/want a full featured buffer for editing that command. Below is my current implementation of vim Ctrl-f (keep in mind that I neither know vim nor emacs all to well and this code was mostly generated by trial and error):

`(defun evil-ctrl-f-execute (config result)
  (select-window (active-minibuffer-window) t)
  (set-window-configuration config)
  (delete-minibuffer-contents)
  (insert result)
  (exit-minibuffer)
)

(defun evil-ex-ctrl-f ()
  "Vim ctrl-f in command-mode"
  (interactive)
  (let ((current (minibuffer-contents))
		(config  (current-window-configuration)))
	(evil-ex-teardown)
	(select-window (minibuffer-selected-window) t)
	(evil-command-window (cons current evil-ex-history)
						 ":"
						 (apply-partially 'evil-ctrl-f-execute config))))

(defun evil-search-ctrl-f ()
  "Vim ctrl-f in search"
  (interactive)
  (let ((current (minibuffer-contents))
		(config  (current-window-configuration)))
	(select-window (minibuffer-selected-window) t)
	(evil-command-window (cons current evil-ex-search-history)
						 (evil-search-prompt (eq evil-ex-search-direction 'forward))
						 (apply-partially 'evil-ctrl-f-execute config))))

(evil-select-search-module 'evil-search-module 'evil-search)
(define-key evil-ex-completion-map "\C-f" 'evil-ex-ctrl-f)
(define-key evil-ex-search-keymap "\C-f" 'evil-search-ctrl-f)
(setq enable-recursive-minibuffers t)
`

xabbu42:
[...]

PR #1240, opened 2020-01-05
Implement Vim Ctrl-f in minibuffer.

xabbu42:
Open apropriate command windows for Ctrl-f while in minibuffer for an ex command
or a search. As all ways to abort a minibuffer do not return and restore the
active buffer, we have to set the content of the minibuffer and exit normally to
trigger the apropriate action. Also restoring the window configuration is
necessary so minibuffer-selected-window is not changed.

This pull request should resolve issue #1147


TheBB:
Thanks!

Next time please make sure not do indent with tabs in elisp code.

PR #1248, opened 2020-01-13
Enable undo-tree-mode only when evil-mode is turned on

TheBB:
This brings us closer to the ideal that loading files should be side-effect-free. The evil-integration file is a sinner.

This moves the undo-tree-mode activation from top-level to being conditioned on the enabling of evil-mode.

Will make #1229 nicer.


TheBB:
Also took the liberty of updating the bundled undo-tree that we use for testing to the newest version.

PR #1251, opened 2020-01-18
Add support for submenu entries in goto-definition-imenu

CeleritasCelery:
Currently `evil-goto-definition-imenu` will only look in top level entries and ignores everything in a submenu. Unfortunately this means that many definitions that could be found are being skipped. This PR fixes that by looking in submenus as well.


TheBB:
Thanks for this.

The way I read the imenu documentation, it is possible to get nested sublists of arbitrary level. It looks like your code only looks one level deep. Should we turn this into a proper recursive lookup?

TheBB:
There's also the possibility of leveraging `(imenu--in-alist STR ALIST)` or at least its logic.

CeleritasCelery:
> There's also the possibility of leveraging (imenu--in-alist STR ALIST)

Good find! That is exactly what we need. It gets nested elements for us.

TheBB:
This was merged but something went wrong with github --- I'm not sure why it didn't go purple. Closing manually.

PR #1255, opened 2020-01-25
Ex search highlights: use selected window as fallback

TheBB:
In some cases search is triggered not from the minibuffer, in which case `minibuffer-selected-window` will not accurately capture the window in which search highlights should be created.


TheBB:
@hlissner I think this should fix your issue.

PR #1259, opened 2020-01-29
Change the way to get universal argument

bet4it:
It's better to unify the way between `evil-paste-*` and `yank*` to handle with universal argument, so spacemacs can do extra hook to them.
Refer: syl20bnr/spacemacs#13161

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -1927,7 +1927,8 @@ The default for width is the value of `fill-column'."
   "Pastes the latest yanked text before the cursor position.
 The return value is the yanked text."
   :suppress-operator t
-  (interactive "p<x>")
+  (interactive "*P<x>")
+  (setq count (prefix-numeric-value count))
   (if (evil-visual-state-p)
       (evil-visual-paste count register)
     (evil-with-undo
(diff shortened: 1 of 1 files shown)


TheBB:
I don't mind doing this exactly, but could you clarify how exactly spacemacs uses `evil-paste-*` that this will help?

TheBB:
I'm asking because as far as I can tell this change is inconsequential, and these two functions will behave the same:

```elisp
(defun a (count)
  (interactive "p")
  count)

(defun b (count)
  (interactive "P")
  (prefix-numeric-value count))
```

bet4it:
spacemacs use `advice-add evil-paste-* :around` and need to intercept the universal argument to do extra check.

TheBB:
To be honest, my first instinct is that Spacemacs wants to wrap the paste functions I think it would make more sense to rebind `p` and `P` to new functions that can interpret the universal argument as they wish, and then call the evil functions.

The fact that the argument used to be in raw form I would consider as an implementation detail that it'd be unwise to rely upon. After all it's called 'count'.

bet4it:
Yes, you are right, but `(interactive "P<x>")` has existed for nearly ten years so it's better to keep it in this way😀

TheBB:
Fair enough.

Issue #1260, opened 2020-01-29
[Feature request] ability to append to default register 


yyoncho:
# Issue type
- Enhancement request

ATM you could copy-append doing the following sequence
- `"tyy` -> copy current line in register `t`
- `"Tyy` -> copy-append current line in register `t`
- `"tp` -> paste `t` register

I propose adding the ability to append to default register `0` (not sure if this is the correct term). So the sequence from above will become:

- `yy` -> copy line
- `")yy` -> copy-append line
- `p` -> paste the two lines

11 chars vs 7 chars. 

Note: If this is rejected due to the fact that this is not what vim do I would propose a slight restructuring of the code to allow implementing this as an extension.

TheBB:
This will probably be rejected for that exact reason, yeah. Deviations from vim behaviour should be justified based on difficulty of implementation relative to existing solutions available in Emacs. We don't generally deviate just for the sake of deviating, even for QoL improvements.

I'd be happy to accept a PR that allows you to implement this externally if it is not possible with current evil.

leoalekseyev:
If anybody stumbles upon this issue, note that the advice in the referenced commit will work for register 0 only, as the code indicates. In general, be aware that evil has special handling of registers `"` and `0-9` (among others), where it may manipulate or read from the kill ring instead of a register. This causes the provided advice to work as intended for register `0` only.

PR #1263, opened 2020-01-30
Extract evil-append-register function

yyoncho:
- Fixes https://github.com/emacs-evil/evil/issues/1260

```elisp
(advice-add 'evil-set-register :around
           (lambda (orig register text)
             (if (eq register ?\) )
                 (evil-append-register ?0 text)
               (funcall orig register text))))
```


TheBB:
Excellent. Thanks!

PR #1269, opened 2020-02-08
Makefile: add 'info' target

jonEbird:
Useful to have a specific info target when some Emacs package installers
expect this target (such as el-get).


TheBB:
Thanks! Although surely there must be plenty of packages that don't have makefiles, let alone info targets.

PR #1279, opened 2020-02-20
Correct typo (COMPOUNT->COMPOUND) in symbol name

bassettmb:
Correct typo (COMPOUNT->COMPOUND) in symbol name in window server selection data-type list.
Should correspond with the symbol used [in the emacs source code](https://github.com/emacs-mirror/emacs/blob/master/lisp/select.el#L306) and [in the emacs ipc encoding documentation](https://www.gnu.org/software/emacs/manual/html_node/emacs/Communication-Coding.html).


TheBB:
Thanks!

PR #1282, opened 2020-03-01
Don't pass &rest to ignore in evil-define-command

mamapanda:
At the moment, having `&rest` in a command's argument list results in the following error when running the command:
```
Symbol's value as variable is void: &rest
```


TheBB:
Thanks!

PR #1283, opened 2020-03-03
fix `wrong number of argument error` in evil-ex-command-window

dsdshcym:
- evil-ex-command-window only accepts 0 arguments
- and it doesn't make sense to call itself recursively
- this issue was introduced in [1: 48ceb43]

1: 48ceb43f8a1e95e6d7eafe428ff640d83952462f
   Clean up compilation warnings from lexical binding


TheBB:
Thanks!

Issue #1295, opened 2020-04-11
Visual mode put fails from registers > 1


tomdl89:
# Issue type
- Bug report

# Environment
Emacs version: 27.0.50 x86_64-pc-linux-gnu
Operating System: RHEL 7.3 maipo
Evil version: 1.2.14
Evil installation type: MELPA
Graphical/Terminal: Graphical
Tested in a `make emacs` session (see [CONTRIBUTING.md](https://github.com/emacs-evil/evil/blob/master/CONTRIBUTING.md#bug-reports)): No

# Reproduction steps
- Start Emacs
- Write 3 or more lines in a scratch buffer
- Yank each of them (`yy`)
- Do `:reg` to see the lines in reverse order in the buffers 0-9
- Enter visual mode (or visual line or visual block mode) and select some text
- Do `"2p` and observe that nothing happens
- Do `"3p` `"4p` `"5p` etc and observe that the selection is deleted, but nothing is put in its place

# Expected behavior
Putting from registers 2-9 should behave just like registers `0`, `1` and `a`-`z`

# Actual behavior
Putting from register `2` doesn't do anything, and putting from registers `3`-`9` replaces the selected text with nothing (actually it replaces the text with only a new-line if the register chosen contains a new-line)

# Further notes
- Registers `0`, `1` and all of the letter registers behave as expected
- I'm fairly sure this is a different issue to #906

tomdl89:
Fixed in PR #1296

PR #1296, opened 2020-04-12
Fix issue 1295: failed visual-paste from registers >1

tomdl89:
The `kill-ring` let binding in `evil-visual-paste` meant that
`evil-get-register`'s view of `kill-ring` (used in `evil-paste-before`
and `evil-paste-after`) was limited to only entry `0` (and whatever
`evil-delete` may have added to rebound `kill-ring`). Therefore,
registers >1 were not actually visible to functions that needed
them. Condensing the `let*` binding so `kill-ring-yank-pointer` is
bound directly to `(list (current-kill 0))` was chosen as a simple and
effective method to fix the bug.

Fix for issue #1295

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -2041,8 +2041,7 @@ The return value is the yanked text."
          new-kill
          paste-eob)
     (evil-with-undo
-      (let* ((kill-ring (list (current-kill 0)))
-             (kill-ring-yank-pointer kill-ring))
+      (let ((kill-ring-yank-pointer (list (current-kill 0))))
         (when (evil-visual-state-p)
           (evil-visual-rotate 'upper-left)
           ;; if we replace the last buffer line that does not end in a
@@ -2058,13 +2057,12 @@ The return value is the yanked text."
                      (not (= evil-visual-end (point-max))))
             (insert "\n"))
           (evil-normal-state)
-          (setq new-kill (current-kill 0))
           (current-kill 1))
         (if paste-eob
             (evil-paste-after count register)
           (evil-paste-before count register)))
       (when evil-kill-on-visual-paste
-        (kill-new new-kill))
+        (current-kill -1))
       ;; mark the last paste as visual-paste
       (setq evil-last-paste
             (list (nth 0 evil-last-paste)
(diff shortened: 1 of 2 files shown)


tomdl89:
I'm not sure the CI ran properly:
`No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.`

TheBB:
Yeah, the way we're running CI has some problems. If an error occurs during test setup and not test running, then the backtrace won't be caught and Emacs will keep spinning forever.

In this case the problem is that you're missing a closing paren.

tomdl89:
Ah I see - thanks for pointing that out! I didn't have smartparens on at the time. Now I have failing tests, hopefully I can fix.

tomdl89:
Sorry it took me so long. All tests now pass (including one I added). @TheBB is there anything more you need from me? Thanks

tomdl89:
It's there something I can do to help move this forward? Thanks

TheBB:
I was about to merge a few days ago actually. I just had some problems understanding the test you wrote, and I couldn't reproduce it in vim.

Let me try again tomorrow.

tomdl89:
Ah that's awesome, thanks. The test yanks 3 words then uses the `"3` register to visual paste over the first word of the next line. At least that's what I meant it to do! Let me know if you need me to give it another look.

tomdl89:
I just tried replicating this in Neo Vim 0.5.0-427 and I couldn't. Looking at the docs, it seems that vim only populates the number registers >0 with what has been deleted, not yanked (http://vimdoc.sourceforge.net/htmldoc/change.html#quote_number). However, in emacs we `kill-ring-save` when we copy text, it populates the kill-ring when copied _and_ deleted. My change makes the behaviour more consistent with evil's (non-visual-) paste, although I can see that it is not strictly compatible with vim (although the current behaviour of visual paste just wiping the selected text on visual paste is neither compatible with vim nor emacs).

TheBB:
Okay, that's fine. Could you put a comment to that effect in the test? Then I'll merge tomorrow.

TheBB:
Thanks!

PR #1308, opened 2020-05-08
Binding "g C-g" to count words

condy0919:
`g C-g` to count words in motion state

### DIFF
--- a/evil-maps.el
+++ b/evil-maps.el
@@ -336,6 +336,9 @@
 (when evil-want-C-d-scroll
   (define-key evil-motion-state-map (kbd "C-d") 'evil-scroll-down))
 
+(when evil-want-C-g-bindings
+  (define-key evil-motion-state-map "g\C-g" 'count-words))
+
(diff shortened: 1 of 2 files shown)


condy0919:
@TheBB This change is small enough.

TheBB:
Thanks, but conventionally in Emacs, C-g should not be used as the final part of a key binding, because it is used for canceling. See [key binding conventions](http://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html).

When Vim defaults conflict with Emacs conventions we generally err on the side of Emacs conventions and leave a customization variable for people who want to override it (see all the other `evil-want-...` variables). It seems almost silly to do so here because this is such a minor binding, but we can call it `evil-want-C-g-bindings` and use it for all Vim bindings that use C-g, even if there's only one. What do you think?

condy0919:
Fine with me

TheBB:
Thanks!

PR #1313, opened 2020-05-26
Fix make clean and document build

sanel:
When sphinx is not found, `make clean` (and others) will fail. This patch prevents running sphinx if sphinx-build binary is not found.

Using `command` instead of `which` to make it more portable across shells and unixes.


TheBB:
Are those other commits supposed to be present? Can you rebase on current master, picking only what's relevant?

sanel:
Thanks for catching that; those commits are already present in history. I guess my fork was stalled a bit, but I'm not sure why magit attached them on top again... I'll reopen PR again when I fix t his

PR #1314, opened 2020-05-26
Fixed make clean and doc build

sanel:
When sphinx is not found, make clean (and others) will fail. This
change prevents running sphinx if sphinx-build binary is not found.

Here is used 'command' instead of 'which', for checking presence of
binary in PATH, to make it more portable across shells and unixes.


TheBB:
Thanks!

PR #1329, opened 2020-06-24
Add `go` motion command (evil-goto-char)

rettakjak:
This goes along with the ex command `:goto` that already exists in Evil.


condy0919:
> This goes along with the ex command `:goto` that already exists in Evil.

ping @TheBB ?

tomdl89:
Fixed in #1424

PR #1332, opened 2020-07-08
Fix the behavior of `C-w C-c`

rettakjak:
Here's another small PR, this time to fix an inconsistency I've noticed with respect to Vim behavior.

In Evil `C-w C-c` has been closing the current window.  This behavior appears to have been silently introduced in commit ec53465dafc99357161dffc53d49592914d628e5.

Vim, however, doesn't close the window in this case.  From the Vim documentation:

> You might have expected that CTRL-W CTRL-C closes the current
> window, but that does not work, because the CTRL-C cancels the
> command.

Now use `ignore` instead of `evil-window-delete` so nothing happens in this case (specifically, we do keep it defined so we don't get the message "C-w C-c is undefined").

Thanks!

### DIFF
--- a/evil-maps.el
+++ b/evil-maps.el
@@ -146,7 +146,7 @@
 (define-key evil-window-map "=" 'balance-windows)
 (define-key evil-window-map "|" 'evil-window-set-width)
 (define-key evil-window-map "\C-b" 'evil-window-bottom-right)
-(define-key evil-window-map "\C-c" 'evil-window-delete)
+(define-key evil-window-map "\C-c" 'ignore)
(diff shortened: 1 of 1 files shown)


axelf4:
I do not see the point in binding it to `ignore` - should `g C-c` etc. also be bound to `ignore`? If anything my vote is on just removing it.

axelf4:
Actually, reading into it more, the comment in the Vim help file is saying that if it were possible then `C-w C-c` would be bound to closing the window. Therefore, I think the binding should be kept, since we do not suffer from the same restriction, as Emacs does not reserve `C-c` (but instead `C-g`). Closing this for now, please comment if you disagree with my reasoning.

PR #1351, opened 2020-09-04
Fix documentation building.

cperl82:
Currently, cloning master and running:

cd doc && make info

Does nothing for me.

The problem seems to be the use of /usr/bin/[ and a command without a
comparison introduced in #1314.

If you remove the redirection of stderr to /dev/null, you'll see:

$ if [ command -v sphinx-build ]; then echo foo; fi
bash: [: -v: binary operator expected

Simply running the command without /usr/bin/[ has the intended behavior.

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -10,7 +10,7 @@ BUILDDIR      = build
 
 # Put it first so that "make" without argument is like "make help".
 help:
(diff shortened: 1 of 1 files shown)


cperl82:
This change is no longer relevant given the change in 3f3423ec4ca5bb87e49cd6b9ac02228e88aaa42a.  Closing.

Issue #1275, opened 2020-02-18
Compile warning: `undo-elt-crosses-region` is an obsolete function


JeremyDurnell:
The following compile warning is observed upon initial installation with following code in `~/.emacs`:
```
(use-package evil
  :ensure t
  )
```
```
Compiling file /home/jkdurnell/.emacs.d/elpa/evil-20200203.2006/evil.el at Mon Feb 17 19:38:21 2020

Compiling no file at Mon Feb 17 19:38:21 2020
Warning (bytecomp): ‘undo-elt-crosses-region’ is an obsolete function (as of 25.1). [2 times]
```

It seems like it might be the following line in `undo-tree.el`
https://github.com/emacs-evil/evil/blob/d14e9e2cac539f388a1f4699eb3ffab0834aa3e4/lib/undo-tree.el#L2231

TheBB:
Closing as outdated, vendored library removed.

Issue #1144, opened 2019-04-19
undo-tree generating an error when starting Emacs


jpprime:
# Issue type

- Bug report

# Environment

Emacs version: 26.1
Operating System: OpenBSD 6.4
Evil version: 1.2.14
Evil installation type: MELPA stable
Graphical/Terminal: Graphical
Tested in a `make emacs` session (see [CONTRIBUTING.md](https://github.com/emacs-evil/evil/blob/master/CONTRIBUTING.md#bug-reports)): Yes

# Reproduction steps
- Set `debug-on-error` to `t` before loading Emacs
- Set `evil-want-integration` to `t` before loading Emacs or activate
  `evil-collection` with `(evil-collection-init)` in the init file;
- Start Emacs

# Expected behavior
- Emacs starts with `evil-mode` and `undo-tree` without any error.

# Actual behavior
- Emacs starts with `evil-mode` and `undo-tree`, but it issues the following
  error (because `debug-on-error` is active):

```
Debugger entered--Lisp error: (wrong-type-argument undo-tree nil)
  signal(wrong-type-argument (undo-tree nil))
  undo-tree-current(nil)
  (undo-tree-node-previous (undo-tree-current buffer-undo-tree))
  (and undo-tree-mode (not buffer-read-only) (not (eq t buffer-undo-list)) (undo-tree-node-previous (undo-tree-current buffer-undo-tree)))
  redisplay_internal\ \(C\ function\)()
```

This is the error message I get on my normal Emacs installation.

# Further notes
To test this error with `make emacs` I had to make some changes to the `Makefile` and I also created a `initTest.el` with minimal conditions to reproduce the bug.

#### Changes made to the `emacs` call on the `Makefile`:
```sh
# Load Evil in a fresh instance of Emacs and run all tests.
emacs:
	$(EMACS) -Q -L . $(LIBS) -l goto-chg.el -l evil-tests.el \
--eval "(evil-mode 1)" -l ../initTest.el \
--eval "(evil-tests-initialize '(${TAG}) '(${PROFILER}) t)"
```

#### `initTest.el` file (which I placed outside the `evil/` folder):
```lisp
(setq debug-on-error t)

(require 'package)

(setq package-archives
      '(("melpa-stable" . "https://stable.melpa.org/packages/")))

(package-initialize)

(package-install 'evil-collection)
(load "~/numacs/test/evil/evil.el")
(setq evil-want-integration t)
(evil-mode 1)
```

[...]

Issue #601, opened 2015-12-19
Redo cursor position is broken.


TheBB:
Originally reported by: **Anonymous**

----------------------------------------

Redo cursor position should match Vim's.
Vim always puts the cursor at the *start* of the change that was just "redone". I notice that Evil does not do this correctly.d

----------------------------------------
- Bitbucket: https://bitbucket.org/lyro/evil/issue/601

PR #1360, opened 2020-10-01
Make undo-tree an optional dependency

wasamasa:
I'd be thrilled if people tested this. Maybe even if a starter kit like Doom or Spacemacs enabled it. CC @hlissner @syl20bnr


fbergroth:
Using `undo-redo`, I get an error when `evil-redo` calls `(undo-redo nil)`, i.e. when `count` is `nil`. It seems to work if I provide a digit argument.

wasamasa:
@fbergroth Thanks, I've fixed this now.

leungbk:
Thanks for doing this. It's working well for me.

leungbk:
Since `goto-chg.el` is still a dependency, and since `undo-tree` is a dependency of `goto-chg.el`, `evil` still winds up depending on `undo-tree`. From a quick skim of `undo-tree`'s code, it doesn't look like the advice and hooks it sets are particularly intrusive when `global-undo-tree-mode` is turned off, though.

wasamasa:
Hm, that's a good point. The code of goto-chg.el does have a soft dependency on undo-tree.el, but it still does declare a hard dependency on it and errors out if it cannot be loaded up successfully.

TheBB:
Seems to work well for me, using the native Emacs (28+) functions.

evil-commands.el
```diff
@@ -1623,12 +1623,12 @@ of the block."
 
 (evil-define-command evil-undo (count)
   "Undo COUNT changes in buffer using `evil-undo-function'."
-  (interactive "<c>")
+  (interactive "p")
```
fbergroth, line 1626: Perhaps `(interactive "*p")` (error if buffer is read-only)?

wasamasa:
I've found that setting this to undo-tree only works if you've made sure to evaluate `(global-undo-tree-mode)` first. For some reason I haven't been successful with enabling it automatically for users, but eh, I can live with that.

evil-commands.el
wasamasa: Hm, it doesn't seem to ever hit that case sinde read-only buffers tend to not have any undo data in the first place or allowing you to edit them to create it.

fbergroth: Ah ok. A user could potentially enable `read-only-mode` after having made some changes though. I just noticed that `undo-redo` used `"*p"`.

TheBB:
Might be worth pointing that out somewhere, like the docstring of the *evil-undo-system* variable at least.

wasamasa:
Alright, I've found out my issue and pushed a fix. Now you'll get a helpful error message if undo-tree-mode has not been enabled.

evil-commands.el
wasamasa: Fair enough, fixed.

TheBB (approved):

wasamasa:
Before merging this I'd like to try removing the hard dependency of goto-chg.el on undo-tree.el, otherwise the changes to the README claiming that Evil no longer requires undo-tree.el aren't true.

fbergroth (approved):

wasamasa:
I went through the issue tracker and noticed that this PR might end up closing the following issues as well:

- https://github.com/emacs-evil/evil/issues/1275 (undo-tree is no longer vendored)
- https://github.com/emacs-evil/evil/issues/1144 (undo-tree is no longer the default)
- https://github.com/emacs-evil/evil/issues/1074 (the issue that prompted this PR)
- https://github.com/emacs-evil/evil/issues/601 (specific to undo-tree, cosmetic)
- https://github.com/emacs-evil/evil/issues/522 (specific to undo-tree)
- https://github.com/emacs-evil/evil/issues/418 (specific to undo-tree, resolved by using undo-fu)

TheBB:
Good work. Thanks!

JacksonGariety:
I now get the error `Customize 'evil-undo-system' for redo functionality.` when redoing in `evil-mode`. Is this expected to break? What's the suggested way to get redo now?

leungbk:
> I now get the error `Customize 'evil-undo-system' for redo functionality.` when redoing in `evil-mode`. Is this expected to break? What's the suggested way to get redo now?

`(setq evil-undo-system PREFERRED-UNDO-SYSTEM)`

The options are  `'undo-tree` (extension), `'undo-fu` (extension),  and (Emacs 28 only) `'undo-redo`.

TheBB:
Yeah it's expected. Do as the message says and customize that variable.

PR #1369, opened 2020-10-13
Update docs

TheBB:


evil-vars.el
```diff
@@ -2001,7 +2001,11 @@ Customized via `evil-undo-system'.")
     (error "Unknown undo system %s" system))))
 
 (defcustom evil-undo-system nil
-  "Undo system Evil should use."
+  "Undo system Evil should use.  If equal to `undo-tree' or
+`undo-fu', those packages must be installed.  If equal to
+`undo-tree', `undo-tree-mode' must also be activated.  If equal
+to `undo-redo' on Emacs 28 or later, Emacs' own undo system is
```
wasamasa, line 2007: This is factually wrong. By default the vanilla system is used which doesn't have redo in the sense we know it. The `undo-redo` and `undo-only` commands use the same vanilla system, it's possible to backport these commands to earlier Emacs versions than 28 and undo-fu serves as technical demonstration of this fact.

TheBB: Sure, updated.

PR #1377, opened 2020-11-09
Add missing indent declaration to evil-define-key*

daviwil:
While chatting with @angrybacon we realized that `evil-define-key*` was missing the `(declare (indent defun))` which prevents it from being indented the same as `evil-define-key`.  This PR adds the expected declaration to get that indentation behavior.


TheBB:
Thanks!

PR #1397, opened 2020-12-23
Expand FAQ discussion of word vs. symbol text objects

### LINKED ISSUE #1388: Update FAQ entry for “Underscore is not a word character” with a reference to evil-a-symbol
For a long time I have been looking for the place where I learned about the text object map `o` as an alternative to `w`, i.e. as in training the Vim muscle memory to do `cio` instead of `ciw`, to overcome the underscore-is-not-a-word-character “problem”. Today I finally found it again, and that place is [this answer on Stack Overflow](https://stackoverflow.com/a/25067807), which refers to this line in `evil-maps.el`:
https://github.com/emacs-evil/evil/blob/4a336e9c0e0940ad21716260be9bb393c89f6553/evil-maps.el#L307

I am creating this issue as a place to discuss the best way to incorporate this knowledge into [the FAQ entry](https://evil.readthedocs.io/en/stable/faq.html#underscore-is-not-a-word-character), before I open any PR on it.

amiryal:
Add reference to the `o` key in the default text objects keymap, as an alternative to `w` in some cases.

Closes #1388.

### DIFF
--- a/doc/source/faq.rst
+++ b/doc/source/faq.rst
@@ -83,9 +83,21 @@ e.g.:
 
 This gives the underscore the word syntax-class in all C-like buffers.
 
-Alternatively, many find that motion by *symbols* is more convenient
-than motion by *words*.  One way to make word motions operate as
-symbol motions is to alias the ``evil-word`` *thing* [#thingatpt]_ to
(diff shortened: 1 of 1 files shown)


TheBB:
Thanks!

PR #1398, opened 2020-12-23
`evil-ex': `eval-expression' auto-completion

nbfalcon:
Add auto-completion for `evil-ex` LISP expressions leveraging
`elisp-completion-at-point` if available.

When executing an Elisp expression with `:(`, there should be auto-completion for lisp functions instead of a "wrong number of arguments" error. Implement that.

### DIFF
--- a/evil-ex.el
+++ b/evil-ex.el
@@ -229,6 +229,12 @@ Otherwise behaves like `delete-backward-char'."
   (insert result)
   (exit-minibuffer))
 
+(defun evil-ex-elisp-completion-at-point ()
+  "Complete an `evil-ex' Elisp expression."
+  (when (and (fboundp 'elisp-completion-at-point)
+             (string-prefix-p "(" (minibuffer-contents-no-properties)))
+    (elisp-completion-at-point)))
+
 (defun evil-ex-setup ()
   "Initialize Ex minibuffer.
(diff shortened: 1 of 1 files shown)


evil-ex.el
```diff
@@ -229,6 +229,12 @@ Otherwise behaves like `delete-backward-char'."
   (insert result)
   (exit-minibuffer))
 
+(defun evil-ex-elisp-completion-at-point ()
+  "Complete an `evil-ex' Elisp expression."
+  (when (and (fboundp 'elisp-completion-at-point)
+             (string-prefix-p "(" (minibuffer-contents-no-properties)))
```
nbfalcon, line 235: `evil-ex-cmd` is probably the right thing to use here. However, if I replace `minibuffer-contents-no-properties` with it, I get an error ("Args out of range: 0, 0").

evil-ex.el
```diff
@@ -242,7 +248,8 @@ interactive actions during ex state."
   (with-no-warnings
     (make-variable-buffer-local 'completion-at-point-functions))
```
nbfalcon, line 249: As a side note, why `make-variable-buffer-local` instead of `setq-local` or `(set (make-local-variable 'completion-at-point-functions))`?

PR #1405, opened 2021-01-10
Abort repeat with evil-undo

aaronjensen:
Without this, `.` will repeat undo, at least when using undo-tree.

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -1628,12 +1628,14 @@ of the block."
 
 (evil-define-command evil-undo (count)
   "Undo COUNT changes in buffer using `evil-undo-function'."
+  :repeat abort
   (interactive "*p")
   (evil--check-undo-system)
   (funcall evil-undo-function count))
 
 (evil-define-command evil-redo (count)
   "Undo COUNT changes in buffer using `evil-redo-function'."
+  :repeat abort
   (interactive "*p")
   (evil--check-undo-system)
   (funcall evil-redo-function count))


aaronjensen:
Actually, the same needs to apply to evil-redo

aaronjensen:
Updated

aaronjensen:
I should mention, that, at least on my machine, some very bad things (like emacs 100% cpu spins) happen in some situations without this change when undoing/redoing and repeating things.

xsrvmy:
Ran into this on spacemacs. Someone please merge.

tomdl89:
I'm not the maintainer (he's a bit busy right now) but I do have merge permissions. Happy to do so, but feel I need some explanation first:

- Why 'abort rather than 'ignore or nil?
- Is this a recent regression? How has it not come up before?
- It's not a problem for me using undo-fu - is it specific to undo-tree, and if so, why?
- Could we add a test for this?

aaronjensen:
> * Why 'abort rather than 'ignore or nil?

The `:repeat` property of both `undo` and `redo` is `'abort`. These were the commands `u` and and `C-r` were [previously bound to](https://github.com/emacs-evil/evil/pull/1360/files#diff-7357181e821c62f52ab9f57dbd6b2d2012a293ae7cb29298c12da144412cda32L114-L115)

Furthermore, from `evil-repeat.el`:

```
;; Not all commands are recorded. There are several commands that are
;; completely ignored and other commands that even abort the currently
;; active recording, e.g., commands that change the current buffer.
```

Undoing changes the current buffer. It would be odd if one was in the middle of typing in a buffer in insert mode, issued an undo with something like `c-/`, then kept typing if a repeat would insert both the pre-undo and the post-undo typed characters.

> Is this a recent regression? How has it not come up before?

Yes. It was introduced when the configurable undo system was introduced, i.e. when https://github.com/emacs-evil/evil/pull/1360 was merged. 

> It's not a problem for me using undo-fu - is it specific to undo-tree, and if so, why?

When using `undo-tree`, evil-repeat sees the command as `evil-undo` which has a default repeat of `t`. When using `undo-fu`, it sees it as `undo-fu-only-redo`. I do not know why this is, the workings of `this-command` are a bit of a mystery to me.

> Could we add a test for this?

Someone possibly could, but I don't have the time to get into it. Doing the research to answer your other questions was more time than I had to spend on this at the moment, unfortunately.

aaronjensen:
By the way, as an undo-fu user, you may want to try out binding `C-/` to `undo-fu-only-undo` then typing in insert mode:

`aaa` `C-/` `bbb` 

Then leave insert mode and press `.`. See if `aaabbb` is what you really expect to happen there.

tomdl89:
@aaronjensen thanks for taking the time to explain. I had a feeling it would have been a regression from the recent introduction of the configurable undo system. I'm satisfied this should be merged. Please keep an eye out for any unexpected behaviour.

aaronjensen:
Sure thing, thank you for merging!

PR #1413, opened 2021-02-01
Add option global mark goto buffer instead of line

dalanicolai:
This PR is a first proposal to add,  in a first commit, an option to jump to last location in buffer instead of specific line for `evil-mark-goto-line` for global marks.

A second commit additionally adds functionality to show marks in the `*evil-marks*` buffer after some delay (similar to which-key behaviour) when the value of the `evil-mark-goto-buffer-not-line` is `t` (default). This is especially handy to get reminded about which mark is associated with which buffer.

I have implemented it as an option, and because of adding the various conditionals, the code became a little more "complicated".
I would propose to make the buffer jump the default behavior (breaking with vim behavior, although the change could be proposed there also). If users want to jump to a specific line via a global mark they can just jump with backtick (followed by pressing `^`).

I have seen various requests for functionality to easily jump to buffers by shortcut. This (arguably simpler) solution was inspired by @yyoncho his [evil-markers-plus](https://github.com/yyoncho/evil-markers-plus) package.

I am opening this PR as a proposal. If you like it, then I would appreciate some feedback and I would be happy to add documentation for it also.


yyoncho:
Just want to mention that `evil-markers-+` allows using normal global markers and special(which will go to the buffer). IMHO if this PR is accepted it should support that as well.

dalanicolai:
@yyoncho This PR also supports both. It uses the backtick ` for normal global marker functionality, but it "replaces" the goto-mark-line via single quote with the 'go to buffer' variant (jump via quote `'` [currently jumps to first non-blank character of line](https://vim.fandom.com/wiki/Using_marks#Using_marks), which I guess is of low priority for global jumps as there is jump via back tick ` already)

dalanicolai:
I am closing this PR for now because I have found some errors.

Still I think that when making a global jump (i.e. between different buffers), using the `evil-goto-mark-line` could better be used to merely jump to the buffer and then stop (instead of continue to also find the line). In that way, the quote can naturally be used to quickly jump between buffers (with very little implementation effort). It is nice if no extra and less efficient key-sequence has to be found/reserved for that (jumping to the first non blank character of the line can still be achieved by using the backtick jump followed by a `^`).

Also, it might be nice to additionally offer the `which-key`-like behavior that was implemented here in the second commit.

I will probably reopen this PR later, but I'd like to take a little more time to find a cleaner implementation.

PR #1414, opened 2021-02-02
Add option global mark goto buffer instead of line and mark preview

dalanicolai:
This is the updated version of PR #1413, please refer to its PR for motivation.

This PR is mainly created to start the discussion. It has been implemented as an option that is controlled by the variable `evil-mark-goto-buffer-not-line`. 

If you agree it might be a nice feature/option, then I can merge/rewrite the extra functions below into the the original `evil-goto-mark` and `evil-goto-mark-line` functions.

So this PR offers an option that for global markers, and ONLY for global jumps (i.e. to different buffer), the user can choose to not jump to the start of the marked line, but instead just jump to the buffer (i.e. just last cursor position).
Additionally it adds the option to, after initiating a `goto`, see the available marks after some delay in the *evil-marks* buffer. The behavior is equivalent to that of the 'which-key' package.


dalanicolai:
I have added a third commit with the refactored code. I have kept the three commits for clarity. The first commit only adds the option to (global) jump to the last position in buffer instead of the first character of the line.

The second commit adds the `which-key kind of behavior` and some unnecessary logic functions.

The third commit contains the refactored code. I have added only one variable to enable/disable the jump to buffer instead of line AND the delay preview (equivalent to which-key) buffer simultaneously...

PR #1419, opened 2021-02-11
Try to byte-compile evil-delayed forms and conditions

jakanakaevangeli:
Give `evil-delay` a compiler-macro, similar to `eval-after-load`.


axelf4:
Thank you for the PR, it is a shame this was not merged earlier. However, commit 44add36e972ed22ded6cb89654986e80d1d8989e achieved the same results in a slightly more proper way similar to `with-eval-after-load`, which is possible since we may rewrite (virtually) all code that used `evil-delay`.

Issue #1391, opened 2020-12-08
evil-find-file-at-point-with-line not working


ngharrison:
# Issue type
- Bug report

# Environment
Emacs version: GNU Emacs 27.1 (build 1, x86_64-w64-mingw32) of 2020-08-21
Operating System: Windows 10
Evil version: Evil version 1.14.0
Evil installation type: MELPA
Graphical/Terminal: Graphical
Tested in a `make emacs` session (see [CONTRIBUTING.md](https://github.com/emacs-evil/evil/blob/master/CONTRIBUTING.md#bug-reports)): No

# Reproduction steps
- Start Emacs
- Open any buffer
- Create new line that has the name of an existing file on it with a line number, e.g. ~/.emacs.d/init.el:23
- With cursor on file name, call evil-find-file-at-point-with-line

# Expected behavior
The file of the corresponding file name will be opened.

# Actual behavior
It fails and the following error message is shown:
> evil-find-file-at-point-with-line: Wrong number of arguments: #[(filename) "Å ÆÇ	!*È
> !\0
> É!)2\0
> :1\0
> @È!-\0É!))2\0Ê)" [wconfig filename val b x151 current-window-configuration #[nil "Á!" [wconfig set-window-configuration] 2] find-file-at-point bufferp switch-to-buffer-other-window ...] 2 ("c:/Program Files/emacs-27.1-x86_64/share/emacs/27.1/lisp/ffap.elc" . 42373) (list (ffap-prompter nil " other window"))], 0

# Further notes
The same thing happens when no line number is included, e.g. ~/.emacs.d/init.el

usq:
Quickfix: It works for me if I change

https://github.com/emacs-evil/evil/blob/6316dae58e95fe019fcb41d2d1ca5847227a16e6/evil-commands.el#L3397

to

`(with-no-warnings (ffap-other-window fname))`

and reevaluate the function.

ngharrison:
Good find. Works for me too.

PR #1421, opened 2021-02-16
Fix bug in evil-find-file-at-point-with-line

ngharrison:
See issue #1391.


tomdl89:
Looks good. Thanks for the commit!

PR #1424, opened 2021-02-20
Fix evil goto char

tomdl89:
:goto seems to be broken. This is a fix with tests.


rettakjak:
See #1329

tomdl89:
Thanks. Looks like I could close that with this merged.

Issue #360, opened 2013-12-04
Possible evil-search-symbol-forward


TheBB:
Originally reported by: **Anonymous**

----------------------------------------

Recently I've updated evil and function bind to "*" will jump to the next word at point. Before it was jumping to next symbol at point. At least for me the new behavior made the binding "*" useless. For example for emacs-lisp-mode, when place cursor at word "local" in the symbol "evil-local-mode", instead of searching the whole sequence "evil-local-mode" it will only search for part of the name ("local").

I know I can just change the syntax table, but this I would have to do for every mode. Plus changing syntax table seems pretty drastic because many other Emacs packages use it and expect the default definition of a word.

What other solution might be there? Changing syntax table temporary for the duration of the call to evil-search-word-forward? Is that possible?
Maybe adding evil-search-symbol-(backword|forward) functions?

----------------------------------------
- Bitbucket: https://bitbucket.org/lyro/evil/issue/360

PR #1431, opened 2021-03-07
Add info about  `evil-symbol-word-search` to FAQ.

ilyagr:
For me, the Evil default of using words instead of symbols is
only really important for `#` and `*` searches. The fix for them
is much simpler than for the other motions after
https://github.com/emacs-evil/evil/issues/360 (thanks!).

Ideally, these variables would be documented elsewhere
as well. I'm not sure if I'll get around to it.


tomdl89:
@ilyagr Thanks for the PR. Sorry it's taken so long to merge.

PR #1436, opened 2021-03-16
Fix evil-ex-search-backward stuck on first match

### LINKED ISSUE #1435: evil-ex-search-backward stuck on the first match
# Issue type
- Bug report

# Environment

Emacs version: GNU Emacs 28.0.50 (build 6, x86_64-pc-linux-gnu, GTK+ Version 3.24.21, cairo version 1.16.0) of 2021-03-12
Operating System: NixOS 20.09
Evil version: Revision cc9d6886b418389752a0591b9fcb270e83234cf9
Evil installation type: Emacs Doom from GitHub
Graphical/Terminal: Graphical (under Wayland)

# Reproduction steps
- Start Emacs safe mode `emacs -Q`
- Load evil
  ```
  (add-to-list 'load-path "~/.config/emacs/path/to/repos/evil")
  (require 'evil)
  ```
- Write down several words in a buffer for our sandbox search. e.g.
  ```
  (require 'a)
  (require 'b)
  (require 'c)
  ```
- Go to the bottom of buffer.
- Do search backward `evil-ex-search-backward`, match the word `require`.
- Repeat the backward search without moving cursor.

# Expected behavior
It should move from `(require 'c)` then to `(require 'b)` then to `(require 'a)`.

# Actual behavior
It get stuck on `(require 'c)`

# Further notes
The culprit is this goto-char [evil-search.el:1110](https://github.com/emacs-evil/evil/blob/master/evil-search.el#L1110).
It blindly moves the cursor forward without considering the direction.

Abdillah:
Fixes #1435

### DIFF
--- a/evil-search.el
+++ b/evil-search.el
@@ -1107,7 +1107,9 @@ current search result."
                  (goto-char evil-ex-search-start-point)
                  (signal (car err) (cdr err))))))
         ;; pattern entered successful
-        (goto-char (1+ evil-ex-search-start-point))
+        (goto-char (if (eq evil-ex-search-direction 'forward)
+                       (1+ evil-ex-search-start-point)
+                     (- evil-ex-search-start-point 1)))
         (let* ((result
                 (evil-ex-search-full-pattern search-string
                                              evil-ex-search-count
--- a/evil-search.el
+++ b/evil-search.el
@@ -1109,7 +1109,7 @@ current search result."
         ;; pattern entered successful
         (goto-char (if (eq evil-ex-search-direction 'forward)
                        (1+ evil-ex-search-start-point)
-                     (- evil-ex-search-start-point 1)))
+                     (1- evil-ex-search-start-point)))
         (let* ((result
                 (evil-ex-search-full-pattern search-string
                                              evil-ex-search-count


evil-search.el
```diff
@@ -1107,7 +1107,9 @@ current search result."
                  (goto-char evil-ex-search-start-point)
                  (signal (car err) (cdr err))))))
         ;; pattern entered successful
-        (goto-char (1+ evil-ex-search-start-point))
+        (goto-char (if (eq evil-ex-search-direction 'forward)
+                       (1+ evil-ex-search-start-point)
+                     (- evil-ex-search-start-point 1)))
```
tomdl89, line 1112: To be consistent with the line above:
```suggestion
                     (1- evil-ex-search-start-point)))
```

tomdl89 (changes requested):

tomdl89:
Well spotted. I think most people would never find this, because we'd use `evil-ex-search-next` instead of running `evil-ex-search-backward` again. Out of interest, why are you doing it like this?

Abdillah:
> why are you doing it like this?

Well, I was just looking for `forward` and `backward` search functionality and both pop up in suggestion.

evil-search.el
Abdillah: Okay, wait a moment

tomdl89:
The usual way to repeat the same search in the same direction is to press `n`. If you're using `evil-search` you may want to rebind `n` and `N` to `evil-ex-search-next` and `evil-ex-search-previous` respectively.

PR #1438, opened 2021-03-28
Add "gf" to visual mode

rolag:
Adds `evil-find-file-at-point-visual` for finding a file using the
selected visual region. (See `:h v_gf` in Vim)

Adds tests for "gf" in both normal and visual modes.

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -3398,6 +3398,15 @@ If FORCE is non-nil and MARKS is blank, all local marks except 0-9 are removed."
             (forward-line (1- line))))
       (user-error "File does not exist."))))
 
+(evil-define-command evil-find-file-at-point-visual ()
+  "Find the filename selected by the visual region.
+    Returns an error message if the file does not exist."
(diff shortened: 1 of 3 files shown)


evil-commands.el
```diff
@@ -3398,6 +3398,16 @@ If FORCE is non-nil and MARKS is blank, all local marks except 0-9 are removed."
             (forward-line (1- line))))
       (user-error "File does not exist."))))
 
+(evil-define-command evil-find-file-at-point-visual (&optional create-on-error)
```
tomdl89, line 3401: What's `create-on-error` emulating of vim? `:e <cfile>`? As it's not a feature of gf in vim, and I can't see you having used it anywhere (yagni), I'm not sure it should be part of the PR.?

rolag:
I used it in my configuration but you are right it's not used in the pull request and as it's not a Vim feature I'll remove it.

tomdl89:
Thanks for the PR!

PR #1440, opened 2021-03-29
Added `evil-beginning-of-line-or-visual-line`

NightMachinary:
One thing I don't understand is why `evil-beginning-of-line` does not accept the `count` argument.

~~Also, when I move my cursor with `(evil-beginning-of-line)` (using `g 0` and `g $`, essentially) when `evil-respect-visual-line-mode` is on, my whole `evil` setup sometimes (70% of the time?) gets into a buggy state where doing, e.g., `dd` errors `evil-motion-range: Args out of range: nil, 2305843009213693951`.~~

Update: Using `g $` to go to the end of the line, goes to after the last character (which is a config thing I have turned on), and then using `dd` errors `evil-motion-range: Args out of range: nil, 2305843009213693951`. Even though `$` also goes to after the last character, this doesn't happen there. Should I open a separate issue?

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -203,6 +203,18 @@ If COUNT is given, move COUNT - 1 lines downward first."
       (beginning-of-visual-line)
     (beginning-of-line)))
 
+(evil-define-motion evil-beginning-of-line-or-visual-line (count)
+  "Move the cursor to the first character of the current screen
+line if `visual-line-mode' is active and
+`evil-respect-visual-line-mode' is non-nil.  If COUNT is given,
+move COUNT - 1 screen lines forward first."
+  :type inclusive
+  (if (and (fboundp 'beginning-of-visual-line)
+           evil-respect-visual-line-mode
+           visual-line-mode)
+      (beginning-of-visual-line count)
+    (evil-beginning-of-line)))
+
 (evil-define-motion evil-end-of-visual-line (count)
   "Move the cursor to the last character of the current screen line.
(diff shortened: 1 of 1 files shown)


tomdl89:
> why evil-beginning-of-line does not accept the count argument.

evil-beginning-of-line is, by default, triggered by pressing `0` so it couldn't take a count argument in the standard vim way, because we couldn't differentiate between e.g. `5 0` and `50`. neither `0` nor `g0` take a `[count]` in vim anyway. I think I would therefore be inclined to close this PR, but if you think I'm wrong, please explain :)

Regarding the bug you see, can you create a minimal repro using `make emacs` in your evil repo? And drop that in a separate PR? Assuming you do `(setq evil-move-beyond-eol t)` beforehand, and then do `g$dd`, I can't replicate any out-of-range error.

NightMachinary:
@tomdl89 I think just removing the count parameter and having a `evil-beginning-of-line-or-visual-line` is good, but feel free to close the PR.

Since you can't reproduce the bug, I assume it'll be time-consuming to get to the root of it, and I am just remapping my keys to never respect visual lines.

Thanks.

tomdl89:
Thanks @NightMachinary. As you're not using this motion in a key mapping or as a `:motion` for a command, I can't see that it is helping evil emulate vim in emacs (i.e. this isn't functionality in vim, right?) I think handy extras like this probably would sit better in a package which extends evil - be that an already-existing package, or a new one aimed at improving handling (or ignoring) of visual lines.

PR #1448, opened 2021-04-04
Add bs-mode to evil-emacs-state-modes.

daanturo:
Normal state's bindings don't work there by default. (And maybe) It's
not worthy for evil-collection to define them.

### DIFF
--- a/evil-vars.el
+++ b/evil-vars.el
@@ -704,6 +704,7 @@ expression matching the buffer's name and STATE is one of `normal',
     bookmark-bmenu-mode
     bookmark-edit-annotation-mode
     browse-kill-ring-mode
+    bs-mode
     bubbles-mode
     bzr-annotate-mode
     calc-mode


tomdl89:
>Normal state's bindings don't work there by default.

I've just enabled bs-mode in a `make emacs` session, and I can use normal state bindings. @daanturo can you suggest one to try which won't work? Thanks

daanturo:
> > Normal state's bindings don't work there by default.

Sorry for my bad explanation. What I meant was normal state's bindings overriding `bs-mode-map`'s important ones.

Notably:
```
RET             bs-select
C-o             bs-tmp-select-other-window
SPC             bs-select
!               bs-select-in-one-window
F               bs-select-other-frame
c               bs-select-next-configuratio
f               bs-select
o               bs-select-other-window
<mouse-2>       bs-mouse-select
<mouse-3>       bs-mouse-select-other-frame
```
Like `ibuffer`, we can delegates those to `evil-collection`, although I think it's not worth the efforts, just use emacs-state since that's an edge case anyway.

tomdl89:
Sorry - yes that makes perfect sense. Thanks!

PR #1457, opened 2021-04-24
use `noerror` properly in `evil-get-register`

tomdl89:
Fixes what looks like a mistake in `evil-get-register` where no errors were ever thrown. Looks like it breaks evil-owl, so won't merge until evil-owl can handle (or, most likely, ignore) the errors.


tomdl89:
Using `noerror` rather than changing unless -> when means evil-owl does not break 🎉

Issue #1469, opened 2021-05-31
Alternating between visual line mode aware commands and regular commands breaks the visual line mode aware commands


DevJac:
# Issue type
- Bug report

# Environment

Emacs version: 27.2
Operating System: Linux, Fedora
Evil version: 1.14.0
Evil installation type: MELPA (?)
Graphical/Terminal: X
Not tested with `make emacs`

# Reproduction steps
- Fill a buffer with a line or two of text that are long enough to wrap
- Turn on visual line mode
- Place you cursor on one of the long lines
- Run `evil-end-of-line`
- Run `evil-previous-visual-line`
- Observe that `evil-previous-visual-line` does not work as expected

# Expected behavior
I expect `evil-previous-visual-line` to move the cursor up one visual line.

# Actual behavior
The cursor moves to the beginning of the actual previous line (not the previous **visual** line, the previous **actual** line).

# Further notes
This happens because `evil-end-of-line` sets `last-command` to `next-line`. This then causes problems in `line-move` (a function in `simple.el`) on the following lines:

```
...
      (if (and line-move-visual
...
	       (not (and (memq last-command
			       `(next-line previous-line ,this-command))
...
	  (prog1 (line-move-visual arg noerror)
...
	(line-move-1 arg noerror)))))
...
```

I've chopped up the `line-move` function, you can of course view the entire thing in Emacs. The problem is that when `last-command` is `next-line` it causes the cursor to move via `line-move-1`, but we want to cursor to move via `line-move-visual`. I don't know what the right fix is.

DevJac:
```
(evil-define-motion evil-end-of-line (count)
  "Move the cursor to the end of the current line.
If COUNT is given, move COUNT - 1 lines downward first."
  :type inclusive
  (move-end-of-line count)
  (when evil-track-eol
    (setq temporary-goal-column most-positive-fixnum
          this-command 'next-line))
  (unless (evil-visual-state-p)
    (evil-adjust-cursor)
    (when (eolp)
      ;; prevent "c$" and "d$" from deleting blank lines
      (setq evil-this-type 'exclusive))))
```
Why does this function set `this-command` to `next-line`? `this-command` later becomes `last-command` and is the cause of this bug.

DevJac:
I created a workaround for this issue by changing the following two functions:
```
(evil-define-motion evil-next-visual-line (count)
  "Move the cursor COUNT screen lines down."
  :type exclusive
  (let ((line-move-visual t))
[...]

PR #1470, opened 2021-05-31
Consider line-move-visual and last-command in evil-line-move

DevJac:
line-move is hard-coded such that it will not perform visual movements
when last-command is next-line or previous-line. If
line-move-visual is set, then we are trying to move the point visually,
but this will not happen if last-command is next-line or
previous-line. Thus, we set last-command to nil in these cases.

Fixes https://github.com/emacs-evil/evil/issues/1469

### DIFF
--- a/evil-common.el
+++ b/evil-common.el
@@ -1373,6 +1373,9 @@ Signals an error at buffer boundaries unless NOERROR is non-nil."
         (evil-line-move count)
       (error nil)))
    (t
+    (when (and line-move-visual
+               (memq last-command `(next-line previous-line)))
+      (setq last-command nil))
     (evil-signal-without-movement
       (setq this-command (if (>= count 0)
                              #'next-line


tomdl89:
Thanks for the PR. I was just looking into this myself - looks like a good solution. Are you planning to add tests? I can, but it may be better from you, as you seem more familiar with expected outcomes.

DevJac:
After more testing, this is not a good solution. It causes visual line movements past empty lines to reset the column to 0, the left-most column.

```
aaa|aaa

bbbbbb
```
Press `jj` and then you'll have:
```
aaaaaa

|bbbbbb
```
This would be a serious regression.

@tomdl89 if you want to take this over and come up with your own solution, you're more than welcome. I'll still be hacking on it, so maybe just ping me if you come up with something that works.

tomdl89:
Ah that's a shame. What was wrong with the solutions you proposed in #1469 ?

DevJac:
I'll try those other solutions. Let's move discussion to the issue rather than this closed PR.

PR #1471, opened 2021-06-02
Add :window-width to `evil-test-buffer` args

tomdl89:
So we can reliably test visual-line related functionality.


tomdl89:
Failing because the terminal width on Travis is 80 cols.

PR #1474, opened 2021-06-12
Remove travis.yml, add test.yml github workflow

tomdl89:
@TheBB I've had a go at moving evil over to github actions. Everything looks fine, except emacs 24.5 has 20 failures. Any ideas?


tomdl89:
@wasamasa do you have any ideas on this? My current plan is to change us over to github actions and disable the 24.5 tests tonight, as travis claims it will disable tests running on travis-ci.org tomorrow.

wasamasa:
Leave the 24.5 tests as is because 24.5 runs on the still supported Debian 9 release. As for debugging, spin up Debian 9 (in a VM, container, whatever) and take a look how it differs between GUI and terminal and batch modes. Chances are that Emacs changed something between 24.5 and 25.3 for that.

TheBB:
Sorry for not replying. I'm not sure why the tests are failing.

But don't feel beholden to the Travis deadline. We can remove the Travis config and the world won't end if we're without CI for Emacs 24.5 for a while (or even if we're without CI at all for a while). So don't sweat it.

tomdl89:
Thanks both. I'll take a look at Debian 9 this evening, but assuming I don't find a quick answer, I'll disable 24.5 in CI temporarily, so we at least have some testing until I (or someone else) can fix.

wasamasa:
I don't see why tests failing for 24.5 would mean that no tests are run at all. Travis has a feature of running jobs independently, so that a failing job doesn't prevent others from running. I imagine GitHub Actions to be similar.

TheBB:
By default, at least, the whole matrix is immediately aborted if any instance fails. There might be an option to change that.

wasamasa:
Yes indeed: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast

tomdl89:
See line 9 of test.yml in this PR. I already have that disabled. I was just thinking that having the red x on master wouldn't be desirable. But if we don't mind that, then I'll merge this and we can live with partial failing tests for a little while.

tomdl89:
OK it's merged. If either of you wouldn't mind taking a look at the testing setup, I'd really appreciate it. It's not something I know too much about, but I feel it is important to get it working as well as the travis setup. Thanks for your input so far :)

PR #1475, opened 2021-06-12
Fix evil-find-char on empty line

axelf4:
Previously, with the cursor on an empty line `evil-find-char` would continue searching on the next line. This was because `forward-char` would advance to the next line, after which the `search-forward` bound would be wrongly calculated.

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -766,7 +766,7 @@ Movement is restricted to the current line unless `evil-cross-lines' is non-nil.
         (visual (and evil-respect-visual-line-mode
                      visual-line-mode)))
     (setq evil-last-find (list #'evil-find-char char fwd))
-    (when fwd (forward-char))
+    (when fwd (evil-forward-char 1 evil-cross-lines))
     (let ((case-fold-search nil))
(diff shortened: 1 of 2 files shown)


tomdl89:
Thanks for the PR! Maybe I'm missing something here, but isn't the fix just to change `(forward-char)` to `(evil-forward-char)`?

tomdl89:
@axelf4 I'll merge my suggestion instead unless you have any objection? Thanks

axelf4:
I see now that `search-forward` is a little sketchy in that it does not short-circuit, so I agree that `evil-forward-char` will be better (but it does change the message from "Can't find _" to "End of line"). I can just amend the commit.

tomdl89:
Sounds good. Tests are a bit broken right now, so don't be alarmed at the red x.

tomdl89:
You'll need `(evil-forward-char 1 evil-cross-lines)` instead, in order for this to respect `evil-cross-lines`.

axelf4:
Right, sorry I forgot. :+1:

PR #1478, opened 2021-06-15
evil-core: refactored  `evil-get-auxiliary-keymap`

lebensterben:
In `evil-get-auxilliary-keymap`, there's already a check on `state` when entering the function.
Should `state` be `nil`, it won't enter the `let*` form.

The binding specification
`(aux (if state (lookup-key map key) map))`
always reduces to
`(aux (lookup-key map key))`

And `(parent-aux (when (and ignore-parent (keymap-parent map) state)))`
always reduces to
`(parent-aux (when (and ignorant-parent (keymap-parent map))))`

This commit removed unnecessary `if` form, and removed redundant `state` in `and` form, as described above.

### DIFF
--- a/evil-core.el
+++ b/evil-core.el
@@ -871,8 +871,7 @@ one already."
   (when state
     (let* ((key (vconcat (list (intern (format "%s-state" state)))))
            (parent-aux (when (and ignore-parent
-                                  (keymap-parent map)
-                                  state)
+                                  (keymap-parent map))
                          (lookup-key (keymap-parent map) key)))
            (aux (lookup-key map key)))
       (cond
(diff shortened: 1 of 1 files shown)


evil-core.el
```diff
@@ -874,7 +874,7 @@ one already."
                                   (keymap-parent map)
                                   state)
```
tomdl89, line 875: You may as well remove this `state` as well?

tomdl89 (changes requested):

evil-core.el
lebensterben: Good catch! I removed it in the new commit.

tomdl89:
Thanks for the PR. Always good to keep evil tidy.

PR #1487, opened 2021-07-07
evil-maybe-remove-spaces works for evil-change-line

roxma:
This change make `cc` + `ESC` on indented line produce the same behavior
as vim/nvim

### DIFF
--- a/evil-states.el
+++ b/evil-states.el
@@ -99,6 +99,8 @@ commands opening a new line."
                  evil-open-below
                  evil-append
                  evil-append-line
+                 evil-change
+                 evil-change-line
                  newline
                  newline-and-indent
                  indent-and-newline)))
--- a/evil-states.el
+++ b/evil-states.el
@@ -101,6 +101,7 @@ commands opening a new line."
                  evil-append-line
                  evil-change
                  evil-change-line
+                 evil-change-whole-line
                  newline
                  newline-and-indent
                  indent-and-newline)))


evil-states.el
```diff
@@ -99,6 +99,8 @@ commands opening a new line."
                  evil-open-below
                  evil-append
                  evil-append-line
+                 evil-change
```
tomdl89, line 102: I understand the `evil-change-line` addition, but not `evil-change` if I do `c$<escape>` when at the start of a line, this is an `evil-change` but vim doesn't remove the indentation spaces..?

evil-states.el
```diff
@@ -99,6 +99,8 @@ commands opening a new line."
...
+                 evil-change-line
```
tomdl89, line 103: I'd also be inclined to add `evil-change-whole-line` and any others which you notice are the same in vim.

tomdl89:
p.s. you can ignore the failing test in 24.5 - it's borked and we're still working on it.

evil-states.el
roxma: - `evil-change` is needed for  `cc <escape>` to remove indentation spaces
  - As a side effect, this also makes  `c $ <escape>` remove indentation spaces
- `evil-change-line` is needed for `C <escape>` to remove  indentation spaces
  - But on my vim `C <escape>` does not remove the indentation spaces

I don't know how to make both `cc <escape>` and `c $ <escape>` behave the same as vim. But the minor difference is acceptable for me. 

My actual need is to automatically remove trailing spaces at the cursor line when I leave insert state.

tomdl89: Ah yes, I understand. I've noticed that vi, vim and neovim all behave differently. So evil doesn't really have a solid example to follow. I agree that `c$<escape>` is probably acceptable. I won't merge this without tests, so either you'll have to add them, or you'll have to wait for me to add them (sometime next week I hope).

tomdl89:
@roxma I made PR #1493 to resolve the discussed issues and add tests. Lemme know what you think

Issue #1489, opened 2021-07-08
Scroll up/down scrolls a constant value independent of zoom level


countvajhula:
# Issue type

- Bug report

# Environment

Emacs version: GNU Emacs 27.1 (build 1, x86_64-apple-darwin17.7.0, Carbon Version 158 AppKit 1561.6) of 2020-08-16
Operating System: Mac OS High Sierra
Evil version: 1.14.0
Evil installation type: MELPA
Graphical/Terminal: Graphical
Tested in a `make emacs` session (see [CONTRIBUTING.md](https://github.com/emacs-evil/evil/blob/master/CONTRIBUTING.md#bug-reports)): No

# Reproduction steps
- Start Emacs
- Zoom in a couple of times using `C-x C-+` (`zoom-in`)
- Scroll using `C-u` and `C-d`

# Expected behavior
Should scroll half a screen.

# Actual behavior
Scrolls more than half a screen.

Likewise, if you zoom out from the default zoom level, it scrolls less than half a screen, because it is scrolling a constant value reported by Emacs's `window-body-height` function.

# Further notes
I often zoom in on windows when I'm using a larger monitor, where I have more screen real estate and may be further away from the screen. In such cases, `C-d` and `C-u` scroll too much and lose the working context.

PR #1490, opened 2021-07-08
Scroll half of the visible screen respecting zoom level

countvajhula:
Fix for #1489

### DIFF
--- a/evil-common.el
+++ b/evil-common.el
@@ -3983,7 +3983,6 @@ PROPERTIES is a property-list which supports the following properties:
   "The visible height of WINDOW in lines.
 
 If no WINDOW is specified, use the selected one."
-  (interactive)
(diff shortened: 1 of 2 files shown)


evil-common.el
```diff
@@ -3977,6 +3977,22 @@ PROPERTIES is a property-list which supports the following properties:
          (evil-motion-state))
        (switch-to-buffer-other-window buf))))
 
+;;; Window
+
+(defun evil-window-visible-height (&optional window)
+  "The visible height of WINDOW in lines.
+
+If no WINDOW is specified, use the selected one."
+  (interactive)
```
tomdl89, line 3986: why interactive?

countvajhula: Oh, didn't mean to make it interactive. Removed!

PR #1491, opened 2021-07-09
Add evil-ispell-mark-word-as-good and "zg" binding

condy0919:
According to https://vimhelp.org/spell.txt.html, `zg` would

> Add word under the cursor as a good word to the first name in 'spellfile'. A
> count may precede the command to indicate the entry in 'spellfile' to be used.
> A count of two uses the second entry.
>
> In Visual mode the selected characters are added as a word (including white
> space!). When the cursor is on text that is marked as badly spelled then the
> marked text is used. Otherwise the word under the cursor, separated by
> non-word characters, is used.
>
> If the word is explicitly marked as bad word in another spell file the result
> is unpredictable.

This patch adds basic support.


condy0919:
Why it failed on Emacs 24.5 only?

evil-commands.el
```diff
@@ -636,6 +636,27 @@ and jump to the corresponding one."
           (when ov
             (goto-char (overlay-start ov))))))))
 
+(defun evil-ispell-mark-word-as-good (word)
+  "Add WORD under the cursor as a good word to the Ispell dictionary."
+  (interactive (list (thing-at-point 'word)))
+  (ispell-send-string (concat "*" word "\n"))
```
wasamasa, line 642: `ispell-send-string` isn't autoloaded, so you must add `(require 'ispell)` somewhere, for example at the beginning of the file.

evil-commands.el
```diff
@@ -636,6 +636,27 @@ and jump to the corresponding one."
...
+  ;; dictionary modified!
+  (setq ispell-pdict-modified-p '(t))
+  (ispell-pdict-save ispell-silently-savep)
+  (when-let ((flyspell-mode)
```
wasamasa, line 646: Don't use `when-let`. For this particular case, `(bound-and-true-p flyspell-mode)` is better. The other case is a combination of `let` and `when`.

evil-commands.el
```diff
@@ -636,6 +636,27 @@ and jump to the corresponding one."
...
+             (start (car (bounds-of-thing-at-point 'word))))
+    (flyspell-unhighlight-at start))
+  (message "Word '%s' added to dictionary" (upcase word)))
+
+(defun evil-ispell-mark-word-as-locally-good (word)
+  "Add WORD under the cursor to the local word list."
+  (interactive (list (thing-at-point 'word)))
+  (ispell-add-per-file-word-list word)
+  (when-let ((flyspell-mode)
```
wasamasa, line 655: See previous comment.

wasamasa (changes requested):

evil-commands.el
condy0919: done

condy0919:
I can't figure out why it failed only on Emacs 24.5. Download the Emacs 24.5 tarball, grep everything symbol in the `lisp` directory, all have definitions.

evil-commands.el
```diff
@@ -636,6 +637,29 @@ and jump to the corresponding one."
...
+           (flyspell-unhighlight-at start))))
+  (message "Word '%s' added to dictionary" (upcase word)))
+
+(defun evil-ispell-mark-word-as-locally-good (word)
+  "Add WORD under the cursor to the local word list."
+  (interactive (list (thing-at-point 'word)))
+  (ispell-add-per-file-word-list word)
+  (when (bound-and-true-p flyspell-mode)
+    (let ((start (car (bounds-of-thing-at-point 'word))))
+      (and start
```
wasamasa, line 659: Use `when`, not `and` when dealing with side effects.

wasamasa:
See https://github.com/emacs-evil/evil/pull/1487#issuecomment-876368014

evil-commands.el
condy0919: Updated.

condy0919:
> See [#1487 (comment)](https://github.com/emacs-evil/evil/pull/1487#issuecomment-876368014)

I see...

condy0919:
@wasamasa ping

wasamasa:
More general issue: The Vim helpfile claims a few more things about those commands:

- There may be a count to indicate which spellfile entry to use, defaulting to the first. For this you'd need to use `evil-define-command` with `(interactive "<c>")`
- Visual mode is not handled at all. Some experimentation with Vim is necessary to figure out how exactly it behaves. It claims both that a selection including whitespace may be used, but that a word at point may be used as well.

KSSysLab:
> More general issue: The Vim helpfile claims a few more things about those commands:
> 
> * There may be a count to indicate which spellfile entry to use, defaulting to the first. For this you'd need to use `evil-define-command` with `(interactive "<c>")`. Does the concept of spellfile entries even make sense for ispell? What about repeatability? If not, then it should be marked as non-repeatable.

No, `ispell-personal-dictionary` can't be a list. But I found [doom-emacs](https://github.com/hlissner/doom-emacs/blob/develop/modules/checkers/spell/autoload/%2Bflyspell.el#L31) provides a non-compatible mechanism which behaves like there are 3 spellfile entries (personal/buffer-local/session dict) when "zg" adding words.

> 
>  internal-wordlist
> The internal word list is used for all buffers where 'spell' is set.  It is
> not stored, it is lost when you exit Vim.  It is also cleared when 'encoding'
> is set.

"zG" adds the word to the `internal-wordlist` which is effective in session-level. Yes, this patch implementes wrong. Sorry :'(

We can't mimic vim perfectly, so which one is better is arguable.

1. "zg" `personal`/`buffer-local`/`session` dict, "zG" `buffer-local` dict.
2. "zg" `personal`/`buffer-local`/`session` dict, "zG" `session` dict.
3. "zg" `personal`/`buffer-local` dict, "zG" `session` dict.

Personally I prefer the third one, it's more orthogonal. Thoughts?

> * Visual mode is not handled at all. Some experimentation with Vim is necessary to figure out how exactly it behaves. It claims both that a selection including whitespace may be used, but that a word at point may be used as well.

I try to avoid this when implementating. Consider the following case: `mis speelled` has been inserted in the personal dict, but ispell only recognizes the `speelled` as CORRECT.

image

condy0919:
:'( it's me

wasamasa:
> No, `ispell-personal-dictionary` can't be a list. But I found [doom-emacs](https://github.com/hlissner/doom-emacs/blob/develop/modules/checkers/spell/autoload/%2Bflyspell.el#L31) provides a non-compatible mechanism which behaves like there are 3 spellfile entries (personal/buffer-local/session dict) when "zg" adding words.

Uh-oh. I'd consult with the Doom people then. Maybe they'd prefer your approach. Maybe their approach is worth using.

> "zG" adds the word to the `internal-wordlist` which is effective in session-level. Yes, this patch implementes wrong. Sorry :'(
> 
> We can't mimic vim perfectly, so which one is better is arguable.
> 
> 1. "zg" `personal`/`buffer-local`/`session` dict, "zG" `buffer-local` dict.
> 2. "zg" `personal`/`buffer-local`/`session` dict, "zG" `session` dict.
> 3. "zg" `personal`/`buffer-local` dict, "zG" `session` dict.
> 
> Personally, I prefer the third one, it's more orthogonal. Thoughts?

Agreed, the third option makes most sense.

> EDITED: Playing in vim, it's non-rerepeatable.
> 
> > * Visual mode is not handled at all. Some experimentation with Vim is necessary to figure out how exactly it behaves. It claims both that a selection including whitespace may be used, but that a word at point may be used as well.
> 
> I try to avoid this when implementating. Consider the following case: `mis speelled` has been inserted in the personal dict, but ispell only recognizes the `speelled` as CORRECT.
> 
> image

OK, so not repeatable and no handling of visual mode either. In that case it's debatable whether `evil-define-cmd` adds any benefit.

condy0919:
Updated

condy0919:
Seems like no more comments. 

image

condy0919:
Updated. string without properties will be used. @wasamasa

condy0919:
> Oh, I see! I took some time to read up on it and I think your proposal makes sense; `zg` -> personal dict, `zG` -> session, but I'm not sure about using the universal argument; the universal arg is generally very unvimlike. Perhaps there's another key we could use? Since evil doesn't have `zu` for undo, how about `zu` -> add to buffer-local personal dict and `zU` -> remove from buffer-local personal dict?

> Not sure we need the old `zu` keys since, with ispell, I assume we can undo a `zg` with `zG`, or a `zw` with `zW`. Need to test that assumption though.

PR #1494, opened 2021-07-15
evil-vars: reset evil-overriding-maps to an empty list

condy0919:
It's more intuitive for evil beginners if no keymap overriding evil maps.


condy0919:
@tomdl89

PR #1499, opened 2021-07-23
Use text scale in calculating visible window height

### LINKED ISSUE #1497: evil-scroll-up and evil-scroll-down do not work correctly when end of buffer is visible
# Issue type

Bug report

# Environment

Emacs version: 28.0.50
Operating System: Ubuntu 20.4
Evil version: 1.14.0
Evil installation type: MELPA, MELPA stable, Quelpa, El-Get, manual -->
Graphical/Terminal: Graphical

# Reproduction steps
- Start Emacs with `make emacs`
- Open file with enough lines to fill a whole window
- Using `C-e`, scroll down so only the last line of the buffer is visible
- `M-x evil-scroll-up`

# Expected behavior
The window should be half filled with lines from the buffer.

# Actual behavior
The window doesn't scroll at all.

# Further notes
This is because `evil-scroll-up` (and `evil-scroll-down`) use `evil-visible-window-height` which reports incorrect values for window height when the window is positioned over the end of the buffer.

countvajhula:
Fixes #1497 

@tomdl89 your idea to use text scale works perfectly. Initially I did think it was over-reporting the window height, but upon closer inspection, it is reporting the total number of visible _lines of text_, not _lines of code_, which is exactly what we are looking for, and matches Vim behavior as well as far as I was able to check.

To test:
1. create a buffer with single words, e.g. 100ihello<RET><esc>
2. now check `(evil-window-visible-height)` at different zoom levels. It should report the correct value
3. create a buffer with long lines, e.g. 20ihello<space><RET><esc> along with other, regular-length lines
4. now check `(evil-window-visible-height)` again. It will show a larger number than what is actually displayed in line numbers on the screen.
5. Looking at the line number column (e.g. `global-display-line-numbers-mode`), count the number of blanks between line numbers. These are lines that are actually displayed, but which do not correspond to line numbers because the lines are wrapped. Add this number to the visible number and you should get the number reported by `evil-window-visible-height`.

Looking at the behavior in Vim, I wasn't able to detect a difference at different zoom levels, or at the end of the buffer.
 
cc @chopptimus - please let us know if you have a chance to verify the fix and notice any other differences 🙏

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -1043,7 +1043,7 @@ If the scroll count is zero the command scrolls half the screen."
     (when (= (point-min) (line-beginning-position))
       (signal 'beginning-of-buffer nil))
     (when (zerop count)
-      (setq count (/ (window-body-height) 2)))
+      (setq count (/ (evil-window-visible-height) 2)))
     (let ((xy (evil-posn-x-y (posn-at-point))))
       (condition-case nil
           (progn
@@ -1067,7 +1067,7 @@ If the scroll count is zero the command scrolls half the screen."
     (setq evil-scroll-count count)
     (when (eobp) (signal 'end-of-buffer nil))
     (when (zerop count)
-      (setq count (/ (window-body-height) 2)))
+      (setq count (/ (evil-window-visible-height) 2)))
     ;; BUG #660: First check whether the eob is visible.
     ;; In that case we do not scroll but merely move point.
     (if (<= (point-max) (window-end))
--- a/evil-common.el
+++ b/evil-common.el
@@ -3980,6 +3980,18 @@ PROPERTIES is a property-list which supports the following properties:
          (evil-motion-state))
        (switch-to-buffer-other-window buf))))
 
+;;; Window
+
+(defun evil-window-visible-height (&optional window)
+  "The visible height of WINDOW in lines.
+
+If no WINDOW is specified, use the selected one."
+  (let ((window (or window (selected-window))))
+    (save-window-excursion
+      (select-window window)
+      (let ((current-scale (expt text-scale-mode-step text-scale-mode-amount)))
+        (round (/ (window-body-height) current-scale))))))
+
 (provide 'evil-common)
 
 ;;; evil-common.el ends here


tomdl89:
That's great news, thanks. I'll take a look on the weekend. Agree I'd like @chopptimus to take a look if possible.

countvajhula:
I noticed that one of the edge cases that @chopptimus pointed out still fails -- if you have a long wrapped line that fills the entire screen, then `C-d` doesn't scroll the page.

Some notes towards a fix:
- this happens because even though the window is scrolled by N _visual_ lines, since point is still in the same wrapped line, nothing happens because the entire function is enclosed in a `evil-save-column` macro which returns point to the initial position in the line
- I've noticed that Vim does not preserve the column but instead always goes to first non-blank. If we remove this macro, scrolling works, but I'm not sure this is the right fix since it doesn't conform to either existing evil or Vim behavior, and I think the existing evil behavior (of preserving column) is a good idea in principle.
- Another possibility is to say, if after scrolling we are still on the same line, then call `(next-line)`. It looks like Vim actually does something different here, but I don't see what the pattern is exactly, as it appears to scroll several lines and not just one.

Any preferences here @tomdl89 ?

chopptimus:
Vims behaviour when scrolling long wrapped lines is pretty inconsistent. If I have a window with 50 lines of text visible but only 5 lines of code, vim scrolls 4 lines, and if I have only a single line of code visible it scrolls a single line.

tomdl89:
Some notes from testing:

- You'll need to `(require face-remap)` at the top of evil-commands.el to make sure `text-scale-mode-...` vars are always there when you need them
- I agree it works perfectly when zoomed in, but it's not _quite_ perfect when zoomed out. Probably acceptable though
- I'm confused about `evil-save-column` - it doesn't seem to prevent moving column currently in master (when hitting `C-d` on a line that fills the window)? I'm unclear on how `evil-scroll-up` and `evil-scroll-down` are circumventing it.
- Because of the above point, I do still think we need a fix before we can merge. I'd like to understand how the current implementation works better before changing it.
Thanks :)

countvajhula:
@chopptimus , yeah that's exactly what I noticed too. @tomdl89 good points, I'll take another look when I have a chance unless you get to it first.

Issue #1504, opened 2021-08-02
Infinite prompts "file changed" with revert-buffer-insert-file-contents-delicately + Evil combo


Hi-Angel:
# Issue type

- Bug report

# Environment

Emacs version: GNU Emacs 28.0.50 (build 4, x86_64-pc-linux-gnu, GTK+ Version 3.24.30, cairo version 1.17.4) of 2021-08-03
Operating System: Archlinux
Evil version: Evil version 1.14.0
Evil installation type: MELPA
Graphical/Terminal: Graphical
Tested in a `make emacs` session (see [CONTRIBUTING.md](https://github.com/emacs-evil/evil/blob/master/CONTRIBUTING.md#bug-reports)): Yes

# Reproduction steps
1. Start Emacs with `make emacs` from evil repo
2. Evaluate `(setq revert-buffer-insert-file-contents-function 'revert-buffer-insert-file-contents-delicately)`
3. Evaluate `(global-auto-revert-mode 1)`
4. Execute `echo hello > /tmp/1`
5. In the Emacs, open the created `/tmp/1` file
6. Press <kbd>G</kbd> to cause initial position to be put into evil jumps buffer.
7. Execute `echo hello > /tmp/1`
8. Press <kbd>C-o</kbd> to jump back

# Expected behavior

Caret changes its position

# Actual behavior

Emacs asks `File 1 changed on disk.  Reread from disk? (yes or no)`

# Further notes

I am not sure if it's upstream bug or Evil, but as I presume Evil might be using its own buffer to keep jump positions, so I decided to report it here. Please tell if it needs to be moved upstream *(and why, so I can create the report)*.

The prompt will be popping up infinitely on every move with <kbd>C-o</kbd> and <kbd>C-i</kbd> until you make a change in the buffer.

`2` is the specific step causing the problem. This step is a workaround for Emacs breaking markers in buffers upon reverting them, [see also this report](https://github.com/magit/magit/issues/4442). However, attempt to use the workaround turned up this even worse problem than breaking markers in a buffer.

Hi-Angel:
> I am not sure if it's upstream bug or Evil, but as I presume Evil might be using its own buffer to keep jump positions, so I decided to report it here. Please tell if it needs to be moved upstream *(and why, so I can create the report)*.

[...]

PR #1506, opened 2021-08-05
Do not try to switch file/buffer when jumping inside same file

Hi-Angel:
The vast majority of jumps will end up within the current opened
file/buffer, simply because usually users do more movement inside a
buffer than between multiple buffers.

There is even a special case when evil-jumps-cross-buffers is set to
nil, which causes jumps never ever switch buffers.

So, let's do a small optimization, and not try to call find-file or
switch-to-buffer if we know it is unnecessary.

As a side-effect, this should also help to users to stumble less upon
https://github.com/emacs-evil/evil/issues/1504 bug (it has its own
upstream fix, but until that landed, this small optimization is known to
help as well).


Hi-Angel:
I assume the v24.5 tests fail is unrelated, because looking at the recent PRs, all of them has that.

evil-jumps.el
```diff
@@ -156,9 +156,10 @@
                (pos (car place))
                (file-name (cadr place)))
           (setq evil--jumps-jumping t)
-          (if (string-match-p evil--jumps-buffer-targets file-name)
-              (switch-to-buffer file-name)
-            (find-file file-name))
+          (when (not (string= current-file-name file-name))
```
tomdl89, line 159: `(unless (string= current-file-name file-name)`
tends to be the idiom

tomdl89:
Indeed the v24.5 tests fail is unrelated. We need to get around to fixing that. Thanks for the PR. Looks good.

evil-jumps.el
Hi-Angel: Nice idea, thank you! Done!

Issue #1507, opened 2021-08-07
Mentioning NonGNU ELPA


phikal:
(Not an actual issue)

Hey, I just wanted to let you know that evil has been added to [NonGNU ELPA](https://elpa.nongnu.org/): http://elpa.nongnu.org/nongnu/evil.html

This shouldn't affect development, the main difference is that more people will probably be using a stable version. These are detected by commits that bump the version tag. I checked a few of your last manually tagged releases, and it seems that this is already the case.

NonGNU ELPA will be activated by default from Emacs 28 onwards, but it can be enabled in previous versions just like MELPA. So if the evil-team is interested, you could mention it as one way for installing the package.

wasamasa:
Thanks for the heads-up. I agree that a new stable release is long overdue. Please take a look at https://github.com/emacs-evil/evil/pull/1508.

wasamasa:
I've merged the linked PR. That leaves cutting a new release. I haven't done that before, so maybe someone else can chime in. There's at least the following to consider:

- Documentation needs to be regenerated with Sphinx and included into the release commit.
- Deciding on a version number to pick.
- Picking a suitable commit or features that must be included into the release.
- Testing for stability.
- Announcements (there is a helper script that modifies the Emacswiki page, changes the IRC topic and sends an email to the mailing list).

tomdl89:
I asked TheBB this question back in April (with the intention to do a release) and he helpfully answered: https://github.com/emacs-evil/evil/issues/1403#issuecomment-826788873
If you don't have the time/inclination, lemme know and I'll see if I can take a look, as I said I would back then :)

PR #1508, opened 2021-08-08
Mention NonGNU ELPA

wasamasa:
See https://github.com/emacs-evil/evil/issues/1507.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -12,7 +12,7 @@
 Emacs version: <!-- M-x emacs-version -->
 Operating System: <!-- Name and version, e.g. Windows 10, macOS Sierra, Debian Jessie -->
 Evil version: <!-- M-x evil-version -->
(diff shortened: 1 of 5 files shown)


tomdl89 (approved):

phikal:
I apologize for noticing this somewhat late, but was it also intended to update the list of badges at the top of the README file?

wasamasa:
I briefly considered to do that, but decided against it. Surely GNU stuff doesn't have badges, right?

phikal:
Every ELPA package gets a badge generated:

badge

see http://elpa.nongnu.org/nongnu/evil.html.

A few packages are already using them. Personally I'm not a fan of the concept, but a few packages are already using them and posts like [these](https://www.reddit.com/r/emacs/comments/p181d3/is_there_any_plan_to_move_evil_to_nongnu_elpa/) indicate that it might be easy to miss if a package is available in a package repository just from textual documentation. That being said, there is of course no obligation to add it, I was just wanted to check.

wasamasa:
Alright, I've added it.

PR #1512, opened 2021-09-04
Change evil-window-{next,prev} to respect other-window window parameters

kljohann:
Currently `evil-window-next` and `evil-window-prev`, in the case where no count is provided, use `select-window` instead of `other-window` (even though they effectively are the evil version of `other-window`). This has the side effect that window parameters relevant to the cyclic window ordering are ignored, in particular
* the `other-window` parameter, which can be set to a function that is used to pick the window to jump to, and
* the `no-other-window` parameter, which can be used to exclude a window from being jumped to by the cyclic commands.

FWIW, `evil-window-{left,right,up,down}` already respect the `no-other-window` parameter (via windmove → `window-in-direction`), so I think using `other-window` here could make the commands more consistent. I don't know why `select-window` was used in the first place.


axelf4:
I think this is a good change, however the test fails for me on GNU Emacs 28.2 using `make emacs`, because the commands still change to the `*ert*` buffer. Would be great if you could take a look.

kljohann:
@axelf4 Thanks for looking into it. The test should be fixed now.

axelf4:
Thanks! The tests were still failing on older versions of Emacs due to `display-buffer-in-direction` being first introduced in GNU Emacs 27.1, so I replaced `display-buffer` with `split-window`. That is also less problematic if the test is run in a session where `display-buffer-alist` is customized.

PR #1516, opened 2021-09-22
Add missing company function to `evil-repeat` list

meain:
Added [`company-complete-common-or-cycle`](https://github.com/company-mode/company-mode/blob/3eaac91b27d5041a97a0cbae32bb5b03e361d230/company.el#L2575) to the list of function for company in `evil-repeat` integration.


meain:
This func actually calls `company-complete-common` internally which is already available.

Issue #1518, opened 2021-09-29
Binding any custom function to "0" in the motion state map causes motions with a "0" in the count argument to fail


jmmathena:
# Issue type
- Bug report

# Environment

Emacs version: GNU Emacs 27.2 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.27, cairo version 1.17.4) of 2021-03-26
Operating System: Arch Linux, Linux 5.10.64-1-lts x86_64
Evil version: 1.14.0
Evil installation type: Melpa
Graphical/Terminal: Graphical
Tested in a `make emacs` session (see [CONTRIBUTING.md](https://github.com/emacs-evil/evil/blob/master/CONTRIBUTING.md#bug-reports)): Yes

# Reproduction steps
- Start Emacs
- Evaluate the following lines, which just rename the function bound to 0 while keeping the logic exactly the same:
```
(defalias 'my-beg 'evil-digit-argument-or-evil-beginning-of-line)
(evil-redirect-digit-argument evil-motion-state-map "0" 'my-beg)
```
- Try any motion command with a 0 in the count argument: e.g. `d10j`, `c20l`, `y3040k` and note in the echo area that the interpreted keystrokes are different than expected, meaning the commands don't work as intended.

# Expected behavior
Users should be able to bind custom commands to 0, modifying the behavior of `evil-beginning-of-line` without breaking how motions are read. E.g. I should be able to create a function called `my-beg` that has special behavior, bind it to 0, and still be able to call `d10j` to delete ten lines down.

# Actual behavior
If a custom command is bound to 0, the reading of motions is broken. As an example, when we use `evil-redirect-digit-argument` to bind a command to 0 (just like Evil [does](https://github.com/emacs-evil/evil/blob/82e5becae21b7768ffbb4728bce83dab8751d610/evil-maps.el#L172)), the command `d10j` is interpreted by Emacs as `C-u 62 j`. Therefore, instead of deleting ten lines down, Emacs will jump 62 lines down without modifying the buffer at all.

# Further notes
I discovered this issue after realizing that while `evil-org-mode` was active, any motions with a 0 in the count argument failed, which I documented at Somelauw/evil-org-mode#91. The reason for this issue is as follows:

[...]

PR #1519, opened 2021-09-29
Fix for #1518

jmmathena:
This PR gets rid of `evil-redirect-digit-argument`, and replaces its effects with a custom variable. The variable is a list of functions that can be bound to digits (always 0 in practice, I'd imagine, but I'm sure [someone's workflow](https://xkcd.com/1172/) would be affected). This allows for greater consistency in how functions can be bound to 0 - just use any of the `define-key` variants like normal, then add the relevant function to the custom variable, instead of binding with `evil-redirect-digit-argument`.

It passes all tests, except for the ones that explicitly rely on "0" being bound to the wrapper function for `evil-beginning-of-line`, which I've fixed.

One thing I'd like feedback on is that I don't like the way that the new `evil-digit-bound-motions` custom variable is set with `customize-set-variable` when `evil-respect-visual-line-mode` is enabled. Since I typically `setq` custom variables in my init.el instead of using a custom file, I don't know if setting this one programmatically will cause issues in general.

This PR also removes `evil-beginning-of-line-or-digit-argument`, since it hasn't been touched since 2011.

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -234,19 +234,6 @@ move COUNT - 1 screen lines downward first."
                        -1
                        (/ (with-no-warnings (window-body-width)) 2)))))
 
-(evil-define-motion evil-beginning-of-line-or-digit-argument ()
-  "Move the cursor to the beginning of the current line.
-This function passes its command to `digit-argument' (usually a 0)
-if it is not the first event."
-  :type exclusive
-  (cond
-   (current-prefix-arg
-    (setq this-command #'digit-argument)
-    (call-interactively #'digit-argument))
-   (t
-    (setq this-command #'evil-beginning-of-line)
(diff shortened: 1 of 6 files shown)


tomdl89:
Awesome! Didn't expect an issue _and_ a PR! I'm on hols right now, and no one else is much active around here, so don't be disheartened if you don't get feedback or merged soon. Shout/pester if you get impatient. I'll try to take a look in a week or two.

tomdl89:
Just took a look and it looks (suspiciously) simple, so great job if my suspicions are unfounded. One thing I noticed is that ln78 of `evil-keybindings` still has a reference to `evil-digit-argument-or-evil-beginning-of-line` so pressing 0 when in `Info` mode now doesn't work. Otherwise, I'll need to spend a bit longer trying to break it before giving a 👍.

PR #1531, opened 2021-10-12
Run make docstrings, doc, info

tomdl89:
@TheBB I ran `make docstrings` `make doc` and `make info` and all succeeded without error. Is that all I need to do? 7,076 additions, 6,661 deletions in the json scares me a bit!


tomdl89:
@TheBB bump

PR #1534, opened 2021-10-26
Fix #1518: "custom functions bound to '0' break motion parsing" in a saner way

### LINKED ISSUE #1518: Binding any custom function to "0" in the motion state map causes motions with a "0" in the count argument to fail

jmmathena:
This PR improves on #1519, which I made about a week ago. I'm pretty new to Emacs development, so when I made those changes I was imagining some graybeard poweruser with way more experience than me, willfully rebinding digit keys besides "0" to bizarre functions. However, after some reflection I imagine that this is such a rare use case that it isn't really necessary to support it. By way of explanation:

In Vim/Evil, we know that if "0" is the first number we see in a sequence, we should treat it as a motion, because e.g. "d010j" is meaningless - you could just write "d10j". There will never be a case where "0" is *both* the first digit in a sequence *and* has meaning as a number, so when we see a "0" at the beginning of a sequence we're always safe to treat it as a motion instead of part of a count argument.

The variable I added in the *last* PR would allow this behavior with other digits: e.g. "d1" could be a valid and complete command, but you'd also still be able to have "d31j" do as you expect. This comes with the caveat that you *wouldn't* be able to do "d15j" for instance.

In hindsight, this really doesn't make much sense. People do wild stuff with Emacs - I'm sure there's *someone* who would put that facility to good use. However, I imagine it happens infrequently enough that we shouldn't make a special variable just to handle that case, and summarily force people that want to bind functions to "0" to discover and mess with a custom variable, just to accommodate that one weird graybeard.

This PR (instead of checking for a function's inclusion in a list of motions bound to digit keys) simply checks if an entered key is "0" and decides intelligently if its corresponding function should be treated as a digit argument or as a motion. This means that you can bind motions to 0 completely transparently, just like with any other key - no need for weird `evil-redirect-digit-argument` shenanigans (which didn't work), or the additional overhead of adding the function to a custom variable (blech).

In summary, as far as binding stuff to digits goes:

| Strategy | Does it work? | Allows binding to digits besides 0? | 0 motions defined the same as other keybinds? |
| --- | --- | --- | --- |
| `evil-redirect-digit-argument` | :x: | :x: | :x: |
| `evil-digit-bound-motions` | :heavy_check_mark: | :heavy_check_mark: | :x: |
| This PR | :heavy_check_mark: | :x: | :heavy_check_mark: |

### DIFF
--- a/evil-common.el
+++ b/evil-common.el
@@ -555,7 +555,7 @@ Both COUNT and CMD may be nil."
                                    (list (car cmd) (* (or count 1)
                                                       (or (cadr cmd) 1))))))))
                ((or (eq cmd #'digit-argument)
-                    (and (memq cmd evil-digit-bound-motions)
+                    (and (equal seq "0")
                          count))
                 (let* ((event (aref seq (- (length seq) 1)))
                        (char (or (when (characterp event) event)
@@ -723,7 +723,8 @@ recursively."
            ((functionp cmd)
             (if (or (memq cmd '(digit-argument negative-argument))
                     (and found-prefix
-                         (memq cmd evil-digit-bound-motions)))
+                         (or (equal (substring keys beg end) "0")
+                             (equal (substring keys beg end) [48]))))
                 ;; skip those commands
                 (setq found-prefix t ; found at least one prefix argument
                       beg end
--- a/evil-common.el
+++ b/evil-common.el
@@ -724,7 +724,7 @@ recursively."
            ((functionp cmd)
             (if (or (memq cmd '(digit-argument negative-argument))
                     (and found-prefix
-                         (equal (vconcat seq) [48])))
+                         (equal (vconcat seq) (vector ?0))))
                 ;; skip those commands
                 (setq found-prefix t ; found at least one prefix argument
                       beg end
--- a/evil-common.el
+++ b/evil-common.el
@@ -710,10 +710,11 @@ recursively."
            (end 1)
            (found-prefix nil))
       (while (and (<= end len))
-        (let ((cmd (key-binding (substring keys beg end))))
+        (let* ((seq (substring keys beg end))
+               (cmd (key-binding seq)))
           (cond
            ((memq cmd '(undefined nil))
-            (user-error "No command bound to %s" (substring keys beg end)))
+            (user-error "No command bound to %s" seq))
            ((arrayp cmd) ; keyboard macro, replace command with macro
             (setq keys (vconcat (substring keys 0 beg)
                                 cmd
@@ -723,8 +724,7 @@ recursively."
            ((functionp cmd)
             (if (or (memq cmd '(digit-argument negative-argument))
                     (and found-prefix
-                         (or (equal (substring keys beg end) "0")
-                             (equal (substring keys beg end) [48]))))
+                         (equal (vconcat seq) [48])))
                 ;; skip those commands
                 (setq found-prefix t ; found at least one prefix argument
                       beg end
@@ -735,7 +735,7 @@ recursively."
                              (string-to-number
                               (concat (substring keys 0 beg))))
                            cmd
-                           (substring keys beg end)
+                           seq
                            (when (< end len)
                              (substring keys end))))))
            (t ; append a further event
--- a/evil-integration.el
+++ b/evil-integration.el
@@ -476,7 +476,6 @@ Based on `evil-enclose-ace-jump-for-motion'."
 
 ;; visual-line-mode integration
 (when evil-respect-visual-line-mode
-  (customize-set-variable 'evil-digit-bound-motions '(evil-beginning-of-visual-line))
   (evil-define-minor-mode-key 'motion 'visual-line-mode
     "j" 'evil-next-visual-line
     "gj" 'evil-next-line
--- a/evil-vars.el
+++ b/evil-vars.el
@@ -611,15 +611,6 @@ in insert state."
   :type  'boolean
   :group 'evil)
 
-(defcustom evil-digit-bound-motions
-  '(evil-beginning-of-line)
-  "The motion commands that can be bound to some digit key (typically 0).
-While Evil is reading a motion command, functions in this list act as themselves
-if their corresponding key was the first digit in the key sequence, and behave
-like `digit-argument' for the purposes of `evil-keypress-parser' otherwise."
-  :type '(repeat function)
-  :group 'evil)
-
 (defvar dabbrev-search-these-buffers-only)
 (defvar dabbrev-case-distinction)
 (defcustom evil-complete-next-func


tomdl89:
@jmmathena thanks for your work on this - it's great to see you've given it so much thought. At the risk of annoying one or two (but most likely zero) very keen users, I think we're quick enough that obsoleting so soon after adding is probably a bit too proper. I think we should just remove it and apologize should anyone open an issue complaining. I'll take a look soon.

evil-common.el
```diff
@@ -723,7 +724,7 @@ recursively."
            ((functionp cmd)
             (if (or (memq cmd '(digit-argument negative-argument))
                     (and found-prefix
-                         (memq cmd evil-digit-bound-motions)))
+                         (equal (vconcat seq) [48])))
```
tomdl89, line 727: Why can't this just be:
```suggestion
                         (string= seq "0")))
```

jmmathena: @tomdl89 `evil-extract-count` expects either a string or a vector. If you just use `string=`, it'll fail in cases like [this](https://github.com/emacs-evil/evil/blob/17c635f6e1f538bf4b2c3c276ddd9d4d165a52fb/evil-tests.el#L8617), since `string=` chokes on vectors.

To deal with this, my original solution was to deal with the string and vector cases separately:

```
((functionp cmd)
            (if (or (memq cmd '(digit-argument negative-argument))
                    (and found-prefix
                       (or (equal seq [48])
                            (equal seq "0")))) ;;use equal instead of string= here; if seq is a vector != [48], it'll choke
```

However, since any string can be represented by a vector (the converse is not true - see [M-right], above), we can just always convert `seq` to a vector with `vconcat` and compare it with the character code 48, and avoid having to add an extraneous `or` expression.

tomdl89: OK makes sense. I think `(vector ?0)` or `(vconcat "0")` may be more readable than `[48]` but otherwise all good.

Somelauw:
I've just accepted the former change to `evil-digit-bound-motions` in evil-org.
I agree with this PR, though it would temporarily break evil-org again if merged.

jmmathena:
Hey @Somelauw! Appreciate the updates to evil-org-mode. I've just submitted a PR that'll fix evil-org-mode if this PR ends up getting merged. I've been using the commit from this PR for a few weeks now locally, and forked evil-org-mode in the meantime so everything would be copacetic, so I can verify that the change has been working for me. It's a super simple one line edit, but I figured that if I'm breaking it, I should fix it.

tomdl89:
OK I've finally found the time to review this and it's all good - happy to merge. @Somelauw would you like to coordinate merges? If so, I'm UTC+1 so non-nightime hours should be fine. No worries if not.

Somelauw:
Sure, Wednesday 21:30 in UTC+1 would work for me.
Otherwise, Thursday at 20:00.

tomdl89:
@Somelauw great I'll merge Wednesday 21.00 CET

PR #1536, opened 2021-11-06
Add option of stop updating X PRIMARY selection with the current visual region

redguardtoo:
Vim user can select a region in visual mode and press the key `p` to
replace the selected text.

But in GUI Emacs, the selected text won't be replaced because the commands
in visual state automatically insert the selected text into the X selection.
Press key `p` or run `evil-paste-after` only copies from the X selection which
is same as the selected text.

Console Emacs does not have this problem.

### DIFF
--- a/doc/docstringdb.json
+++ b/doc/docstringdb.json
@@ -17052,5 +17052,16 @@
     "functionp": true,
     "macrop": null,
     "keymap-inv": null
+  },
+  "evil-visual-update-x-selection-p": {
+    "default": "t",
+    "local": false,
+    "default-type": "symbol",
+    "var-docstring": "Whether to update the X PRIMARY selection with the current visual region automatically.",
+    "fn-docstring": null,
+    "arglist": true,
+    "functionp": null,
+    "macrop": null,
+    "keymap-inv": null
   }
 }
--- a/doc/docstringdb.json
+++ b/doc/docstringdb.json
@@ -17052,16 +17052,5 @@
     "functionp": true,
     "macrop": null,
     "keymap-inv": null
-  },
-  "evil-visual-update-x-selection-p": {
-    "default": "t",
-    "local": false,
-    "default-type": "symbol",
-    "var-docstring": "Whether to update the X PRIMARY selection with the current visual region automatically.",
-    "fn-docstring": null,
-    "arglist": true,
-    "functionp": null,
-    "macrop": null,
-    "keymap-inv": null
   }
 }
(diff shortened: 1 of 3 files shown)


tomdl89:
Thanks for the PR @redguardtoo, but I'm not able to replicate the problem. When I visually select some text and press `p`, the text is replaced. Can you replicate this in a `make emacs` session? If so, can you give steps for me to replicate?

redguardtoo:
@tomdl89 , thanks for the review. 
I dig my code a bit. It seems `(setq select-enable-primary t)` is required to reproduce the issue.

`evil-paste-after` => `current-kill` => `interprogram-paste-function` => `gui-selection-value`. 
`gui-selection-value' returns clipboard text from CLIPBOARD or "PRIMARY" clipboard which are also controlled by `select-enable-clipboard` (it's `t` by default) and `select-enable-primary` (it's `nil` by default).

tomdl89:
@redguardtoo why don't you just set `select-enable-primary` to `nil`? Presumably the reason you set it to `t` is that you want to copy into the primary selection when you select text? You can always use the `+` register to force use of the clipboard. For example: yank some text; visually select the text you want to paste over; do `"+p` to paste the text you just yanked.

redguardtoo:
@tomdl89 , 
yes, change `select-enable-primary` to `nil` could solve my problem.

But as I analysed, people could have same problem even after set this this flag to `nil`.

For example, "xclip.el" could set `interprogram-paste-function` to its own function. Here is bug reported by another user. https://github.com/spudlyo/clipetty/issues/8

I think it's better to give users an option to turn off the updating X PRIMARY selection automatically. For example, I have my own clipboard manager. Right now pressing `vw` or `ve` in GUI Emacs will automatically insert the selected text into my clipboard managers, that's too much noise for my clipboard manager.

Besides, some Linux clipboard manager could synchronise  PRIMARY and CLIPBOARD selections, so I got same issue even I set `select-enable-primary` to `nil`,
- `evil-visual-update-x-selection` set PRIMARY selection
- clipboard manager copy the content into CLIPBOARD selection
- `evil-paste-after` calls `current-kill` which read from CLIPBOARD selection.

tomdl89:
@redguardtoo thanks for explaining. I agree an option to disable is a good idea. It seems to reflect vim's [clipboard](http://vimdoc.sourceforge.net/htmldoc/options.html#'clipboard') or [guioptions](http://vimdoc.sourceforge.net/htmldoc/options.html#'guioptions') option "autoselect" from what I can see. I will merge shortly - just want to take a look at a tidy-up beforehand, as the when-let-when-when nesting is starting to make my eyes water ;)

tomdl89:
@redguardtoo can you check my refactoring commit? I briefly checked it still worked, but if you could double-check, that'd be great. re docs generation, they're out of date anyway, so I'll do that separately, as another PR

redguardtoo:
@tomdl89 , looks good. Thanks for reviewing and fixing my code.

PR #1542, opened 2021-11-17
Fix #1541: Advice point for kill-sexp in normal/visual mode

### LINKED ISSUE #1541: kill-sexp missing point advice in normal/visual mode
# Issue type

- Bug report

# Reproduction steps
- Start Emacs
- in scratch buffer type in new line:
- `(+ 1 1)`
- Move point to end of the line (.e.g. `$`)
- Use `backward-kill-sexp` in normal mode

# Expected behavior
The whole line should be removed as it does when in **insert** mode.

# Actual behavior
Result: `(+ 1 )`

# Further notes
This looks to me that the underlying `kill-sexp` needs advising as is done with other functions already: `(advice-add 'elisp--preceding-sexp :around 'evil--preceding-sexp '((name . evil)))`.

knusprjg:
Added advice for `kill-sexp` based on the existing functions. 

That fix works fine for me a while already.

### DIFF
--- a/evil-integration.el
+++ b/evil-integration.el
@@ -260,6 +260,7 @@
       (apply command args)))
 
   (advice-add 'elisp--preceding-sexp :around 'evil--preceding-sexp '((name . evil)))
+  (advice-add 'kill-sexp             :around 'evil--preceding-sexp '((name . evil)))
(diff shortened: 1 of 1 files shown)


knusprjg:
Closing this pull, see discusion in #1541.

PR #1554, opened 2021-12-14
Make `evil-record-macro' add the same macro to the kmacro ring

tttuuu888:
Sorry, but this is the same as #1064.
#1064 was forked from a very old code, so I opened a new pull request while updating it, so please take a look.

This allows kmacro to edit macros independently of the original evil macro.

Update is,
1. Add the same macro recorded as `evil-record-macro` to the kmacro ring.
2. Make the added kmacro editable

This simply adds the same macro to the kmacro ring and allows the user to edit it(without touching original evil macro). So I believe this update will not break the workflow for existing evil users.

Thank you.

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -2276,7 +2276,7 @@ The return value is the yanked text."
   "The buffer that has been active on macro recording.")
 
 (defun evil-end-and-return-macro ()
-  "Like `end-kbd-macro' but also return the macro.
+  "Like `kmacro-end-macro' but also return the macro.
 Remove \\<evil-insert-state-map>\\[evil-execute-in-normal-state] from the end."
   (kmacro-end-macro nil)
   (let ((end-keys-seq (append evil-execute-normal-keys nil))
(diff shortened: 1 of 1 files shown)


tomdl89:
Thanks for the PR @tttuuu888  - I'll take a proper look when I can

tomdl89:
Looks good. Before merging, I'd like to wait for a reply from @bigodel on #1064 if possible.

bigodel:
> Looks good. Before merging, I'd like to wait for a reply from @bigodel on #1064 if possible.

hey, i am so very sorry about the delay. its the end of the year, things get a little crazier than usual... i was suspecting that this change was making it so that macros some stored on registers to appear as "Garbage:" in `consult-registers`, but it turns out that that is the case even without this modification, so i suppose it shouldn't break anything, given that `kmacro-end-macro` calls `end-kbd-macro` anyways inside it. i've also been using this since then and it appears to me that it works flawlessly.

PR #1566, opened 2022-01-06
evil-find-file-at-point-with-line | support column number too

eeshugerman:
Add support for the `<filename>:<line>:<column>` format. Previously, given this format, `column` was treated as line number and `line` was ignored.

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -3546,21 +3546,28 @@ If FORCE is non-nil and MARKS is blank, all local marks except 0-9 are removed."
 
 (eval-when-compile (require 'ffap))
 (evil-define-command evil-find-file-at-point-with-line ()
-  "Opens the file at point and goes to line-number."
+  "Opens the file at point and goes to position if present."
   (require 'ffap)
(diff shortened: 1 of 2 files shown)


tomdl89:
Thanks for the PR @eeshugerman - is this functionality present in vim/neovim? Or proposing as a bonus feature for evil? I'm not opposed if the latter, but I think it's worth adding a test (look at `evil-with-temp-file`'s use in `evil-test-find-file` for guidance) either way.

eeshugerman:
I just tried it out in Neovim. It seems this functionality is not present there, but also it works differently than Evil does without this patch: In Neovim, the first number is treated as the line number and the second number is ignored. In other words, Neovim doesn't break on the `<filename>:<line>:<column>` format (as does Evil w/o this patch), but it doesn't fully support column number (as does Evil w/ this patch).

I'll look into adding tests -- thanks for the tip!

eeshugerman:
Hi @tomdl89, I've added a couple tests -- PTAL.

PR #1574, opened 2022-02-01
Fix beginning/end-of-current-sentence marker

leungbk:
* evil-vars.el (evil-markers-alist): Adjust to new name


tomdl89:
Thanks @leungbk!

PR #1580, opened 2022-02-08
Ex bar

tomdl89:


tomdl89:
@leungbk if you have time to take a look and find out what does/doesn't work, that'd be very much appreciated. Eventually I'd like to allow a single unambiguous bar as the separator, and to let it work for :global and others, but this is a first step.

leungbk:
@tomdl89 Thanks for working on this! I'm not the most advanced Vim user, so I've done only basic operations with your branch. But from my brief testing, I haven't noticed anything off.

Maybe you could try soliciting feedback on r/emacs? There are probably some skilled Evil users there.

tomdl89:
Thanks @leungbk. Yes I probably should ask r/emacs - thanks for the tip. I think I'll give it more thought first - looking at vim's source code, I can see every single ex command is marked as either supporting or not supporting trailing bar. So potentially lots more work to do.

PR #1608, opened 2022-04-22
Add `evil-v$-excludes-newline`

### LINKED ISSUE #897: Add option for `v$` to exclude new line?

tomdl89:
Deprecate `evil-want-visual-char-semi-exclusive`.
Fixes #897

### DIFF
--- a/doc/source/settings.rst
+++ b/doc/source/settings.rst
@@ -116,6 +116,8 @@ of them can be tweaked using the following variables.
 
 .. elisp:autovariable:: evil-move-beyond-eol
 
+.. elisp:autovariable:: evil-v$-excludes-newline
+
 .. elisp:autovariable:: evil-cross-lines
 
 .. elisp:autovariable:: evil-respect-visual-line-mode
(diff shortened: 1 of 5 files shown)


mpereira:
Sorry for the late reply here, but thanks for getting this done!

PR #1609, opened 2022-04-23
fix C-h behavior when evil-want-C-h-delete is non-nil

ksladowski:
C-h is behaving as C-u behaves, seemed like a simple typo; behaves as expected after a couple days of use.


tomdl89:
Yup simple typo. Thanks for the fix @ksladowski

PR #1611, opened 2022-04-24
Add `evil-word-object` & `evil-WORD-object` which stay on line

### LINKED ISSUE #834: diw kills above line's line ending
# Issue type
- Bug report

# Environment

Emacs version: GNU Emacs 25.2.1 (x86_64-apple-darwin16.4.0, NS appkit-1504.81 Version 10.12.3 (Build 16D32)) of 2017-03-11

Operating System:  macOS Sierra.

Evil version: 1.2.12
Evil installation type: MELPA through use package.
Graphical/Terminal: Graphical
Terminal multiplexer: does not apply. 

# Reproduction steps
- Start Emacs
- open any blank buffer in text-mode.
- now type exactly this lines:
- cursor is in []. 
```txt
foo       
    []        bar
```
- notice that on the second line there is an arbitrary number of spaces before the `bar` word.
- now, with the cursor placed on top of any of the spaces before the `bar` word type `diw`.

# Expected behavior
the buffer should now display:
```txt
foo
bar
```

# Actual behavior
the buffer instead displays:
```txt
foobar
```

# Further notes

vim behaves as expected. emacs eats the above line's line ending.

tomdl89:
Fixes #834

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -1277,7 +1277,7 @@ or line COUNT to the top of the window."
 
 (evil-define-text-object evil-inner-word (count &optional beg end type)
   "Select inner word."
-  (evil-select-inner-object 'evil-word beg end type count))
+  (evil-select-inner-object 'evil-word-object beg end type count))
 
 (evil-define-text-object evil-a-WORD (count &optional beg end type)
   "Select a WORD."
@@ -1285,7 +1285,7 @@ or line COUNT to the top of the window."
 
 (evil-define-text-object evil-inner-WORD (count &optional beg end type)
   "Select inner WORD."
-  (evil-select-inner-object 'evil-WORD beg end type count))
+  (evil-select-inner-object 'evil-WORD-object beg end type count))
 
 (evil-define-text-object evil-a-symbol (count &optional beg end type)
   "Select a symbol."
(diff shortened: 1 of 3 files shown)


Dickby:
That broke something. Now WORD will cross line boundries, and `evil-cross-lines` set to nil seems to have no effect anymore.

tomdl89:
Thanks for letting me know @Dickby - I'll look to get a fix in asap.

tomdl89:
@Dickby I can't see any buggy behaviour here. `evil-cross-lines` is only meant to affect the keys listed in its docstring; by default
`h` `l` `f` `F` `t` `T` `~`. If you observe otherwise, please provide some concrete steps for replication. Thanks!

Dickby:
I see, i thought 'w' was effected by evil-cross-lines too, sorry.
But it still is the case that if the cursor is on the last word of one line that has no trailing spaces, and the next line has no leading spaces, `diW` will delete the first part of the second line too. That bug was introduced by this commit.
I tested this with `make emacs` in the source directory

tomdl89:
No problem - thanks for checking. WRT to the diW bug on the last word of the line, I can replicate that. Working on a fix.

tomdl89:
I decided to revert the previous approach in favour of restricting evil-select-inner-object to the current line. @Dickby please take a look and let me know of any issues you find. Thanks again for checking.

PR #1613, opened 2022-04-30
Fix up evil-indent

rbrtb:


tomdl89:
Thanks for the PR. I'll need a description or (preferably) tests before I can merge @rbrtb

rbrtb:
Test added.

tomdl89:
Thanks for the test @rbrtb but that test passes _without_ your code change. I don't understand what your PR is trying to _fix_ so either I need a test which fails before and passes afterwards, or an explanation to that effect.

rbrtb:
It seems that the test cannot pass without the change on my machine though (Emacs 28):
```
F evil-test-indent
    Test ‘evil-indent’
    (args-out-of-range 36 1 32)
```
And this PR is trying to fix that, when the whole buffer is selected for `evil-indent`, the `indent-region` call may shrink the buffer so that the previous saved `end` is larger than `(point-max)`

rbrtb:
I noticed my original test won't pass even after the change in `make test` though.(It did pass in my Emacs setup)
So it's updated.

rbrtb:
Is it ok to merge this? @tomdl89

tomdl89:
Yup - I understand now. Thanks for the PR @rbrtb 🚀

PR #1616, opened 2022-05-01
Improve Makefile

rbrtb:
Make sure byte-compiled files are cleaned up before running tests.

Avoid running tests on stale code.


tomdl89:
@rbrtb all tests except 24.5 must pass on CI before this can be merged.

rbrtb:
It seems set `load-prefer-newer` would be a simpler approach, I'll open another PR to do that instead.

PR #1625, opened 2022-05-18
update evil-quit to handle tabs

### LINKED ISSUE #1624: :q on last window in tab kills emacs
# Issue type

- Bug report

# Environment

Emacs version: 28.1
Operating System: Arch
Evil version: evil-git-1fde8719e
Evil installation type: Not sure, spacemacs
Graphical/Terminal: GUI

# Reproduction steps
- Start Emacs
- Create at least one new tab with `tab-new`
- Run `:q` in a tab

# Expected behavior

Close tab

# Actual behavior

Closes emacs

# Further notes

I believe the issue is in the `evil-quit` command. The following modified version of that command solves the issue I am describing by attempting to close a tab (`tab-bar-close-tab`) before falling back to killing emacs.

```elisp
(evil-define-command evil-quit (&optional force)
  "Closes the current window, current frame, current tab, Emacs.
If the current frame belongs to some client the client connection
is closed."
  :repeat nil
  (interactive "<!>")
  (condition-case nil
      (delete-window)
    (error
     (if (and (boundp 'server-buffer-clients)
              (fboundp 'server-edit)
              (fboundp 'server-buffer-done)
              server-buffer-clients)
         (if force
             (server-buffer-done (current-buffer))
           (server-edit))
       (condition-case nil
           (delete-frame)
         (error
          (condition-case nil
              (tab-bar-close-tab)
            (error
             (if force
                 (kill-emacs)
               (save-buffers-kill-emacs))
             ))
          ))))
    ))
```

JoshMcguigan:
Closes #1624 

This PR updates `evil-quit` to properly handle tabs.

I think `tab-bar-close-tab` woud have been introduced in emacs 27, while evil supports emacs back to version 24. However, in this case we are already handling errors so this solution is still compatible with those versions. I couldn't come up with a clean way to use an `fboundp` here, and given it isn't strictly needed I omitted it, but I'm happy to take suggestions if there is a cleaner way to accomplish this.

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -3329,7 +3329,7 @@ for the last window in each frame."
             wins))))
 
 (evil-define-command evil-quit (&optional force)
-  "Closes the current window, current frame, Emacs.
+  "Closes the current window, current frame, current tab, Emacs.
 If the current frame belongs to some client the client connection
 is closed."
   :repeat nil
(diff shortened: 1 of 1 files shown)


tomdl89:
Thanks for the PR! Agree `fboundp` doesn't fit here.

PR #1633, opened 2022-06-29
Restore `evil-select-inner-object` API

Hi-Angel:
Before commit

    04b25f6 Add `evil-select-inner-unrestricted-object`

evil-select-inner-object was used for unrestricted selection. 04b25f6
made this function restrict selection to a line, and introduced a new
function evil-select-inner-unrestricted-object that works same way as
evil-select-inner-object before.

But evil-select-inner-object is an API that might be used by people in
custom configuration, and changing behavior of the function would result
in silent breakage.

It is also worth noting that documentation at doc/source/extension.rst
and doc/build/texinfo/evil.texi still mentions the older unrestricted
behavior.

So let's restore the original behavior of evil-select-inner-object by:

1. Renaming evil-select-inner-object → evil-select-inner-restricted-object
2. Renaming evil-select-inner-unrestricted-object → evil-select-inner-object

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -1280,7 +1280,7 @@ or line COUNT to the top of the window."
 
 (evil-define-text-object evil-inner-word (count &optional beg end type)
   "Select inner word."
-  (evil-select-inner-object 'evil-word beg end type count))
+  (evil-select-inner-restricted-object 'evil-word beg end type count))
 
 (evil-define-text-object evil-a-WORD (count &optional beg end type)
   "Select a WORD."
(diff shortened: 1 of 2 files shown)


Hi-Angel:
I presume the CI failure is unrelated — I see similar failures on other PRs as well

Hi-Angel:
CC: @tomdl89

tomdl89:
Thanks @Hi-Angel.
What about `evil-select-an-unrestricted-object`? seems like this should be changed for consistency.

Hi-Angel:
Thanks! Sure, will do

Hi-Angel:
Done!

Hi-Angel:
♥

PR #1640, opened 2022-07-21
Fix daw deleting indentation when deleting the only word in a line

bcc32:
In vim, deleting an indented word when it's the only word on a line
will just not delete any whitespace.  This applies specifically to
word objects in vim (see
https://github.com/vim/vim/blob/c7bd2f08e531f08723cdc677212a3633d11c9a97/src/textobject.c#L809),
and not any other objects.


tomdl89:
Thanks for the PR!

PR #1653, opened 2022-08-12
Add variable to prevent evil-change from inserting into kill-ring

NicholasBHubbard:
This PR adds a variable `evil-change-prevent-kill-ring` that if non-nil causes the text deleted from a call to `evil-change` to be stored in the void register by default.

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -1708,7 +1708,8 @@ If TYPE is `line', insertion starts on an empty line.
 If TYPE is `block', the inserted text in inserted at each line
 of the block."
   (interactive "<R><x><y>")
-  (let ((delete-func (or delete-func #'evil-delete))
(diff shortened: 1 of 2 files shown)


tomdl89:
Thanks for the PR @NicholasBHubbard. Does this emulate a setting in vim? Have you considered advising evil-change? Something like this (only v briefly tested) could work:
```
(define-advice evil-change (:around (fn beg end type reg yh df) reg-or-black-hole)
  (funcall fn beg end type (or reg ?_) yh df))
```

NicholasBHubbard:
No it doesn't emulate a setting in vim (that I'm aware of). I use similar advise to what you provided in my own config, I just thought it may be useful for others to have the functionality available as a variable setting. I understand if you don't think this PR aligns with the goals of evil-mode.

PR #1655, opened 2022-08-16
Make sure visual markers are up to date in evil-visual-range

justbur:
With composed characters in the buffer, such as with prettify-symbol-mode
active, emacs may adjust point after the post-command-hook is run. In this case,
evils visual markers do not get updated prior to say evil-delete and the deleted
region does not agree with what is seen visually (only part of the composed
characters are deleted). Forcing a last update using evil-visual-refresh
in evil-visual-range resolves this issue.

### DIFF
--- a/evil-states.el
+++ b/evil-states.el
@@ -743,6 +743,7 @@ This is a list (BEG END TYPE PROPERTIES...), where BEG is the
 beginning of the selection, END is the end of the selection,
 TYPE is the selection's type, and PROPERTIES is a property list
 of miscellaneous selection attributes."
+  (evil-visual-refresh)
   (apply #'evil-range
          evil-visual-beginning evil-visual-end
          (evil-visual-type)


justbur:
Here's an example. Put a buffer in `latex-mode` using auctex and add the text

```
$ \alpha \beta $
```

Then turn on `prettify-symbols-mode` which visually shows the associated unicode characters. Then try to delete just the `\alpha` with visual mode. Start with point on the first dollar sign and use `vllld`. This will only delete part of the underlying `\alpha` string, because there is an update to the location of point that is not captured by `evil-visual-refresh`. [More info here](https://www.gnu.org/software/emacs/manual/html_node/elisp/Adjusting-Point.html).

tomdl89:
@justbur weirdly this is causing some bugs I've noticed recently. Bisected my way to this commit [#7133220](https://github.com/emacs-evil/evil/commit/71332209386e431ae13442f20468b2edb59f727b).
Doing `viw` on a word, moving to another line and then doing`'>` no longer works after this commit.
I'll take a look soon if you don't beat me to it.

PR #1656, opened 2022-08-16
[WIP] Account for composed characters in evil-types.el

justbur:
Previously the next position in an evil-expand operation was assumed to be one
position to the right (or left), but this position might be invalid if there are
composed characters in the buffer. This can make it difficult to delete a
composed character with vd for example, because the visual range is not expanded
properly to account for the composition.

This commit adds two new functions, evil-next-adjusted-pos and
evil-prev-adjusted-pos, which are aware of composed characters. The new
functions are then used in evil-types.el.


justbur:
I realized there was more to the story around composed characters. I think this fixes many of the remaining issues (try to `vd` on a composed character like one in `prettify-symbols-mode` for example). 

It looks like the tests are failing though, so it'll need to be revised...

tomdl89:
@justbur is this ready to merge? lgtm but a test would be nice if it's easy to do.

justbur:
I got hung up on how far this should go. Here are different levels of intervention

1. Just try to take care of places that are surprising. This is where I started and what this goes after. It looks like you are selecting the whole composed character in visual mode, but then you only delete part of it with `d`. 
2. Treat positions inside composed characters as invalid. The emacs command loop does this, but there are a number of places in the evil code where the next/previous character is just assumed to be one position to the left or right. I can imagine making the `evil-normalize-` functions which are there to prevent movement into invalid positions aware of composed characters. 
3. Treat composed characters as single characters for the purpose of all evil commands. The one that comes to mind is `x`. This might automatically be taken care of if 2 is done correctly, but I'm not sure. 

Any thoughts?

justbur:
Opening a separate pull request with a different approach

PR #1667, opened 2022-08-29
Compatibility: use $SED instead of sed

windymelt:
Hi, I'm using FreeBSD, which has BSD `sed` program, not GNU `sed` . So in my environment `Makefile` doesn't work especially at line 5.

It's great if I could pass actual sed program to use like `SED=gsed make` (especially when installing via package manager like `el-get`) . So I made it variable. 

I think this change doesn't break current behavior.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,8 @@
 SHELL = /bin/sh
 EMACS ?= emacs
+SED ?= sed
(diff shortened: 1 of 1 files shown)


tomdl89:
Thanks for the PR!

PR #1673, opened 2022-09-18
evil-select-parens selects next pair if not in parens

Garklein:
In modern versions of Vim (and Neovim), doing an action on something in parens will jump to the next instance of those parens and perform the action on them if you're not in those parens.

For example, if you are on the line `int main(int argc, char **argv) {` and type `0cib`, it will change everything inside the parens even though you initially went to the beginning of the line. This pull request adds this functionality.

This feature already works on strings.

One bug is that if you yank something in parens and it needs to jump to the next pair because you're not in parens, in Vim it moves your cursor to the beginning of the next pair, while here it doesn't move your cursor. I couldn't figure out how to fix this.

### DIFF
--- a/evil-common.el
+++ b/evil-common.el
@@ -3396,6 +3396,9 @@ respectively. BEG END TYPE are the currently selected (visual)
 range.  If INCLUSIVE is non-nil, OPEN and CLOSE are included in
 the range; otherwise they are excluded.
 
+If you aren't inside a pair of the opening and closing delimiters,
+it jumps you inside the next one. If there isn't one, it errors.
+
 The types of OPEN and CLOSE specify which kind of THING is used
 for parsing with `evil-select-block'. If OPEN and CLOSE are
 characters `evil-up-paren' is used. Otherwise OPEN and CLOSE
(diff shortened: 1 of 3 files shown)


tomdl89:
Thanks for this! I'm in hols right now but will take a look in a couple of weeks if no one beats me to it.

tomdl89:
@Garklein before I take a proper look, will you add some tests? I can if you can't, but it may delay things a little.

Garklein:
Yep, here you go!

evil-common.el
```diff
@@ -3404,50 +3407,60 @@ must be regular expressions and `evil-up-block' is used.
...
+                                         inclusive
+                                         (or (< extbeg beg) (> extend end))
+                                         t)))))))
+         (t
+          (evil-select-block #'(lambda (&optional cnt)
+                                 (evil-up-block open close cnt))
+                             beg end type count inclusive))))
+    (error ; we aren't in the parens, so find next instance
+     (save-match-data
+       (goto-char (or (if (and count (> 0 count)) end beg) (point)))
```
tomdl89, line 3457: Can we have a line break between the end of the `if` expression and the rest of the `or` expression, for readability?

tomdl89:
@Garklein I put in some fixes. I decided I'm not interested in the edge case of `vi(` on empty parens 🤷 - take a look if you like - I'll merge if you're happy.

Garklein:
@tomdl89 Thanks for fixing that yank issue, I would never have figured that out!  
I found one odd edge case that I will try and fix (if either the opening or closing paren is commented out, it stack overflows) -- after that, I'm good with this merge.

tomdl89:
Oh good spot! Yes please try to take a look - sounds a bit tricky.

Garklein:
@marcolgl Thank you for your contributions and thorough explanation!
Everything looks good to me and I'm happy with this. @tomdl89 thoughts?

tomdl89:
Looks great. Thanks both! That's merged to master 🚀

PR #1677, opened 2022-09-27
Fix pasting from ?+ register in Emacs 29

eeshugerman:
I found that on Emacs v29, `" + p` does not paste from the system clipboard if the clipboard was populated outside emacs, rather it pastes from the kill ring.

It looks like this same bug briefly existed in the built-in `gui-selection-value` until it was fixed in [this commit](https://github.com/emacs-mirror/emacs/commit/2289fafeaf) by checking a new var `gui--last-cut-in-clipboard`.

It also looks like `gui--selection-value-internal` (previously `x-selection-value-internal`) has drifted from the evil code that's based on it.  Rather than attempt to bring the evil code up to date while keeping compatibility with Emacs 24, I opted for a version check so we can just call `gui--selection-value-internal` if it exists.

It's unclear to me why this code was duplicated in evil in the first place. It's entirely possible it was for a good and still-applicable reason, in which case merging my fix here would be problematic. Anyway, [here's the commit where it happened](https://github.com/eeshugerman/evil/commit/3841842) -- @f-fr, do you happen to recall?

### DIFF
--- a/evil-common.el
+++ b/evil-common.el
@@ -2148,7 +2148,7 @@ The following special registers are supported.
                      (current-kill reg t))))
              ((memq register '(?* ?+))
               (let ((what (if (eq register ?*) 'PRIMARY 'CLIPBOARD)))
-                (if (version<= "25" emacs-version)
+                (if (version<= "29" emacs-version)
                     (gui--selection-value-internal what)
                   ;; the following code is modified from `x-selection-value-internal'
                   (let ((request-type (or (and (boundp 'x-select-request-type)
--- a/evil-common.el
+++ b/evil-common.el
@@ -2150,7 +2150,10 @@ The following special registers are supported.
               (let ((what (if (eq register ?*) 'PRIMARY 'CLIPBOARD)))
                 (if (version<= "29" emacs-version)
                     (gui--selection-value-internal what)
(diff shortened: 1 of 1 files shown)


tomdl89:
Thanks for the PR. If you'd like to err on the safe side, we could change `(version<= "25" emacs-version)` to `(version<= "29" emacs-version)`, merge it, and give the bleeding edge folks a chance to report back? p.s. Frank hasn't been active on here for years, so don't be too optimistic about hearing from him.

eeshugerman:
👍 I like that idea -- will update this evening

f-fr:
Actually I did reply (by mail, I don't know why the answer doesn't appear hear ...). Anyway, that was my (not very helpful) answer:

I do not remember anymore, sorry. At that time we also tried to keep compatibility with Emacs 22, 23 or so, maybe that was the reason. I would suggest going ahead and see if it works with Emacs ≥ 26 or so.

eeshugerman:
Got it, thanks for the input Frank! To play it safe I went with just 29 for now.

smartepsh:
Thank you so much for the fix. I've been using simpleclip before this and it was very awkward.

dvzubarev:
I described a problem that is related to this fix in #1843

PR #1678, opened 2022-09-28
Fix Ex completion of empty string

axelf4:
An empty string will not parse as an Ex command, meaning `evil-ex-cmd` would be nil even after `evil-ex-update` returns. This pull request makes sure `evil-ex-command-completion-at-point` handles this.

Also remove the unnecessary nil-check in `evil-ex-binding`.

### DIFF
--- a/evil-ex.el
+++ b/evil-ex.el
@@ -284,13 +284,12 @@ in case of incomplete or unknown commands."
   (let* ((prompt (minibuffer-prompt-end))
          (string (or string (buffer-substring prompt (point-max))))
          arg bang cmd count expr func handler range tree type)
-    (cond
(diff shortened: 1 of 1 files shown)


axelf4:
> Is this easy to write a test for?

Not sure, to call `evil-ex-command-completion-at-point` the Ex minibuffer has to have been initialized. But to test manually:

* Run `make emacs`
* Type <kbd>:</kbd> followed by <kbd>Tab</kbd>
* With this patch the default completion UI shows up, without you get the following error message in the minibuffer:

      : [Args out of range: 0, 0]

tomdl89:
Yeah, sounds difficult to write a test for. Can confirm the patch fixes the bug. Thanks for the PR!

PR #1681, opened 2022-09-29
fix: replace deprecated functions

dalugm:


tomdl89:
Cool - interesting. Where do you get the impression these functions are deprecated? I can see that `point-at-bol` and `point-at-eol` and just aliases, but `lsh` and `ash` look like they do slightly different things, no?

dalugm:
> `point-at-bol` and `point-at-eol` and just aliases

https://git.sv.gnu.org/cgit/emacs.git/tree/lisp/subr.el?h=master#n1902

On Emacs 29, it's marked as obsolete.

>  `lsh` and `ash` look like they do slightly different things, no?

https://git.sv.gnu.org/cgit/emacs.git/tree/etc/NEWS?h=master#n2838

Since the NEWS says so, I guess we should follow it?

tomdl89:
Ah, TIL. Thanks for these!

Issue #618, opened 2016-02-08
evil-scroll-down cannot make it to the end of the file


TheBB:
Originally reported by: **Aaron Jensen (Bitbucket: [aaronjensen1](https://bitbucket.org/aaronjensen1), GitHub: Unknown)**

----------------------------------------

The new `C-d` works pretty well, but for some reason, it does not always make it to the end of the file. You get `End of buffer` when you're still about 5-6 lines from the bottom.

----------------------------------------
- Bitbucket: https://bitbucket.org/lyro/evil/issue/618

PR #1683, opened 2022-09-30
Fix up the cmds that scroll relative to cursor

axelf4:
Reverts commit bbe92d1 (Handle scroll margin in scroll-to-top/bottom functions (re #618), 2016-03-31) which introduced a bug in evil-scroll-bottom-line-to-top (due to recenter argument becoming negative with scroll-margin set to 0.) recenter itself has clamped to scroll_margin for at least 18 years now, so it is unclear why that commit was even necessary.

Also fixes the docstring for evil-scroll-top-line-to-bottom and avoids scanning down from (point-min) to the line the point is already at.

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -1133,33 +1133,33 @@ If the scroll count is zero the command scrolls half the screen."
   :repeat nil
   :keep-visual t
   (interactive "<c>")
-  (evil-save-column
-    (let ((line (or count (line-number-at-pos (point)))))
+  (when count
+    (evil-save-column
       (goto-char (point-min))
-      (forward-line (1- line)))
-    (recenter (1- (max 1 scroll-margin)))))
+      (forward-line (1- count))))
+  (recenter 0))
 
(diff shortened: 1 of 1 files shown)


tomdl89:
Thanks for the PR. Could you either add tests or give exact instructions on how to manually test? (I'll ask this for _nearly all_ PRs without tests btw)

axelf4:
> Could you either add tests or give exact instructions on how to manually test?

For regression:

- Enter buffer with more lines than the window height.
- Ensure `scroll-margin` is zero.
- Execute Normal mode command `z+`.
- Marvel at the fact you are the 2nd person to do so in >6 years, since had anyone else done it they'd become very confused indeed by it scrolling the line below the window to the bottom of the window instead of to the top, and immediately come scrambling to this very issue tracker to file a bug report. This patch fixes that.

For the other commands, `zb`; `zz` and `zt`, this PR just cleans up their code and I have verified that they all behave the same as before with different values of `scroll-margin`, which is easy to do.

PR #1684, opened 2022-09-30
Fix logical/visual line move followed by the other

### LINKED ISSUE #1469: Alternating between visual line mode aware commands and regular commands breaks the visual line mode aware commands

axelf4:
This pull request does not have the issue that #1470 had: The workaround is only applied at the boundaries of logical and visual line movements, which were the only places where things went awry.

I removed support for `next-line-add-newlines`, almost feel like it would be better to add that code to the `evil-next-line` and `evil-next-visual-line` functions. In any case, it feels wrong to have `evil-line-move` call `next-line`/`previous-line`.

No tests yet, maybe someone could help out with that?

Resolves #1469

### DIFF
--- a/evil-common.el
+++ b/evil-common.el
@@ -1400,32 +1400,23 @@ If STATE is given it used a parsing state at point."
 ;; the line (we never want point to leave its column). The code here
 ;; comes from simple.el, and I hope it will work in future.
 (defun evil-line-move (count &optional noerror)
-  "A wrapper for line motions which conserves the column.
+  "A wrapper for line motions that conserves the column.
 Signals an error at buffer boundaries unless NOERROR is non-nil."
-  (cond
-   (noerror
-    (condition-case nil
-        (evil-line-move count)
-      (error nil)))
-   (t
+  (setq this-command (if (< count 0) #'previous-line #'next-line))
+  (let ((last-command
+         (when (eq line-move-visual (consp temporary-goal-column))
+           ;; Reset tmp goal column between visual/logical movement
+           last-command))
+        (opoint (point)))
     (evil-signal-without-movement
-      (setq this-command (if (>= count 0)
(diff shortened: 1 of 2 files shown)


tomdl89:
Thanks for the PR. Could you either add tests or give exact instructions on how to manually test? (I'll ask this for _nearly all_ PRs without tests btw)

axelf4:
> Could you either add tests or give exact instructions on how to manually test?

Please see the added commit.

PR #1696, opened 2022-10-06
Fix evil-select-paren #1673 for regexp special chars

marcolgl:
Hi,
sorry for the scattered PRs, but i figured out a bug on a very common use case, that is jumping to next bracket.
When the open delimiter is a single character which has special behavior in regexps, the call `(string open)` gives the
char not escaped, and the following functions that use such regexp fail.
It causes the common case of commands on `evil-a/inner-bracket` textobjects to fail, for instance:
`|prova di [testo]`
`ci[`
gives evil-motion-range: Invalid regexp: "Unmatched [ or [^"
Now that should be fine, thanks for your consideration.

### DIFF
--- a/evil-common.el
+++ b/evil-common.el
@@ -3456,7 +3456,7 @@ is ignored."
      (save-match-data
        (goto-char (or (if (and count (> 0 count)) end beg)
                       (point)))
-       (let ((re (if (characterp open) (string open) open)))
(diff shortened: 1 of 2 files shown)


tomdl89:
No problem :) scattered PRs is considerably better than bugs remaining in the codebase!

PR #1702, opened 2022-10-13
Remove make-variable-buffer-local from variables (fixes #1416)

### LINKED ISSUE #1416: Why are evil-shift-width etc. forced buffer-local?
# Issue type

- Question

# Environment

Emacs version: GNU Emacs 27.1 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.22, cairo version 1.17.3) of 2020-08-28
Operating System: Manjaro Linux 20.2.1 Nibia
Evil version: 1.14.0
Evil installation type: straight.el pinned to cc9d6886b418389752a0591b9fcb270e83234cf9
Graphical/Terminal: Graphical

# Further notes

Why are the following variables (and maybe more) forced to be buffer-local using `make-variable-buffer-local`?

- `evil-auto-indent`
- `evil-shift-width`
- `evil-shift-round`
- `evil-bigword`
- `evil-symbol-word-search`

This seems to contradict the [relevant section of the Emacs Manual](https://www.gnu.org/software/emacs/manual/html_node/elisp/Creating-Buffer_002dLocal.html), which clearly states:

> **Warning:** Don’t assume that you should use `make-variable-buffer-local` for user-option variables, simply because users might want to customize them differently in different buffers. Users can make any variable local, when they wish to. It is better to leave the choice to them.
> The time to use `make-variable-buffer-local` is when it is crucial that no two buffers ever share the same binding. For example, when a variable is used for internal purposes in a Lisp program which depends on having separate values in separate buffers, then using `make-variable-buffer-local` can be the best solution.

This does not seem to be the case with these variables. This decision also causes some concrete issues for me. For example, I would like to put this in my `.emacs`:

```elisp
(defvaralias 'evil-shift-width 'standard-indent)
```

This, however, does not work because `evil-shift-width` has been passed to `make-variable-buffer-local`. I would therefore prefer the offending lines to be removed, if there is no valid reason to keep them.

Ghosty141:
Fixes #1416 and removes make-variable-buffer-local from other variables since it's generally not correct to use this function for user-options. It should be set by individual users who want this behavior.

### DIFF
--- a/evil-vars.el
+++ b/evil-vars.el
@@ -165,7 +165,6 @@ Whether to auto-indent when opening lines with \\[evil-open-below] \
 and \\[evil-open-above]."
   :type  'boolean
   :group 'evil)
-(make-variable-buffer-local 'evil-auto-indent)
 
 (defcustom evil-shift-width 4
   "\\<evil-normal-state-map>
@@ -174,7 +173,6 @@ This applies to the shifting operators \\[evil-shift-right] and \
 \\[evil-shift-left]."
   :type 'integer
   :group 'evil)
-(make-variable-buffer-local 'evil-shift-width)
 
 (defcustom evil-shift-round t
   "\\<evil-normal-state-map>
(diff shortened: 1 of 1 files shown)


tomdl89:
Thanks for the PR @Ghosty141

PR #1717, opened 2022-11-15
Simplify temporary-goal-column logic

axelf4:
In `evil-ensure-column` there is no point in normalizing `temporary-goal-column` a second time after BODY, since that variable is used to store the saved column that must not be overwritten. Also tracking EOL after `evil-end-of-line` is already handled as it fakes being a `next-line` command.

In `evil-apply-on-block` only the case of `temporary-goal-column` equaling `most-positive-fixnum` is relevant, as other column values would be contained inside BEG..END.

Adds a test for commit 7aa71576d258b9546238f6dbd4cf5ca0a1b70088.

### DIFF
--- a/evil-tests.el
+++ b/evil-tests.el
@@ -1630,7 +1630,12 @@ New Tex[t]
     :state visual
     "<    Line with too much indentation.>"
     ("=")
-    "Line with too much indentation."))
+    "Line with too much indentation.")
+  (ert-info ("Can repeat evil-indent without errors")
+    (evil-test-buffer
+     "[ ]x\n x"
+     ("==j.")
+     "x\nx")))
 
 (ert-deftest evil-test-keypress-parser ()
   "Test `evil-keypress-parser'"
--- a/evil-tests.el
+++ b/evil-tests.el
@@ -7274,7 +7274,14 @@ Tiny ln"
      "Short 
 A much
 A me[d]
-Tiny ")))
+Tiny "))
+  (ert-info ("Visual block does not extend up to curswant unless tracking EOL")
+    (evil-test-buffer
+     "[a]bc
+"
+     ("\C-vfcjd")
+     "[b]c
+")))
 
 (ert-deftest evil-test-visual-restore ()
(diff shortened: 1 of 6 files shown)


evil-tests.el
```diff
@@ -7260,7 +7265,14 @@ Tiny ln"
...
 A much
 A me[d]
-Tiny ")))
+Tiny "))
+  (ert-info ("Visual block does not extend up to curswant unless tracking EOL")
+    (evil-test-buffer
+     "[f]oo bar
+foo"
+     ("\C-vfrjd")
+     "[ ]bar
```
tomdl89, line 7274: I tried this in vim and see that the space before `bar` is also deleted. With this, just before I delete I see that the cursor is "beyond-eol" on the second line (which is fine), but the block selection stops at the end of foo, rather than on the space.

tomdl89:
Thanks for the PR. Other than the comment I made, this looks good to me.

evil-tests.el
axelf4: I think that is a separate issue where `evil-visual-range` results are the same beyond-eol and not, since you can replicate it even without this patch if you follow "j" with "hl" to make it ignore `temporary-goal-column`. 

Would need to change [this code](https://github.com/emacs-evil/evil/blob/5ce46a1fc175a8f13507ce2b6ec4c3618923f093/evil-types.el#L175-L208) to allow returning an inclusive range. Perhaps that could be delegated to a follow-up PR, since I do not know how you'd go about doing that, and since this PR at least does not make that matter worse.

tomdl89: Sure, a separate PR is fine. But we shouldn't add tests which assert incorrect behaviour. So either we remove the test or we fix that first, then merge this. It's a test I think we should have, so I'm inclined to favour the latter. I may have time to take a look this weekend if you don't beat me to it.

axelf4: I changed the test now so that the incorrect behavior is not exhibited, because I would be positively surprised if that ever gets fixed.

PR #1744, opened 2023-01-03
Stop the '[/'< and ']/'> marks from intertwining

axelf4:
The function evil-visual-paste would assign to evil-visual-point/-mark the same markers used by the '[ and '] marks. Therefore after a single visual paste, Normal mode "gv" in that buffer would act funkily ever after. This commit fixes that.

Also removes unused definitions related to evil-visual-previous-point/-mark.

To reproduce:
* Enter new buffer with content:
  ```
  x
  y
  ```
* With cursor on `y`, type `ylvpkxgv`.

The expectation is that `y` should be re-selected. Instead the empty upper line is selected due to it being the previously changed text.

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -2407,10 +2407,8 @@ leave the cursor just after the new text."
       (when evil-kill-on-visual-paste
         (current-kill -1))
       ;; Ensure that gv can restore visually pasted area...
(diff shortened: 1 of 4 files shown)


tomdl89:
Looks good. Can we add tests to these sort of PRs please.

PR #1753, opened 2023-01-14
Use original state for populating shortcut map

### LINKED ISSUE #1073: Incorrect operator shortcut keys when SPC used as prefix
# Issue type
- Bug report

# Reproduction steps
Bind `SPC <key>` to an operator and attempt to use the operator linewise with `SPC <key> <key>`

# Expected behavior
Operator should act on line.

# Actual behavior
`<key>` and `SPC <key>` are not bound correctly in `evil-operator-shortcut-map`. This is because `evil-extract-count` in  `evil-operator-range` returns `SPC` as the seq instead of `SPC <key>`, so `SPC` is bound to `evil-line` instead of the expected `SPC <key>` and `<key>`. I haven't looked further into the issue in `evil-extract-count`.

# Further notes
I couldn't find any other issues like this by searching but apologize if it is a duplicate.

Originally reported in noctuid/general.el#124
@aqua36 Follow this issue for updates.

axelf4:
Changing to Operator-pending state before executing

    (evil-extract-count (this-command-keys))

can give the wrong result since the actual binding used to invoke the operator may not be present in the new state.

Resolves #1073

### DIFF
--- a/evil-common.el
+++ b/evil-common.el
@@ -714,12 +714,12 @@ recursively."
            (beg 0)
            (end 1)
            (found-prefix nil))
-      (while (and (<= end len))
+      (while (<= end len)
         (let* ((seq (substring keys beg end))
                (cmd (key-binding seq)))
           (cond
            ((memq cmd '(undefined nil))
-            (user-error "No command bound to %s" seq))
+            (user-error "No command bound to `%s'" seq))
            ((arrayp cmd) ; keyboard macro, replace command with macro
             (setq keys (vconcat (substring keys 0 beg)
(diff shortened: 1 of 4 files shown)


aaronjensen:
This fix broke evil-surround as it relied on the second argument to `evil-operator-range`. Was that change essential to this fix?

axelf4:
No, sorry about that, I mistakenly assumed it was only used internally. Should be restored now.

aaronjensen:
No problem, thank you for the quick fix!

PR #1762, opened 2023-02-08
Compile Ex parser

axelf4:
See individual commits.

On my machine this pull request makes `make test` 40% faster, however the test suite not being byte- nor native compiled makes that figure less interesting.

### DIFF
--- a/evil-common.el
+++ b/evil-common.el
@@ -2022,18 +2022,17 @@ sometimes moving point, so `C-a' `C-w' etc. would miss their intended target.")
 
 (defun evil-ex-remove-default ()
   "Remove the default text shown in the ex minibuffer.
-When ex starts, the previous command is shown enclosed in
+When Ex starts, the previous command is shown enclosed in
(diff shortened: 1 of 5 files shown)


tomdl89:
Nice! As I don't have so much time these days to read through such a big PR, could you explain each test which has been removed, or if many have been removed for the same reason, outline that here?

axelf4:
Other than the two tests testing now-removed semantic actions not used elsewhere, only [these ones](https://github.com/emacs-evil/evil/blob/880613823a572d9294001b63468d8c39face5acf/evil-tests.el#L9026-L9059), for the reason that with `evil-parser` now requiring all productions be provided in the grammar up front, they would end up somewhat unreadable. Furthermore, most are still indirectly tested by `evil-test-ex-parse-ranges`.

The rest of the `evil-tests.el` diff is from merging three `range` productions. This way `line` does not get reparsed as much, at the low cost of complicating the expressions that get `eval`:ed, which only matters for the tests that compare the terms literally.

PR #1770, opened 2023-03-26
set :jump property to t for evil-edit

roxma:


axelf4:
Since the Evil version of `:edit` already adds to the jump list when switching buffers, and https://github.com/emacs-evil/evil/blob/de07e9a59231d280166c03e05407dcc1e5aed112/evil-jumps.el#L188-L189 ensures that e.g. `:e %<CR>` does not add a jump even with this patch, I presume this PR only makes sense after #1771 is merged? Ideally the behavior should be changed to match `:edit` in Vim where it unconditionally adds a jump. Otherwise, I still think this is a good change, however a note could be added, as I got confused by the lack of noticeable change by this PR. Or am I missing something?

roxma:
@axelf4  

I'm sorry. I don't remember why I changed this. I'm closing this PR. I'll re-open if this change is found necessary.

Please help me reviewing PR #1772  . I believe it fixes an existing issue, which is described in the PR.

PR #1771, opened 2023-03-26
Support evil--jumps-push for dired-mode

roxma:
1. Add `evil-jumps-current-file-name-function` that returns `dired-directory` in `evil--jumps-push` when dired-mode is active. The `dired-directory` can be properly re-opened by `find-file` called by future evil-jump-backward or evil-jump-forward
2. Set property :jump to t for #'dired-find-file #'dired-goto-file and #'dired-up-directory

### DIFF
--- a/evil-jumps.el
+++ b/evil-jumps.el
@@ -137,13 +137,18 @@ Otherwise the jump commands act only within the current buffer."
                                   (list pos file-name))))
                           (ring-elements (evil--jumps-get-window-jump-list))))))
 
+(defun evil--jumps-current-file-name ()
+  "Get the current buffer file name for `evil--jumps-push'."
+  (or buffer-file-name
+      (when (derived-mode-p 'dired-mode) default-directory)))
+
 (defun evil--jumps-jump (idx shift)
   (let ((target-list (evil--jumps-get-window-jump-list)))
     (evil--jumps-message "jumping from %s by %s" idx shift)
     (evil--jumps-message "target list = %s" target-list)
     (setq idx (+ idx shift))
-    (let* ((current-file-name (or (buffer-file-name) (buffer-name)))
-           (size (ring-length target-list)))
+    (let ((current-file-name (or (evil--jumps-current-file-name) (buffer-name)))
+          (size (ring-length target-list)))
       (unless evil-jumps-cross-buffers
         ;; skip jump marks pointing to other buffers
         (while (and (< idx size) (>= idx 0)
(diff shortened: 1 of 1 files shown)


tomdl89 (commented):
Looks good. I do think having `evil-jumps-current-file-name-function` as a variable may be overkill. I think hard-coding it for now is fine, then if we get people wanting to customize, we can add it in later.

roxma:
> I do think having evil-jumps-current-file-name-function as a variable may be overkill

@tomdl89 I want to make evil-jumps available for buffers of other modes. (e.g. magit mode). The vim-jumps of [vim-fugitive](https://github.com/tpope/vim-fugitive) works very well. I can jump back to a fugitive buffer even if the buffer has been deleted.

I think it is possible to implement this feature for evil and magit to customize `evil-jumps-current-file-name-function` and `find-file` fucntion.

evil-jumps.el
```diff
@@ -165,10 +165,16 @@ Otherwise the jump commands act only within the current buffer."
           (setf (evil-jumps-struct-idx (evil--jumps-get-current)) idx)
           (run-hooks 'evil-jumps-post-jump-hook))))))
 
+(defun evil--jumps-current-file-name ()
+  "Get the current buffer file name for evil--jumps-push"
+  (or (buffer-file-name)
```
axelf4, line 170: This should be
```elisp
(or buffer-file-name
    (when (derived-mode-p 'dired-mode) default-directory))
```
I think. In the cases where `dired-directory` differs from `default-directory` the code will break anyway.

Also, could this be inlined, or do you need it for adding a piece of advice on?

evil-keybindings.el
```diff
@@ -58,6 +58,11 @@
 
 (eval-after-load 'dired
   '(progn
+     ;; support evil-jumps
```
axelf4, line 61: This belongs in `evil-integration.el`, and could use `mapc`.

axelf4 (changes requested):

evil-keybindings.el
roxma: evil-jumps works fine for jumping around dired buffers without these lines. I don't know why it was needed last time. Maybe I was working on old version of evil that requires settings `:jump` for dired dired commands.

I have removed these lines. We may add :jump for dired later if it is found necessary.

axelf4 (approved):
I made sure `evil--jumps-jump` agrees on the buffer file name and squashed the commits. Could you check if things still look alright to you and if so I'll go ahead and merge this.

PR #1772, opened 2023-03-26
Fix: Should not (evil-set-jump previous-pos) jump after evil-jump-forward

roxma:
Steps to reproduce the issue:

  - Create files `a` `b` `c` with the following content:
    
```
        $ head a b c
        ==> a <==
        b
    
        I'm A
        ==> b <==
        c
    
        I'm B
        ==> c <==
        I'm C
 ```

  - Run `emancs a` to open bufer `a`
  - Press `gf` to jump to bufer `b`
  - Press `gf` to jump to bufer `c`
  - Press `C-o` to jump back to bufer `b`
  - Press `C-o` to jump back to bufer `a`
  - Press `C-i` to jump foward to bufer `b`
  - Press `C-i` to jump foward to bufer `c`, but the window stays on buffer `b` unexpectedly.

### DIFF
--- a/evil-jumps.el
+++ b/evil-jumps.el
@@ -62,7 +62,7 @@ Otherwise the jump commands act only within the current buffer."
 
 (defvar evil--jumps-jumping nil)
 
-(defvar evil--jumps-jumping-backward nil
+(defvar evil--jumps-jump-command nil
   "Set by `evil--jump-backward', used and cleared in the
 `post-command-hook' by `evil--jump-handle-buffer-crossing'")
 
@@ -242,7 +242,7 @@ POS defaults to point."
 (put 'evil-set-jump 'permanent-local-hook t)
 
 (defun evil--jump-backward (count)
-  (setq evil--jumps-jumping-backward t)
+  (setq evil--jumps-jump-command t)
   (let ((count (or count 1)))
     (evil-motion-loop (nil count)
       (let* ((struct (evil--jumps-get-current))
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by roxma** [comment]
@tomdl89  @axelf4
Hi, could you please  comment on this pull request?

File: evil-jumps.el [hunk 1]
```diff
@@ -255,6 +255,7 @@ POS defaults to point."
         (evil--jumps-jump idx 1)))))
 
 (defun evil--jump-forward (count)
+  (setq evil--jumps-jump-command t)
```
tomdl89, line 258: Is this the only line which is not renaming? I.e. could the PR just be this line (well, renaming aside)?

roxma: @tomdl89 

> Is this the only line which is not renaming?

Yes, it is.

> could the PR just be this line (well, renaming aside)?

Since this line is added, the  `backward` in the name becomes  misleading. I prefer to use a better name. 

I'll revert the renaming as you requested.

tomdl89: Thanks for your reply. I'm not requesting that, I just wanted to understand the PR properly. I'm happy to merge with the renaming.

tomdl89:
@roxma thanks for the PR. As mentioned in the comment, I don't think you need to revert the renaming.
Can you add tests for this? I've noticed there are tests in the deftest `evil-test-jump`, namely in the ert-info `Jump across files`, which you may be able to add to. Could you take a look? Thanks.

dschrempf:
Has there been any progress? I also have this problem. Whenever I jump back repeatedly, I can only jump forwards once.

tomdl89:
@dschrempf apologies for the delay. Now fixed in master. Please let me know if you observe otherwise.

dschrempf:
Wow, that was fast, thanks a lot!

roxma:
@tomdl89  Thank you for adding the tests.

dschrempf:
I can also confirm that this is working! Thanks!

PR #1775, opened 2023-04-04
Improve evil-change's handling of screen-lines

### LINKED ISSUE #1407: Certain "change-object" functions delete more than expected, including end of line
# Issue type
- Question

# Environment

Emacs version: GNU Emacs 28.0.50 (build 1, x86_64-apple-darwin19.6.0, NS appkit-1894.60 Version 10.15.7 (Build 19H15))
Operating System: macOS Catalina (though I'm pretty sure this is the same on NixOS)
Evil version: Evil version 1.14.0
Evil installation type: MELPA
Graphical/Terminal: Graphical

# Reproduction steps
- Start Emacs

- open a scratch buffer in text-mode (or any other mode should also do)
- use this as content:
```
first line
      "line in quotes"
last line
```
- place cursor inside quoted string and `daw`
- the last line is joined with the second line:
```
first line
      last line
```

# Expected behavior

When changing an entire line or changing around an quotes at the end of a line, this should not delete the newline. Changing entire lines could place cursor at indentation.

# Actual behavior

Changing an entire line or around quotes (single or double) at the end of a line also deletes the following newline.

# Question

Hi! Evil deletes more than I expect it to on certain motions. I suspect that this is just a setting that I have or have not set somewhere, but I haven't been able to find it in the docs. I also have not found a way to search for it online, so this might be easily explained somewhere else.

There's a few scenarios where this comes up (outlined below), but the short form is: how do I make `cc` delete the content on the current line and place the cursor at indentation; and how do I make `da"` (and other objects) not delete newlines?

I suspect it's related to Evil's types (`exclusive`, `inclusive`, etc) somehow, but I don't know much about that. It might also be worth mentioning that this behavior is different from Vim.

## `cc`

The first and most noticeable one is `cc` to change the current line. It deletes the entire line including the newline. In other words, this (with the cursor on the second indented line):
```
first line
      "line in quotes"
last line
```

becomes this (with the cursor at the start of the second line):

```
first line
last line
```
Ideally I'd want this to move to the first non-whitespace character and change everything from there on out (essentially act as `^ C`), especially in cases where there is a 'correct' indentation. At least I'd expect the current line to remain and the buffer to look like this (with the cursor either at the start of line 2 or at the 'correct' indentation): 
```
first line

last line
```

## `ca"` / `da"`

The second one  is when a line is encased in quotes. In this case too, the newline is included in the changing/deleting action, and I end up with the third line at the end of the second. This, with the cursor inside the quoted line:
```
first line
      "line in quotes"
last line
```
becomes 
```
first line
      last line
```
-------

I have not gone through the reproduction steps yet, as I think this is related to a setting I'm not aware of. If this turns out to be an actual problem, I'll happily see if I can reproduce it.

Thanks!

justbur:
When the type is screen-line and multiple real lines are spanned, make evil-change behave like it does for the line type (e.g., by adding and indenting a new line after deletion).

Partially fixes #1407

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -1696,7 +1696,9 @@ of the block."
       (evil-start-undo-step))
     (funcall delete-func beg end type register yank-handler)
     (cond
-     ((eq type 'line)
+     ((or (eq type 'line)
+          (and (eq type 'screen-line)
+               (> nlines 1)))
       (setq this-command 'evil-change-whole-line) ; for evil-maybe-remove-spaces
       (cond
        ((/= opoint leftmost-point) (evil-insert 1)) ; deletion didn't delete line


axelf4 (changes requested):
This seems to break `cc` on the last continuation visual line as `leftmost-point` will refer to the start of the logical line.

Also what is the purpose of the `(> nlines 1)` check? For handling the last buffer line the special case in `evil-delete` needs to be extended to `screen-line` (e.g. `2cc`).

justbur:
(Sorry for the delay)

> This seems to break cc on the last continuation visual line as leftmost-point will refer to the start of the logical line.

I'm not sure what you mean by "break". I don't see any obvious problems. Maybe you could give an example? 

> Also what is the purpose of the `(> nlines 1)` check?

The idea was that if you are in the middle of a long line that is wrapped and you select a single visual line you don't want to introduce a line break after `cc`. You just want to delete the corresponding text and insert something new. Note that if you delete a whole logical line, the selection covers two newline characters so you get `nlines` equal to 2.

>  For handling the last buffer line the special case in evil-delete needs to be extended to screen-line (e.g. 2cc).

Thanks, added a fix

PR #1789, opened 2023-05-09
evil-find-file-at-point-with-line: Support format path(line,col).

forty:
This format can, for example, be found in tsc (the Typescript compiler) output.

I have been using this in my local emacs config, and I find it useful, so I figured it could be useful for others. I'm not very used to writing Lisp code, so I'm won't be vexed if you tell me the code is horribly wrong :D

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -3647,21 +3647,28 @@ If FORCE is non-nil and MARKS is blank, all local marks except 0-9 are removed."
 
 (eval-when-compile (require 'ffap))
 (evil-define-command evil-find-file-at-point-with-line ()
-  "Open the file at point and go to position if present."
+  "Open the file at point and go to position if present.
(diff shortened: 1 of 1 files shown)


tomdl89:
Thanks @forty - looks like a good addition. Couple of things I'd add: can your solution handle `file(line)` ? I think not, and I think it would be a good addition. Also (and this has been a bug for a while) the `message` should handle no column number better than `line: 2, column: nil`

forty:
I added a commit to handle `file(line)`, good idea.

For the `message`, what do you have in mind ideally? Note that `line-number` can also be `nil`, so we should also probably plan something for `line: nil, column: nil`

forty:
I added one extra commit with some better handling of nil in the message. Please let me know what you think.

tomdl89:
Great, thank you. I added some small fixes, to be squashed

PR #1818, opened 2023-08-09
Make evil-search-wrap-ring-bell work with evil-search

jn64:
evil-search-wrap-ring-bell only worked when evil-search-module was isearch. Make it work with evil-search as well.


tomdl89:
Thanks for the PR @jn64!

PR #1821, opened 2023-08-14
add keep-active-mark property to states

LemonBreezes:
This property is used to determine whether we should enter visual state when the mark is activated. This is useful for states like emacs state where we want to keep the mark active but not enter visual state. This change enables users to customize this behavior for their own states.

This addresses issue https://github.com/gridaphobe/evil-god-state/issues/4 in `evil-god-state`.

### DIFF
--- a/evil-core.el
+++ b/evil-core.el
@@ -1168,7 +1168,8 @@ the local keymap will be `evil-test-state-local-map', and so on.
          (modes (intern (format "%s-modes" toggle)))
          (predicate (intern (format "%s-p" toggle)))
          arg cursor-value enable entry-hook-value exit-hook-value
-         input-method key message-value suppress-keymap tag-value)
+         input-method key message-value suppress-keymap tag-value
+         keep-active-mark)
     ;; collect keywords
     (while (keywordp (car-safe body))
       (setq key (pop body)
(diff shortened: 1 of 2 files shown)


axelf4:
I think the idea is sound but would be interested to hear what others think. The name `keep-active-mark` is a bit confusing - Visual state does not deactivate the mark.

The other solution I could think of which is cleaner but not 100% backward compatible is making the states themselves add the `evil-visual-activate-hook` hook when changed to.

LemonBreezes:
> I think the idea is sound but would be interested to hear what others think. The name `keep-active-mark` is a bit confusing - Visual state does not deactivate the mark.
> 
> The other solution I could think of which is cleaner but not 100% backward compatible is making the states themselves add the `evil-visual-activate-hook` hook when changed to.

I was thinking about this PR just now and I don't like how as a state property, it's meant to be immutable. Because I just realized that customizing the modes in `evil-visual-activate-hook` could also be used to make marking commands change the state from `insert` to `visual`. I've been experimenting with this combined with https://github.com/LemonBreezes/restore-point and I think that that combination of transiently entering visual state and restoring the point + insert state with `<escape>` feels very natural. So maybe we can have the states add the hook themselves.

PR #1825, opened 2023-08-21
Fix Vim URL in docs

wleightond:
The URL [`https://vim.org`](https://vim.org) fails to resolve; this PR replaces the two instances of that in the docs to [`https://www.vim.org`](https://www.vim.org) which behaves as expected. This is already correct in [`README.md:12`](https://github.com/emacs-evil/evil/blob/master/README.md?plain=1#L12).


axelf4:
Thanks!

PR #1827, opened 2023-08-22
(Failing) test for `&` with very-magic vim-style backreferences

tomdl89:
@axelf4 I noticed this bug recently and thought surely it must have been introduced recently, but I went back pretty far and could still replicate the bug, so I guess not. Nonetheless, I think you may have been closer to this area of code recently, so if you have any thoughts, feel free to have a stab at the fix. If not, I'll get to it in a few weeks, hopefully. Cheers


axelf4:
> Nonetheless, I think you may have been closer to this area of code recently, so if you have any thoughts, feel free to have a stab at the fix.

No, sorry, `evil-search` is still a jungle to me. :)

tomdl89:
Cool no worries. I started taking a look earlier today and think I made some progress. Fix should be some time next week 🤞

PR #1844, opened 2023-12-04
Assign :jump to evil-{next-close,previous-open}-{brace,paren}

leungbk:
This is how Vim behaves.


leungbk:
cc @tomdl89

tomdl89:
Thanks for the PR @leungbk . In future, for functional changes like this, please include tests so we can catch regressions in the future. If you would like to write them retrospectively, I'll happily merge. N.b. if you do for this, make sure the jumps tested are multi-line, as I'd rather not enforce within-the-same-line jumps, as they are not usually how vim works, and maybe one day we'll switch them off, or at least allow it.

PR #1846, opened 2023-12-09
Add :retab command

FrostyX:
See "Changing tabs" section in the Vim manual
https://vimhelp.org/change.txt.html#change.txt

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -5168,6 +5168,23 @@ event that triggered the execution of this command."
     (push event unread-command-events)))
 (evil-declare-ignore-repeat 'evil-exit-visual-and-repeat)
 
+(evil-define-command evil-retab (tabstop)
+  "Convert all tabs to spaces or the other way around.
+Replace all sequences of white-space containing a <Tab> with new
+strings of white-space using the new TABSTOP value given.
+If you do not specify a new TABSTOP size or it is zero, Evil uses the
+current value of `tab-width'."
+  (interactive "<a>")
+  (unless (or (not tabstop) (string-match-p "^[0-9]*$" tabstop))
+    (user-error "Invalid argument: %s" tabstop))
+  (let ((beg (if (use-region-p) (region-beginning) (point-min)))
+        (end (if (use-region-p) (region-end) (point-max)))
+        (retab (if indent-tabs-mode #'tabify #'untabify))
+        (tab-width (cond ((not tabstop) tab-width)
+                         ((equal tabstop "0") tab-width)
+                         (t (string-to-number tabstop)))))
+    (funcall retab beg end)))
+
 (provide 'evil-commands)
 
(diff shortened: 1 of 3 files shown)


evil-commands.el
```diff
@@ -5168,6 +5168,21 @@ event that triggered the execution of this command."
...
+strings of white-space using the new TABSTOP value given.
+If you do not specify a new TABSTOP size or it is zero, Evil uses the
+current value of `tab-width'."
+  (interactive "<a>")
+  (let ((beg (if (use-region-p) (region-beginning) (point-min)))
+        (end (if (use-region-p) (region-end) (point-max)))
+        (retab (if indent-tabs-mode #'tabify #'untabify))
+        (tab-width (cond ((not tabstop) tab-width)
+                         ((equal tabstop "0") tab-width)
+                         (t (string-to-number tabstop)))))
```
tomdl89, line 5183: I think `string-to-number` is too permissive. e.g.
```elisp
(string-to-number "foo")
0
(string-to-number "7foo")
7
```
This  (before the let binding, even) would catch most things:
```elisp
(string-match-p "^[0-9]*$" tabstop)
```

tomdl89:
Thanks fort the PR @FrostyX . Please add some tests so a) I can easily see what your implementation does and b) to guard against regressions in the future. Cheers

evil-commands.el
FrostyX: Good catch, fixed.

FrostyX:
Thank you very much for the review @tomdl89,
I was a bit worried that it will be complicated to write the tests but I was sooo wrong. The tests are so comfortable to write thanks to the `evil-test-buffer`.

Updated, PTAL :-)

tomdl89:
Looks great - thanks for the tests @FrostyX ! Just need to get that `Retab only selected lines` test passing and I'm happy to merge :)

FrostyX:
Yeah, I am trying to figure out what is wrong. As these things usually go, the test works fine on my machine.

tomdl89:
FWIW, they pass on my machine too with `make emacs` and `make terminal` but not with `make test` (which is what github actions uses).

tomdl89:
OK, having looked at this, and got it passing for each, but never together, I'd be happy to remove this test. It shows a difference between `--batch` mode and normal mode which I don't understand. A minimal repro (without evil) for the emacs dev mailing list may be interesting if you can't work out what causes the difference. Up to you.

FrostyX:
Thank you very much for trying to debug the issue.

>  I'd be happy to remove this test. 

Do you mind if we skip the tests in the batch mode instead? I updated the PR. I think it is worth showing what the command is supposed to do, as you said in

> a) I can easily see what your implementation does

tomdl89:
Yup, all good. Thanks again for the PR!

FrostyX:
Thank you very much.

PR #1847, opened 2023-12-12
Add racket-repl-mode to evil-insert-state-modes

leungbk:


leungbk:
While this wasn't necessary before due to the old `racket-repl-mode` deriving from `comint-mode`, upstream has since rewritten `racket-repl-mode` to derive from `fundamental-mode` instead.

PR #1856, opened 2024-01-09
Added clarification to evil-define-key for escaping modifier keys

Naokotani:
It is useful to have clarification in the doc string for evil-define-key that it is necessary to escape modifiers keys, such as "C" and "M", for users that are accustomed to using functions like global-set-key or keymap-global-set where this is not necessary. This is also not made particularly clear in the define-key function itself, which is a legacy function.

### DIFF
--- a/evil-core.el
+++ b/evil-core.el
@@ -958,7 +958,13 @@ the following lead to identical bindings:
 The symbol `local' may also be used, which corresponds to using
 `evil-local-set-key'. If a quoted symbol is used that is not
 `global' or `local', it is assumed to be the name of a minor
-mode, in which case `evil-define-minor-mode-key' is used."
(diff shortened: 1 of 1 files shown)


axelf4 (changes requested):
This is a good idea, but there are some things that I think could be reworded:

* Users should just use `kbd` instead of typing out the key sequence representations by hand, unless they already know what they are doing.
* That the macro uses `define-key` is an implementation detail, could instead say roughly something like this:

  > KEY is an internal Emacs representation of a key, as for `define-key`. To bind key sequences that use modifier keys such as ..., convert the key sequences using `kbd`.
* As `C`/`M`/`\` are not Lisp symbols they should be quoted with double quotes instead of `` `...' ``.

axelf4 (approved):

PR #1860, opened 2024-01-28
WIP: CI: enable check for lack of new warnings while byte-compiling

Hi-Angel:
This fixes one warning and adds a check that makes sure that CI will fail if a pull-request contains byte-compilation errors. As part of that we replace "emacs-snapshot" in CI with the latest released one because it's "frozen in time", while depending on snapshot may result in CI failing just because upstream introduced new warnings.

This is WIP because: there's one "deprecation warning" which Idk what to do about, because it's internal in Evil. It sounds like:

```
evil-types.el:96:22: Error: ‘evil-want-visual-char-semi-exclusive’ is an obsolete variable (as of 1.15.0); Semi-exclusivity prevents selecting text + 1st char of next line,
without having to introduce new niche functionality.
```

Thoughts? One possible solution is to disable check for "deprecation warnings", but then we will miss new deprecations from upstream.

---------

P.S.: the changes in Makefile also result in a wall of weird messages `Makefile:34: warning: overriding recipe for target 'evil-search.elc'`. I would be very grateful if someone points out how to fix that because I'm not that good in Makefiles *(I usually migrate everything I maintain to Ninja or Meson)* and Makefiles often behave weirdly.


Hi-Angel:
I sent the "warning fix" commit separately at https://github.com/emacs-evil/evil/pull/1861 to have the fix in tree while the discussion here is ongoing.

Hi-Angel:
Weird… so, I fixed everything, but on the older Emacs'es it fails with "undo-redo" undefined. The problem is that it **is** defined, I put a `declare-function` as part of the make compilation call. No idea what's going on…

Hi-Angel:
Okay, I think I'm done for now, I just don't have anymore resources, this is utter chaos. In some versions some functions were in one or another file. [This fail](https://github.com/emacs-evil/evil/actions/runs/7978929972/job/21785140111?pr=1860) I don't even understand. It claims to not have found `string-trim-right` which at that version was in `subr-x`, I checked in Emacs sources. So I have added an optional `require subr-x` for Emacs versions lower than 29 — but the error just won't go away and I don't know why.

PR #1857, opened 2024-02-10
evil-mode text insertion breaks emacs text property inheritance (stickiness of text properties)

dizzyyogi:
# Issue type

- Bug report

# Environment

Emacs version: 29.0.91
Operating System: any
Evil version: 1.15.0
Evil installation type: straight.el + use-package
Graphical/Terminal: X

# Reproduction steps
- Start Emacs
- Open a scratch buffer
- Put some arbitrary text property on a whole line
For example, from the following initial scratch buffer content:
==============================
;; This buffer is for text that is not saved, and for Lisp evaluation.
;; To create a file, visit it with C-x C-f and enter text in its buffer.
==============================
Where position 72 corresponds to the first semi-column at the beginning of second line, and position 145 corresponds to the end of the second line
- Execute:
(put-text-property 72 145 'some-text-property t)
- open a new line in evil normal state, with cursor placed somewhere on second line, and type some text:
M-x evil-open-below ("o")
Type "hello world"

# Expected behavior
Because default stickiness of text properties is rear-sticky (see emacs manual, 33.19 Text Properties -> Sticky Properties), the text "hello world" and the newline character added after line 2 should have inherited the text property 'some-text-property set to t in the step 4 above.
By self-insertion of a new line character and "hello world" string in emacs state, the property is correctly inherited:
Place cursor on second line
C-z
C-e
TYpe Return key
Type "hello world"
Place cursor on any of the inserted characters
Use command what-cursor-position with universal arg to get the text properties at point:
C-u C-x = 
Any of the inserted characters have inherited property 'some-text-property, because of default rear-stickiness rule.

# Actual behavior
After evil-open-below + text insertion
The text property 'some-text-property is not attached to the inserted text "hello world".

# Further notes
The functions evil-insert-newline-above/below use (insert "\n").
As indicated in elisp manual, 33.19.6 Stickiness of Text Properties":
 "The ordinary text insertion functions, such as 'insert', do not inherit any properties. They insert text with precisely the properties of the string being inserted, and no others."
Shouldnt' function "insert-and-inherit" be used instead?


evil-states.el
```diff
@@ -881,7 +881,7 @@ CORNER defaults to `upper-left'."
          (this-command #'evil-replace-backspace))
     (when prev
       (delete-char 1)
-      (when char (save-excursion (insert char))))))
+      (when char (save-excursion (insert-and-inherit char))))))
```
tomdl89, line 884: As mentioned, I think this is one function that shouldn't inherit when inserting.

evil-commands.el
tomdl89, line 1: I prefer to delete lines than commenting them out. We have the git log if we need to get them back.

dizzyyogi:
Hi @tomdl89 and happy new year :-)
Just a friendly reminder on this PR. I had provided an update based on your last review. Would any other change be needed? I've been using the branch in my personal config over the past months and did not notice any particular issue.

dizzyyogi:
Dear Axel,

The built-in yank function appears however to preserve killed text properties,
while evil-past does not.
In the example use case I have described in the original issue description,
attaching some arbitrary text property to some characters in a buffer, copying
them (kill-ring-save) and then pasting them in some remote place of the buffer
with built-in yank, preserves the text property.
By contrast, with evil, copying some character with attached text property,
e.g. with "yl", and pasting it with evil-paste, the text property is lost.

bo-tato:
I ran into this issue when using SLIME (common lisp dev environment for emacs). When you kill and yank slime "presentations" with normal emacs commands, they maintain their text properties and function correctly. But when you paste it with `evil-paste-after`, it loses the text properties and the presentation no longer works. Curiously when you copy the whole line that contains the presentation with `yy` and then paste with `p evil-paste-after`, it does work, as `yy` registers a yank-handler of `evil-yank-line-handler` which preserves the text properties. This PR, specifically commit [cba701c](https://github.com/emacs-evil/evil/pull/1857/commits/cba701c8f31b60cac62bf5a3341fa531fe5c4a2d) would fix this, and make copy and pasting presentations in SLIME work as it does outside of evil-mode.

PR #1875, opened 2024-03-30
Add interactive highlights for ex global command pattern

blahgeek:


tomdl89:
Hi @blahgeek thanks for the PR. How does your implementation compare to that in https://github.com/mamapanda/evil-traces? Could code for other highlighting be brought over? Could recommending evil-traces in the readme be preferable? I'm not sure if you're aware of its existence, and if not, if knowing about it would have changed your approach.

tomdl89:
Ah, and I guess more importantly, does your functionality and that of evil-traces conflict at all?

blahgeek:
Ah, I did not know the existence of evil-traces. Let me take a took and get back to you. Thanks!

blahgeek:
I briefly tried evil-traces and skimmed through its code, here's what I found:

Evil-traces does already cover the feature of this patch, along with the support for many other commands. Its implementation is similar to this patch, and it would override current implementation (so no conflict).

However I do find several edge cases that evil-traces is incorrect:

- It does not handle patterns with `\c` or `\C` correctly
- It does not handle vim style patterns (e.g. with `\d`) correctly
- It does not support ex-global within visual range (`:'<,'>g/`)

Surely these issues can be fixed in evil-traces, but IMO it would be more likely to work correctly in the long run being implemented in evil package itself (with more integrated codes and tests).

Ultimately I think it depends on whether you think this feature (and possibly other highlights features) belong to evil core package. IMO, since evil already provides substitution highlight feature, it's nature to also provide ex-global-command highlight feature, but other highlight features maybe not so much.

PR #1889, opened 2024-04-20
Fix Visual Insert affecting one too many lines

axelf4:
This fixes a regression caused by commit 004ac4e0cd766d49d48d53270e9c0c080ad9f173, where Visual mode "I" repeated the insertion on one too many lines unless at EOB.

### DIFF
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -2655,20 +2655,19 @@ corner and point in the lower left."
     (evil-visual-refresh))))
 
 (evil-define-command evil-visual-rotate (corner &optional beg end type)
-  "In Visual Block selection, put point in CORNER.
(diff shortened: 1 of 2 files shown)


Hi-Angel:
@axelf4 is it just me *(being a devel-version Emacs user)*, or does this commit result in failure to byte-compile Evil?

Upgrade to recent evil from MELPA only compiles part of Evil files, but compiling `evil-commands.el` fails. Here's a snippet from `make compile` when run on latest Evil code:

```
evil-commands.el:2644:2: Error: List contains a loop: (lower-right lower-left upper-left upper-right lower-right lower-left upper-left upper-right lower-right lower-left . #6)
```

PR #1892, opened 2024-04-27
 Remove evil--visual-eol-anchored

axelf4:
There already exists a method for sticking to EOL: Setting `temporary-goal-column` to `most-positive-fixnum`. As such, there is no need for the variable `evil--visual-eol-anchored` introduced by commit e31bff8cb64d773bbfd9a8e326db8645eaee43fd.

@tomdl89 I cannot replicate `evil-test-end-of-visual-line` in Vim, what motivated its addition? I have removed it here.


tomdl89:
Hi @axelf4 I'm away for a bit but will review in a few days. Please don't merge until I take a look. Cheers

evil-maps.el
```diff
@@ -238,7 +238,7 @@
 (define-key evil-motion-state-map "gj" 'evil-next-visual-line)
 (define-key evil-motion-state-map (vconcat "g" [down]) 'evil-next-visual-line)
 (define-key evil-motion-state-map "gk" 'evil-previous-visual-line)
-(define-key evil-motion-state-map (vconcat "g" [up]) 'evil-previous-visual-line)
+(define-key evil-motion-state-map (vconcat "g" [up]) 'evil-visual-previous-line)
```
tomdl89, line 241: @axelf4 I can't see that this function exists anywhere..?

tomdl89:
I've had a quick look and agree that that test doesn't seem to correspond to behaviour in vim. I could have sworn I did test it against vim. Apart from my comment about g-up above, I think I'd rather merge this now as it does seem like something worth reversing. Then I can try to work out what I was thinking later! Cheers

evil-maps.el
axelf4: Good catch! I renamed the functions while testing, but accidentally did not change this one back.

axelf4:
Maybe a suggestion for an advice like
```elisp
(define-advice evil-end-of-visual-line (:after (&rest _))
  (setq temporary-goal-column most-positive-fixnum
        this-command 'next-line))
```
could be added to the documentation? (Makes `g$` anchor *both* `j`/`k` and `gj`/`gk` to EOL, but that is symmetric with `$`.)

Issue #1835, opened 2023-09-09
Cursor color does not persist


pietroiusti:
# Issue type

- Bug report

# Environment

Emacs version: GNU Emacs 29.1.50 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.38, cairo version 1.17.8) of 2023-09-02
Operating System: Arch Linux
Evil version: Evil version evil-git-5fc1677
Evil installation type: I'm using make emacs in Evil's source directory
Graphical/Terminal: Graphical
Tested in a `make emacs` session (see [CONTRIBUTING.md](https://github.com/emacs-evil/evil/blob/master/CONTRIBUTING.md#bug-reports)): Yes

# Reproduction steps
- Start Emacs
- switch to the scratch buffer
- switch to Emacs state using C-z
- yank the following lines:
  ```
   (setq evil-disable-insert-state-bindings t)
   (setq evil-default-state 'emacs)
   (setq evil-normal-state-cursor '("#e80000" box)) ;; red
   (setq evil-insert-state-cursor '("#e80000" bar)) ;; red
   (setq evil-emacs-state-cursor '("#839496" box)) ;; grey
- M-x eval-buffer
- Press C-z. The cursor correctly becomes red.
- Place the cursor on any characters of `evil-emacs-state-cursor` on the last line.

# Expected behavior
Cursor's color should remain red.

# Actual behavior
Cursor's color becomes grey.

# Further notes
I suspect the color change is triggered in other contexts as well. This was the easiest way to replicate.

Here is the backtrace using debug-on-entry on set-cursor-color:

```
Debugger entered--entering a function:
* set-cursor-color("#839496")
  (if (equal (frame-parameter nil 'cursor-color) color) nil (set-cursor-color color))
  evil-set-cursor-color("#839496")
  (cond ((functionp spec) (condition-case nil (progn (funcall spec)) (error nil))) ((stringp spec) (evil-set-cursor-color spec)) (t (setq cursor-type spec)))
  (let ((spec (car tail))) (cond ((functionp spec) (condition-case nil (progn (funcall spec)) (error nil))) ((stringp spec) (evil-set-cursor-color spec)) (t (setq cursor-type spec))) (setq tail (cdr tail)))
[...]

PR #1910, opened 2024-06-15
Fix cursor colour

### LINKED ISSUE #502: Cursor is not refreshed in some cases
Originally reported by: **Sylvain Benner (Bitbucket: [syl20bnr](https://bitbucket.org/syl20bnr), GitHub: [syl20bnr](https://github.com/syl20bnr))**

We have a regression with the fix #441. The cursor is reset back to default state color whenever a buffer is created without changing the focused window and without changing the window configuration.

See the comment on #441 for more info.

Tested with last MELPA version.

- Bitbucket: https://bitbucket.org/lyro/evil/issue/502

fnussbaum:
Fixes https://github.com/emacs-evil/evil/issues/502, fixes https://github.com/emacs-evil/evil/issues/1835. Refreshing the cursor when changing the evil state is
only needed (or almost only, see below) when the current buffer is displayed in the
selected window. If this is not the case, and the buffer is only displayed or
its window selected at some later point in time, then `evil-refresh-cursor` will
be called as part of the `window-configuration-change-hook` or as advice to
`select-window`.

However, this introduces the following tiny (and maybe acceptable?) regression:
When doing something like
```elisp
(with-current-buffer some-buffer-displayed-in-another-window
  (some-evil-state-with-a-different-cursor-type))
```
the cursor will not be refreshed in the other window before selecting it. The
cursor colour should indeed not be refreshed, because it is defined for the
whole frame; however, the cursor type should in principle be changed, as it is
defined per buffer and also defines the shape of the cursor in non-selected
windows.

There exist different ways to also handle this case, but they mostly seem ugly
or needlessly complicated to me. I think the most elegant way to fix this would
involve implementing per-buffer cursor colours in Emacs (as suggested in
https://debbugs.gnu.org/cgi/bugreport.cgi?bug=24153d). What do you think?

### DIFF
--- a/evil-core.el
+++ b/evil-core.el
@@ -1279,7 +1279,8 @@ If ARG is nil, don't display a message in the echo area.%s" name doc)
                  (when deactivate-current-input-method-function
                    (deactivate-input-method)))
                (unless evil-no-display
-                 (evil-refresh-cursor ',state)
+                 (when (eq (window-buffer) (current-buffer))
+                   (evil-refresh-cursor ',state))
                  (evil-refresh-mode-line ',state))
                ,@body
                (run-hooks ',entry-hook)


tomdl89:
Thanks for the PR @fnussbaum I think this is a net improvement. Hopefully we won't see a rush of complaints wrt the regression you mentioned. I'm happy to take the risk.

PR #1922, opened 2024-07-09
make C-w in ex, search, and eval match insert and replace

### LINKED ISSUE #1921: `C-w` in ex command-mode doesn't match insert mode (or Vim)
# Issue type

- Bug report

# Environment

Emacs version: GNU Emacs 29.4 (build 2, x86_64-unknown-linux-gnu, GTK+ Version 3.24.42, cairo version 1.18.0) of 2024-06-25
Operating System: Void Linux
Evil version: Evil version 1.15.0
Evil installation type: Straight (Doom)
Graphical/Terminal: Graphical

# Reproduction steps
- Start Emacs
- type `:test -- -t` (or similar)
- do `C-w` twice

# Expected behavior
`:test -- ` should be left

# Actual behavior
`:` is left

# Further notes
The behaviour is correct in insert mode, so it's just a matter of ex calling `evil-delete-backward-word` instead of `backward-kill-word`.
(I believe that's a change to `evil-ex-completion-map`. Remapping there seems to fix it for me)

**EDIT**: In the process of trying to write a PR to fix this, I've discovered that this also affects search and the eval register.

totallyuniquelily:
Fixes #1921

The test may be a lil hacky, and doesn't cover search or eval.

### DIFF
--- a/evil-maps.el
+++ b/evil-maps.el
@@ -624,7 +624,7 @@ included in `evil-insert-state-bindings' by default."
 (define-key evil-command-line-map "\C-u" 'evil-delete-whole-line)
 (define-key evil-command-line-map "\C-v" #'quoted-insert)
 (when evil-want-C-w-delete
-  (define-key evil-command-line-map "\C-w" #'backward-kill-word))
+  (define-key evil-command-line-map "\C-w" #'evil-delete-backward-word))
 (define-key evil-command-line-map [escape] #'abort-recursive-edit)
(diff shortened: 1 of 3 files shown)


totallyuniquelily:
I've fixed the test, sorry for not catching it the first time.

I've ran the tests locally and it works, (although `evil-test-flyspell-motions` freezes for me and I have to skip/quit it.)

PR #1924, opened 2024-07-17
Bind all possible ex cmd abbreviations

tomdl89:
Means that :de expands to delete

### DIFF
--- a/evil-ex.el
+++ b/evil-ex.el
@@ -585,13 +585,15 @@ in case of incomplete or unknown commands."
 
 (defun evil-ex-define-cmd (cmd function)
   "Bind the function FUNCTION to the command CMD."
-  (if (string-match "\\[\\(.*\\)\\]" cmd)
-      (let ((abbrev (replace-match "" nil t cmd))
-            (full (replace-match "\\1" nil nil cmd)))
+  (if (not (string-match-p "\\[" cmd))
+      (evil--add-to-alist evil-ex-commands cmd function)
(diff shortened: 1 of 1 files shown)


axelf4:
This is not correct: `evil-ex-completed-binding` already uses `try-completion`. At the very least it should not anymore, but I am not fond of how this change balloons the size of the `evil-ex-commands` list. Furthermore, it ignores the problem of collisions?

tomdl89:
* > `evil-ex-completed-binding` already uses `try-completion`
true, but I can't see that it's meaningfully different from using `assoc` and `cdr` to look up the binding in `evil-ex-commands` - if I'm missing something, we can make a change on this.
* > I am not fond of how this change balloons the size of the `evil-ex-commands` list.
so, I tested this before I merged to master. `(length evil-ex-commands)` was 167 before, and 334 after, so my conclusion is that it will make no significant difference, performance-wise. again, if I'm missing something (especially some data) that would change things, for sure
* > Furthermore, it ignores the problem of collisions?
I don't think so. it only makes bindings of the length of the mandatory part (i.e. before `[`) and longer (at which point we're not expecting collisions anyway). Do you have an example?

Also, more broadly, I won't pretend this isn't a reasonably quick and dirty fix to a problem that's been bothering me for a while. If you have a better fix in mind, I'm more than happy to replace my fix with yours. I'd be inclined to keep the change in the meantime, unless of course it's demonstrably less performant, or causes collisions I hadn't considered. Cheers

Issue #1778, opened 2023-04-10
Bug with scrolling when pixel-scroll-precision-mode is enabled


ashton314:
# Issue type

- Bug report

Short of it: scrolling up gets stuck when `pixel-scroll-precision-mode` is turned on and Evil is enabled. I can get around it by checking out evil revision `f003ca2`. (Not sure if that's the most recent commit, but there's a good clean starting point.)

# Environment

- Emacs version: GNU Emacs 29.0.90 (build 1, aarch64-apple-darwin22.4.0, NS appkit-2299.50 Version 13.3.1 (Build 22E261)) of 2023-04-10
- Operating System: macOS 13.3.1
- Evil version: 1.15.0 (`master` sometime after `f003ca2` has this issue; `f003ca2` is clean)
- Evil installation type: Elpaca (`master` from source)
- Graphical/Terminal: Graphical
- Tested in a `make emacs` session (see [CONTRIBUTING.md](https://github.com/emacs-evil/evil/blob/master/CONTRIBUTING.md#bug-reports)): No; used the `--init-directory` from Emacs 29 to get a clean env.

# Reproduction steps
- Create a fresh directory, and put this into a file called `init.el`:

```emacs-lisp
(setq inhibit-startup-screen t)

(defvar elpaca-installer-version 0.3)
(defvar elpaca-directory (expand-file-name "elpaca/" user-emacs-directory))
(defvar elpaca-builds-directory (expand-file-name "builds/" elpaca-directory))
(defvar elpaca-repos-directory (expand-file-name "repos/" elpaca-directory))
(defvar elpaca-order '(elpaca :repo "https://github.com/progfolio/elpaca.git"
                              :ref nil
                              :files (:defaults (:exclude "extensions"))
                              :build (:not elpaca--activate-package)))
(let* ((repo  (expand-file-name "elpaca/" elpaca-repos-directory))
       (build (expand-file-name "elpaca/" elpaca-builds-directory))
       (order (cdr elpaca-order))
       (default-directory repo))
  (add-to-list 'load-path (if (file-exists-p build) build repo))
  (unless (file-exists-p repo)
    (make-directory repo t)
    (condition-case-unless-debug err
        (if-let ((buffer (pop-to-buffer-same-window "*elpaca-bootstrap*"))
                 ((zerop (call-process "git" nil buffer t "clone"
                                       (plist-get order :repo) repo)))
                 ((zerop (call-process "git" nil buffer t "checkout"
[...]

PR #1926, opened 2024-07-27
Fix pixel-scroll-precision-mode when point is on a blank line

Stebalien:
When in the point is on a blank line in normal state, scrolling down with `pixel-scroll-precision-mode` enabled "sticks" because Evil ends up invoking `line-move` in the `post-command-hook`. `line-move` resets vscroll to 0, which undoes the scroll if we scrolled less than a line.

Specifically, `evil-normal-post-command` calls `evil-adjust-cursor` which calls `evil-move-end-of-line` which calls `move-end-of-line`, which finally calls `line-move`.

This change avoids this issue by avoiding `evil-move-end-of-line` entirely, instead using `bolp` to check if we're already at the beginning of the line.

(note: I'm not sure if there were subtle reasons for calling `evil-move-end-of-line` here)

Upstream bug report: https://debbugs.gnu.org/cgi/bugreport.cgi?bug=72323. But Evil probably shouldn't be calling `line-move` here anyways.

### DIFF
--- a/evil-common.el
+++ b/evil-common.el
@@ -970,7 +970,6 @@ Like `move-end-of-line', but retains the goal column."
 This behavior is controlled by `evil-move-beyond-eol'."
   (and (not evil-move-beyond-eol)
        (eolp)
-       (= (point) (save-excursion (evil-move-end-of-line) (point)))
        (evil-move-cursor-back t)))
 
 (defun evil-move-cursor-back (&optional force)


axelf4 (changes requested):
Thank you for looking into this. I cannot think of anything important the `evil-move-end-of-line` call is doing, *with the preceding `(eolp)` check* (added by commit 907e7aa4b687b5df1e3f87ca830e8e9e56e00b14).

See: #1778

On a broader note, why does `save-excursion` not restore vscroll? Would that not alleviate this particular issue?

Stebalien:
> On a broader note, why does save-excursion not restore vscroll? Would that not alleviate this particular issue?

`save-excursion` doesn't affect scrolling, just point. E.g., `(save-excursion (set-window-start nil 0)` will scroll to the top of the buffer, leaving point where it was.

PR #1938, opened 2024-11-07
Fix an issue where moving the cursor to end of the buffer would result in no visible text

jamescherti:
This pull request fixes an issue where using `(evil-goto-line nil)` to move the cursor to the end of the buffer would result in no visible text because of `(goto-char (point-max))`.

This issue is difficult to reproduce because it doesn't always occur. I was fortunate enough to encounter it today, and the provided patch fixes the problem. This patch draws inspiration from the built-in `(end-of-buffer)` function, which does not exhibit the same issue as calling `(evil-goto-line nil)`.

### DIFF
--- a/.dir-locals.el
+++ b/.dir-locals.el
@@ -1,3 +1,0 @@
-;; Don't use tabs for el files
-((emacs-lisp-mode .
-   ((indent-tabs-mode . nil))))
--- a/evil-commands.el
+++ b/evil-commands.el
@@ -254,7 +254,18 @@ of the current screen line."
   :type line
   (evil-ensure-column
     (if (null count)
-        (goto-char (point-max))
+        (progn
+          (goto-char (point-max))
+          (when (and (eq (current-buffer) (window-buffer))
+                     (> (point) (window-end nil t)))
+            ;; When the user sets `scroll-conservatively` to a value greater
+            ;; than 100, moving the cursor to `(point-max)` with
+            ;; `(evil-goto-line nil)` can result in all text being located above
+            ;; the window start, preventing the user from seeing it.
+            ;;
+            ;; The following addresses this issue:
+            (overlay-recenter (point))
+            (recenter -1)))
       (goto-char (point-min))
       (forward-line (1- count)))))
 


axelf4:
Thanks for the PR! Could you please try to give reproduction steps? Workarounds that cannot be verified as necessary make me uneasy, as they are impossible to remove later. But this also sounds more like a GNU Emacs bug, and in that case it ought to be fixed there instead?

jamescherti:
This issue is difficult to reproduce consistently. To attempt to trigger it, try opening multiple files and press `G` for each file until no text remains visible. However, I have not been able to reproduce it reliably myself.

I am not sure whether this is a bug in GNU/Emacs. It does not seem to be the responsibility of `(goto-char (point-max))` to adjust the window start position.

Even the built-in `end-of-buffer` command uses `recenter` to adjust the window start:
```elisp
(defun end-of-buffer (&optional arg)
  "Move point to the end of the buffer.
With numeric arg N, put point N/10 of the way from the end.
If the buffer is narrowed, this command uses the end of the
accessible part of the buffer.

Push mark at previous position, unless either a \\[universal-argument] prefix
is supplied, or Transient Mark mode is enabled and the mark is active."
  (declare (interactive-only "use `(goto-char (point-max))' instead."))
  (interactive "^P")
  (or (consp arg) (region-active-p) (push-mark))
  (let ((size (- (point-max) (point-min))))
    (goto-char (if (and arg (not (consp arg)))
		   (- (point-max)
		      (/ (* size (prefix-numeric-value arg)) 10))
		 (point-max))))
  ;; If we went to a place in the middle of the buffer,
  ;; adjust it to the beginning of a line.
  (cond ((and arg (not (consp arg))) (forward-line 1))
	((and (eq (current-buffer) (window-buffer))
              (> (point) (window-end nil t)))
	 ;; If the end of the buffer is not already on the screen,
	 ;; then scroll specially to put it near, but not at, the bottom.
	 (overlay-recenter (point))
	 ;; FIXME: Arguably if `scroll-conservatively' is set, then
         ;; we should pass -1 to `recenter'.
	 (recenter (if (and scroll-minibuffer-conservatively
	                    (window-minibuffer-p))
	               -1 -3)))))
```

The Emacs developers likely called `overlay-recenter` and `recenter` for this purpose.

jamescherti:
Hello @axelf4,

I finally found a way to reproduce the issue.

Step 1:  
- Run the command: `(toggle-truncate-lines 0)`
- Run the command: `(text-scale-set 5)`

Step 2: Repeat the following steps many times until the issue happens:
- Press the "gg" key combination
- Run the command `(text-scale-increase)`
- Press the "G" key

axelf4:
> I finally found a way to reproduce the issue.

Great!

 > I am not sure whether this is a bug in GNU/Emacs. It does not seem to be the responsibility of `(goto-char (point-max))` to adjust the window start position.

Hmm, I got the impression that redisplay should handle this from the `scroll-conservatively` docstring:

> If point moves off-screen, redisplay will scroll by up to `scroll-conservatively` lines in order to bring point just barely onto the screen again.  If that cannot be done, then redisplay recenters point as usual.

The `end-of-buffer` code may be there just to tailor the `scroll-conservatively` behavior, i.e. instead of recentering, placing EOB two lines up from the bottom, and not explicitly fix an issue. (`overlay-recenter` is an obsolete optimization---I do not think we would need to include it.)

The drawback of the proposed fix is that it changes the behavior of <kbd>G</kbd>: Instead of obeying `scroll-conservatively`, it unconditionally makes EOB the bottommost line.

@tomdl89, any thoughts?

jamescherti:
By the way, the value of `scroll-conservatively` in my setup is `1000`. When I change `scroll-conservatively`  to 10, the issue does not happen.

tomdl89:
I agree with @axelf4 about the drawback mentioned, although I guess this could be addressed by checking the value of `scroll-conservatively` beforehand.

My main problem with this PR is that I can't replicate it at all, even with a high value of `scroll-conservatively` and repeated invocations of `gg`, `text-scale-increase` and `G`. I'm not sure I'd be happy to merge anything until I can have repro steps that I can follow successfully. Just to make sure I'm not missing something, what are the repro steps, starting with

1. run `make emacs` in the evil repo

jamescherti:
Does this make any difference?
```elisp
(setq scroll-conservatively most-positive-fixnum)
```

jamescherti:
You can replicate it in Emacs 30 by
1. Using the following `early-init.el` and `init.el`: https://github.com/jamescherti/minimal-emacs.d
2. After loading Emacs, evaluate this expression: `(setq scroll-conservatively most-positive-fixnum)`

jamescherti:
Yes, by "no visible text," I meant that the cursor and window start are at `(point-max)`, and all the text is located above `window-start`, remaining invisible to the user.

tomdl89:
Understood. So I think replacing `(recenter -1)` in your code with
```elisp
(recenter (and (< 100 scroll-conservatively) -1))
```
Could be an improvement. That means it will emulate vim's behaviour for people who have set `scroll-conservatively` to a high value, but will emulate `end-of-buffer`'s behaviour for those who haven't. Seems like a good compromise to me. @axelf4 wdyt?

jamescherti:
I updated the pull request.

Just out of curiosity, why specifically `< 100`?

tomdl89:
According to `scroll-conservatively`'s docstring:
>If the value is greater than 100, redisplay will never recenter point,
but will always scroll just enough text to bring point into view, even
if you move far away.

tomdl89:
(for that reason, I don't think 10000 or `most-positive-fixnum` or 101 are any different)

jamescherti:
It makes sense. Thank you for clarifying, @tomdl89.

jamescherti:
Wouldn't the following be better?

```elisp
(recenter (and (<= 100 scroll-conservatively) -1))
```

tomdl89:
The docstring says
>If the value is greater than 100…

not
>If the value is greater than or equal to 100…

so, no, I don't think your version would be better.

In case lisp isn't something you're super familiar with,
```elisp
(< 100 x)
```
is equivalnt to
```mathematica
100 < x
```
in normal maths. For that reason, I find it easier to read `<` as `strictly increasing?` and `<=` as `monotonically-increasing?` when I see them in lisp code.

jamescherti:
I wrote my response too quickly. Here is what I intended to express:

> If the value is greater than 100, redisplay will never recenter point,
> but will always scroll just enough text to bring point into view, even
> if you move far away.

This means that:
- When scroll-conservatively > 100: Redisplay will never recenter the point.
- When scroll-conservatively <= 100: The window may recenter the point if necessary.

I believe there is a small issue with the change you proposed:
```elisp
(recenter (and (< 100 scroll-conservatively) -1))
```

This is what the above will do:
- When scroll-conservatively >= 100: `(recenter nil)` (This places the point in the middle line of the window.)
- When scroll-conservatively < 100: `(recenter -1)` (This places the point on screen line -1.)

I have changed my pull request to the following instead:
```elisp
(recenter (and (> 100 scroll-conservatively) -1))
```
- When scroll-conservatively <= 100: `(recenter nil)` (when redisplay never recenter point, recenter places the point in the middle line of the window.)
- When scroll-conservatively > 100: `(recenter -1)`  (when the window may be recentered, then places the recenter on screen line -1.)

Let me know if I am missing something.

jamescherti:
Also, I just tried Vim, and the change `(and (> 100 scroll-conservatively) -1)` does not emulate Vim's behavior. I reverted it the change back to (recenter -1).

Here is an example with Vim:
image

I pressed G:
image

Similar to `(recenter -1)`.

Let me know what you think.

tomdl89:
OK, your new approach works just as fine for me. But I still think
```elisp
(>= scroll-conservatively 100)
```

needs to be
```elisp
(> scroll-conservatively 100)
```

because of `scroll-conservatively`'s docstring.

jamescherti:
While Emacs recenters the screen using `(recenter nil)` when `scroll-conservatively` is less than or equal to 100, this behavior does not match Vim's. 

To align with Vim's behavior, we should use `(recenter -1)`, regardless of the value of `scroll-conservatively`. I updated the pull request.

tomdl89:
This PR is getting really noisy with force-pushes. Can you let me know when you have stopped so I can take a look at the code and review?

jamescherti:
I had to add a few additional details to the commit message. I've now finished modifying it.

tomdl89:
@jamescherti I was happy with commit https://github.com/emacs-evil/evil/commit/42c383809095e9d56d778f386976bbb641c71ae1 so have merged that to master. Just an explanation: evil doesn't aim to emulate vim no matter what. If someone hasn't set `scroll-conservatively` to >100 then I'm happy for `G` to recenter to the middle of the screen, despite that not matching what vim does. Thanks for your work!

jamescherti:
> I was happy with commit https://github.com/emacs-evil/evil/commit/42c383809095e9d56d778f386976bbb641c71ae1 so have merged that to master. Just an explanation: evil doesn't aim to emulate vim no matter what. If someone hasn't set scroll-conservatively to >100 then I'm happy for G to recenter to the middle of the screen, despite that not matching what vim does.

I understand your perspective, and it makes sense to prioritize user experience in Evil over strict adherence to Vim's behavior.

> Thanks for your work!

My pleasure, @tomdl89.

Keep up the great work. Emacs Evil is one of the best Vim emulations available.

PR #1950, opened 2025-01-08
Make evil-shift-width safe as local variable

amnore:
The new editorconfig-mode in Emacs 30 automatically sets evil-shift-width ([here](https://github.com/emacsmirror/emacs/blob/8d687695c528e4edb04cca5602470d349e333258/lisp/editorconfig.el#L465)), but a popup window is shown every time because this variable is not marked safe as a local variable. This PR marks it safe to get rid of this annoyance.


tomdl89:
Thanks @amnore !

PR #1961, opened 2025-02-22
Correct typos in evil-repeat.el

ideasman42:


evil-repeat.el
```diff
@@ -131,7 +131,7 @@
      (evil-repeat-abort)))
 
 (defsubst evil-repeat-recording-p ()
-  "Return non-nil iff a recording is in progress."
```
tomdl89, line 134: Not sure this is a typo. I think it's meant to be [iff](https://en.wikipedia.org/wiki/If_and_only_if)

axelf4: I agree. OTOH, the GNU Emacs [Tips for Documentation Strings] discourage use of iff, so it should be fine either way.

[Tips for Documentation Strings]: https://www.gnu.org/software/emacs/manual/html_node/elisp/Documentation-Tips.html

tomdl89: Heh, particularly relevant:
> Avoid using “iff” (a mathematics term meaning “if and only if”), since many people are unfamiliar with it and **mistake it for a typo**.

PR #1975, opened 2025-04-30
Fixes #1974: Correct C-h behavior to ensure electric-pair deletes adjacent pairs

### LINKED ISSUE #1974: C-h does not ensure electric-pair deletes adjacent pairs
When using `electric-pair-mode`, the expected behavior when pressing `C-h` near a pair of adjacent delimiters (e.g., `()` `[]` `{}` `""`) is that both the opening and closing delimiters should be deleted together if they were inserted as a pair.

Electric-pair achieves this by invoking a function bound to the `DEL` key (ASCII code 127) within `electric-pair-mode-map`:
```
electric-pair-mode-map is a variable defined in ‘elec-pair.el’.

Its value is
(keymap 
 (127 menu-item "" electric-pair-delete-pair :filter
  #f(compiled-function (cmd) #<bytecode 0x115035a9952592b3>)))
```

In Evil-mode, `C-h` deletes only the character before point because it calls `evil-delete-backward-char-and-join` instead of calling the same function as the DEL key.

### steps to reproduce
1. Enable `electric-pair-mode`  
2. Ensure evil-want-C-h-delete is set to t and that evil-mode is enabled
3. Insert a pair of delimiters such as `()`  
4. Place point between the delimiters  
5. Press `C-h`  

### expected behavior**
Both delimiters should be deleted together, mirroring the behavior of `DEL` with `electric-pair-mode`.

### actual behavior
Only the character before point is deleted, leaving the unmatched delimiter behind.

### Possible fix

Making `C-h` dynamically call the same function as the `DEL` key would not only resolve the issue with `electric-pair-mode` but also ensure consistent behavior across other modes that modify the default `DEL` key functionality.

jamescherti:
When using `electric-pair-mode`, the expected behavior when pressing `C-h` near a pair of adjacent delimiters (e.g., `()` `[]` `{}` `""`) is that both the opening and closing delimiters should be deleted together if they were inserted as a pair.

This pull request fixes #1974.

This pull request corrects `C-h` behavior to ensure electric-pair deletes adjacent pairs. It calls the same function as `DEL` (code 127) in `electric-pair-mode-map`:
```
electric-pair-mode-map is a variable defined in ‘elec-pair.el’.

Its value is
(keymap 
 (127 menu-item "" electric-pair-delete-pair :filter
  #f(compiled-function (cmd) #<bytecode 0x115035a9952592b3>)))
```

This pull request could be improved. Please don't hesitate to comment it if you have any suggestions.

### DIFF
--- a/evil-vars.el
+++ b/evil-vars.el
@@ -590,8 +590,7 @@ ubiquity of prefix arguments."
                (define-key evil-replace-state-map (kbd "C-h") nil))
               ((and value
                     (not (lookup-key evil-insert-state-map (kbd "C-h"))))
-               (define-key evil-insert-state-map (kbd "C-h")
-                           'evil-delete-backward-char-and-join)
+               (define-key evil-insert-state-map (kbd "C-h") (kbd "DEL"))
                (define-key evil-replace-state-map (kbd "C-h")
                            'evil-replace-backspace))))))
 
--- a/evil-maps.el
+++ b/evil-maps.el
@@ -447,7 +447,7 @@
     ,@(when evil-want-C-u-delete
         '(("\C-u" . evil-delete-back-to-indentation)))
     ,@(when evil-want-C-h-delete
-        '(("\C-h" . evil-delete-backward-char-and-join)))
+        `(("\C-h" . ,(kbd "DEL"))))
     ([mouse-2] . mouse-yank-primary))
   "Evil's bindings for insert & replace states.
 Used in `evil-insert-state-map' and `evil-replace-state-map',


evil-vars.el
```diff
@@ -591,7 +591,11 @@ ubiquity of prefix arguments."
               ((and value
                     (not (lookup-key evil-insert-state-map (kbd "C-h"))))
                (define-key evil-insert-state-map (kbd "C-h")
```
axelf4, line 593: Is this not better achieved by a keyboard macro? I.e.:

```suggestion
               (define-key evil-insert-state-map (kbd "C-h") (kbd "DEL"))
```

axelf4 (changes requested):
I agree that we should try to be compatible with the shenanigans of `electric-pair-mode`, even at the cost of a more fragile `C-h` binding. Some comments:

Will need to update the `lookup-key` check in the `evil-want-C-h-delete` setter:
```elisp
              ((and (not value)
                    (eq (lookup-key evil-insert-state-map (kbd "C-h"))
                        'evil-delete-backward-char-and-join))
```

jamescherti:
I enhanced the `evil-insert-state-bindings` mapping for `C-h` by improving the lambda function.

jamescherti:
Have there been any updates on this pull request, @axelf4?

jamescherti:
I have applied your `(key-binding (kbd "DEL")` suggestion.

jamescherti:
I have enhanced this pull request: I added support for interactive and non-interactive functions

jamescherti:
FYI: I simplified the pull request.

jamescherti:
The only drawback to binding `(kbd "DEL")` directly is that it causes the ElDoc help to disappear from the echo area. Perhaps a function should be used instead, so it can be added to the ElDoc exclusion list.

jamescherti:
@axelf4 This merge request is ready to be merged.

jamescherti (commented):
@axelf4 This pull request is ready to be merged.

jamescherti:
@tomdl89 Could you please help review this pull request? Thanks.

jamescherti (commented):
Done

PR #1994, opened 2025-11-08
Fix evil-up-paren for Emacs 31

### LINKED ISSUE #1993: evil-inner-paren etc is broken in Emacs 31's treesit modes
For example, in python-ts-mode,

```python
print(foo(bar(hello)))
# cursor         ^
```

`vi)` selects `bar` instead of `hello`

hiecaq:
Fix #1993 

Emacs 31 introduces [up-list-function](https://github.com/emacs-mirror/emacs/blob/3cc34865e85328b9f2eeb998662fb2635c281add/lisp/emacs-lisp/lisp.el#L242) which is used in `treesit`, so we need to overrides it too to make the `up-list` calls in `evil-up-paren` work as intended.


tomdl89:
Thanks!

PR #2016, opened 2026-04-27
Fix disabling the cursor 

### LINKED ISSUE #592: Evil makes hidden cursor reappear on window configuration changes
Originally reported by: **Vasilij Schneidermann (Bitbucket: [wasamasa](https://bitbucket.org/wasamasa), GitHub: [wasamasa](https://github.com/wasamasa))**

I wrote a package that hides the cursor by setting `cursor-type` to `nil`.  This works initially, however changing the window configuration restores it because Evil has `evil-refresh-cursor` in `window-configuration-change-hook`.  The comment above suggests it should just be changing the cursor *color*...

See also [zone-nyan#5](https://github.com/wasamasa/zone-nyan/issues/5).

- Bitbucket: https://bitbucket.org/lyro/evil/issue/592

real-or-random:
Without this, evil-set-cursor treats nil as an empty list of specs (i.e., a noop). But nil is a valid single spec; namely it's a valid value of the cursor-type variable, which means "don't show a cursor".

This also inverts rewrites the logic to use (when (or ...)) instead of (unless (and ...)) because that's much more readable, at least to me.

Fixes #592.

### DIFF
--- a/evil-common.el
+++ b/evil-common.el
@@ -731,10 +731,12 @@ filename."
   "Change the cursor's apperance according to SPECS.
 SPECS may be a cursor type as per `cursor-type', a color
 string as passed to `set-cursor-color', a zero-argument
-function for changing the cursor, or a list of the above."
-  (unless (and (not (functionp specs))
-               (listp specs)
-               (null (cdr-safe (last specs))))
+function for changing the cursor, or a non-empty list of the above."
+  ;; if it's a single spec, wrap it in a list
+  (when (or (functionp specs)
+            (null specs)
+            (not (listp specs))
+            (cdr-safe (last specs)))
     (setq specs (list specs)))
   (dolist (spec specs)
     (cond


evil-common.el
```diff
@@ -732,9 +732,11 @@ filename."
 SPECS may be a cursor type as per `cursor-type', a color
 string as passed to `set-cursor-color', a zero-argument
 function for changing the cursor, or a list of the above."
```
monoid: I think the doc should be updated too.

real-or-random: Happy to do that once some maintainer shows interest in this PR

tomdl89: Yep, update the docstring and I'll merge

real-or-random: Okay, I force-pushed to make it clear that only non-empty lists are interpreted as lists. Let me know if you're happy with that change.

tomdl89: Looks good, thank you!

real-or-random:
For posterity: This was reverted in https://github.com/emacs-evil/evil/commit/30c88945002b2a734ebb778054b19b2e58995723; see the discussion in #592.

PR #2033, opened 2026-08-17
Fix some bugs related to repeating company-mode completion commands

bcc32:
Repeating `company-complete` using `evil-repeat` doesn't work
correctly without some workarounds.

Prior to this PR, I identified two bugs that occur, see below for
reproduction.  This change is intended to fix both issues.

## When there is a common prefix to complete to

In the `*scratch*` buffer in `lisp-interaction-mode`:

``` emacs-lisp
(company-mode)
(evil-mode)

(keymap-set company-mode-map "TAB" #'company-complete)

;; Start at the end of the first "req" and hit TAB, which completes to
;; "require".  Hit "j ." and see that repeat doesn't work correctly.
(req)
(req)
(req)

;; Then, evaluate this and see that the above completions work
;; correctly.
(evil-declare-change-repeat #'company-complete)
```

## When there is no common prefix

In the `*scratch*` buffer in `lisp-interaction-mode`:

``` emacs-lisp
(company-mode)
(evil-mode)

(keymap-set company-mode-map "TAB" #'company-complete)

;; Start at the end of the first "re" and hit TAB, then pick some completion.
;; Observe this error in the *Messages* buffer:

;; Error in post-command-hook (evil-repeat-post-hook): (wrong-type-argument number-or-marker-p nil)

(re)
(re)
(re)
```


bcc32:
For context, the post-command-hook issue happens because `company-complete` uses `call-interactively` and that causes `this-command` to be different betwen the pre- and post-hooks.  See https://github.com/company-mode/company-mode/issues/383

